PR #6, opened 2021-06-19
Let autotools know about the samples

orlitzky:
The individual commits should explain themselves pretty well. Basically we tell autotools about the `c/samples` directory so that the samples can be installed automatically. The first commit is unrelated but trivial and should not change anything at all.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,3 +1,5 @@
+SUBDIRS = c/samples
+
 lib_LTLIBRARIES = libplanarity.la
 
 libplanarity_la_SOURCES = \
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,8 +25,7 @@ libplanarity_la_SOURCES = \
 	c/stack.c
 libplanarity_la_LDFLAGS = -no-undefined $(AM_LDFLAGS) -version-info @LT_CURRENT@:@LT_REVISION@:@LT_AGE@
 
-headersdir = $(includedir)/planarity/
-headers_HEADERS = \
+pkginclude_HEADERS = \
 	c/appconst.h \
 	c/graph.h \
 	c/graphDrawPlanar.h \
--- a/Makefile.am
+++ b/Makefile.am
@@ -53,3 +53,5 @@ planarity_SOURCES = c/planarity.c c/planarityCommandLine.c
 
 man1_MANS = planarity.1
 EXTRA_DIST=$(man1_MANS)
+
+TESTS = test-samples.sh
--- a/c/planarityCommandLine.c
+++ b/c/planarityCommandLine.c
@@ -118,9 +118,9 @@ int runSpecificGraphTest(char *command, char *infileName);
 int runQuickRegressionTests(int argc, char *argv[])
 {
 	if (runSpecificGraphTests() < 0)
-		return -1;
+		return NOTOK;
 
-	return 0;
+	return OK;
 }
 
 int runSpecificGraphTests()
--- a/c/samples/Makefile.am
+++ b/c/samples/Makefile.am
@@ -1,0 +1,39 @@
+# The @docdir@ variable will be replaced by the ./configure script.
+# This is "necessary" (that is, I don't know a better way to do it)
+# to install these samples in a "samples" subdirectory of the usual
+# documentation directory.
+docdir = @docdir@/samples
+
+dist_doc_DATA = \
+  Petersen.0-based.txt \
+  Petersen.0-based.txt.ColorVertices.out.txt \
+  Petersen.0-based.txt.K23Search.out.txt \
+  Petersen.0-based.txt.K33Search.out.txt \
+  Petersen.0-based.txt.K4Search.out.txt \
+  Petersen.0-based.txt.OuterplanarEmbed.out.txt \
+  Petersen.0-based.txt.PlanarEmbed.out.txt \
+  Petersen.txt \
+  Petersen.txt.ColorVertices.out.txt \
+  Petersen.txt.K23Search.out.txt \
+  Petersen.txt.K33Search.out.txt \
+  Petersen.txt.K4Search.out.txt \
+  Petersen.txt.OuterplanarEmbed.out.txt \
+  Petersen.txt.PlanarEmbed.out.txt \
+  drawExample.0-based.txt \
+  drawExample.0-based.txt.ColorVertices.out.txt \
+  drawExample.0-based.txt.DrawPlanar.out.txt \
+  drawExample.0-based.txt.DrawPlanar.out.txt.render.txt \
+  drawExample.txt \
+  drawExample.txt.ColorVertices.out.txt \
+  drawExample.txt.DrawPlanar.out.txt \
+  drawExample.txt.DrawPlanar.out.txt.render.txt \
+  maxPlanar5.0-based.txt \
+  maxPlanar5.0-based.txt.ColorVertices.out.txt \
+  maxPlanar5.0-based.txt.DrawPlanar.out.txt \
+  maxPlanar5.0-based.txt.DrawPlanar.out.txt.render.txt \
+  maxPlanar5.0-based.txt.PlanarEmbed.out.txt \
+  maxPlanar5.txt \
+  maxPlanar5.txt.ColorVertices.out.txt \
+  maxPlanar5.txt.DrawPlanar.out.txt \
+  maxPlanar5.txt.DrawPlanar.out.txt.render.txt \
+  maxPlanar5.txt.PlanarEmbed.out.txt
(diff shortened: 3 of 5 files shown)


graph-algorithms:
This is a good catch, thank you, Michael.
The files in the samples directory exist to make it easy to enter on the command line "planarity -test" and have the software demonstrate that it is operating correctly by producing outputs that match the expected outputs in the samples directory from inputs in the samples directory.
This PR makes mention of placing the files in a configurable docs directory, and that a default docs directory is chosen.
Particularly for the default, is the result after make and make install that you can run "planarity -test" or does "planarity -test" fail to find the sample files?
If it fails, could you revise the PR with how you'd envision changing the behavior of the -test parameter to include how to correctly look for the samples? E.g., if -test can't find the samples in the current places where it looks, then is there some standard environment variable that should be queried to find them wherever they have been placed during installation?

orlitzky:
Oh, I didn't realize there was a test suite built in to the command-line program. That's good news!

You're right that `planarity` won't be able to find the samples after installation, but that's not something I would change. With the directory hard-coded in `planarityCommandLine.c`, you're forced to choose between two paths: the before-installation path and the after-installation-path. And the test suite can work only in one of those two situations.

Doing it before installation is more common, for the simple reason that you don't want to install (a new version of the) software before you know that it works. The test -> install procedure is even baked into autotools. There is an optional `make check` step that comes between `make` and `make install`. I can amend the PR to run `planarity -test` with `make check` so that it tests the program before installation. I think that's the best easy, short-term solution, but as a consequence the `-test` option will remain useless once the program is installed. (The "good news" is that this has always been the case and does not represent a regression :smile:)

Longer-term, we could separate the `planarity -test` code into a separate executable, called (say) `planarity-test`, whose sole job would be to run the test suite out of the current directory, i.e. the source tree. We could then change `make check` to invoke `planarity-test`, but omit `planarity-test` from the installation image. That way. the tests would still be run before installation, but after installation, there would be no lingering `-test` flag on the main `planarity` executable.

orlitzky:
Another idea... since the test suite is mostly just running `planarity` on some input files with various options, you could add a `README` to the `samples` directory that explains the inputs, planarity commands, and the expected outputs. That would allow users to verify the results themselves, after installation, and doubles as a bit of documentation.

graph-algorithms:
Hi Michael,

I think what would be the best here is if you could roll back on your latest commit into this PR so that I could more quickly merge your first version (two commits) and then update the 3.0.1.0 release tag with a new tarball that includes the samples. I think it is better to have the samples in that new tarball, and it is reasonable for 3.0.1.0 to say that anyone wanting to run planarity -test should change directories to the samples directory first. I agree, it's not a regression.

It would still be nice to still have your latest two commits, but I'd like to do it as a separate PR because it includes a change to the C code, so that should trigger a 3.0.1.1 release rather than being rolled into 3.0.1.0. Moreover, the C code fix needs to be a tiny bit larger than what's in your current commit, because the reason for the error is a little different than what you mentioned in your commit comment.

The real problem is that one optimization I made some time ago was to redefine OK and NOTOK to be 1 and 0 (in appconst.h) so that many function results in the core algorithms could be tested directly as booleans without making a comparisons to OK and NOTOK. When that optimization happened, I missed adjusting -test (which you caught) but also -help/-h and -info/-i. So, the tiny extra bit is that helpMessage() in planarity.c should also return OK rather than return 0.

I'm sure nobody has noticed up to now that the planarity executable returns the wrong code when invoked with -test, -help/-h, or -info/-i because the end-user is looking at the screen to see the output. That being said, no harm in fixing this up... just in a separate PR so that it can be done in a separate release. You OK to do it this way?

Kind regards,
John

orlitzky:
> Hi Michael,
> 
> I think what would be the best here is if you could roll back on your latest commit into this PR so that I could more quickly merge your first version (two commits) and then update the 3.0.1.0 release tag with a new tarball that includes the samples.

If you don't mind, it's actually easier for me if you wait until 3.0.1.1 to add the samples. While the number of planarity users in Gentoo is still probably in the double digits, a SHA-2 hash of the 3.0.1.0 release tarball was signed with my GPG signature and distributed to everyone's machines a few days ago. Modifying the tarball will invalidate the existing hash, and updating the hash would make the old planarity-3.0.1.0.tar.gz look corrupt for anyone who has already downloaded it. That can all be worked around, but if 3.0.1.1 is around the corner, I'd rather not rock the boat.

graph-algorithms:
Cool, that's the plan then.

graph-algorithms:
Hi Michael,

I tried out the latest two commits, and they aren't quite working out for me. From the edge-addition-planarity-suite directory, I get into bash and then do configure, make dist, and make check. The end of the make check gives this:

make  test-samples.sh
make[2]: Entering directory `/c/Users/John/Documents/eclipse/workspaces-cpp/graph-algorithms/edge-addition-planarity-suite'
make[2]: Nothing to be done for `test-samples.sh'.
make[2]: Leaving directory `/c/Users/John/Documents/eclipse/workspaces-cpp/graph-algorithms/edge-addition-planarity-suite'
make  check-TESTS
make[2]: Entering directory `/c/Users/John/Documents/eclipse/workspaces-cpp/graph-algorithms/edge-addition-planarity-suite'
cp: cannot stat `/*.txt': No such file or directory
FAIL: test-samples.sh
===============================
1 of 1 test failed
Please report to jboyer@acm.org
===============================
make[2]: *** [check-TESTS] Error 1
make[2]: Leaving directory `/c/Users/John/Documents/eclipse/workspaces-cpp/graph-algorithms/edge-addition-planarity-suite'
make[1]: *** [check-am] Error 2
make[1]: Leaving directory `/c/Users/John/Documents/eclipse/workspaces-cpp/graph-algorithms/edge-addition-planarity-suite'
make: *** [check-recursive] Error 1

It also leaves behind a .lib directory that adds several unstaged changes as well as adding a .dirstamp as an unstaged change into the c subdirectory of edge-addition-planarity-suite.

Would you be able to amend this so the test succeeds and the files are removed? (Maybe the files get removed if the test succeeds, in which case, the problem reduces to make the test succeed).

Thank you!
John

orlitzky:
For the test failure, it looks like `$samplesdir` is empty when you're running `test-samples.sh`, but the top-level `Makefile.am` should be setting it. Can you try `git clean -x -f -d` to start over from scratch, and then `autoreconf -fi` to regenerate the autotools stuff? I am really hoping that's all that's wrong.

The leftover files are to be expected. There are actually many others, but they're all hidden by the `.gitignore` file in the repository. If you want, I can add these files to `.gitignore` too?

graph-algorithms:
Hi Michael,

Yes, totally clean and I tried a few experiments, including validating that there's no behavioral difference between using autogen.sh versus autoreconf -fi (I've been using autogen.sh because it was previously contributed, so I wanted to make sure the new process works with whatever Jeroen expected from the previous process). Anyway, the conclusion from the experiments is that the export isn't working under MSYS's bash/make/etc. (which is a required platform).

As a workaround, I tried replacing a few of the commands in test-samples.sh to eliminate reliance on the export. The result is a process that now does work for me (the new check runs successfully as part of make distcheck, and all files are cleaned up afterward, due to the success). I'd like to ask if you could amend the code on your side and see if the fixes also work for you.  If so, then you could amend this PR with the different content for test-samples.sh (and probably just remove the export in Makefile.am).

In short, test-samples.sh works for me this way:

cd ".."                             || exit 1
mkdir samples                       || exit 2
cp ./c/samples/*.txt ./samples/    || exit 3
./_build/planarity -test                   || exit 4
rm -rf ./samples                    || exit 5

Let me know!

Thanks,
John

orlitzky:
The build directory is not guaranteed to be `_build`, unfortunately. I'll take another stab at this without the exports. I should be able to have `./configure` substitute the same paths in directly (creating `test-samples.sh` from `test-samples.sh.in`)

orlitzky:
Ok, I tweaked that last commit. Maybe this one works better.

john-boyer-phd:
This version works for me! 
The only slightly unusual thing I noticed is that the chmod runs (twice) just before test-samples.sh is created, so chmod reports no such file or directory. Nonetheless, make dist and make distcheck work on my end. Here's the order I see:

...
checking for time.h... yes
checking for unistd.h... (cached) yes
configure: creating ./config.status
config.status: creating Makefile
chmod: cannot access `test-samples.sh': No such file or directory
config.status: creating c/samples/Makefile
chmod: cannot access `test-samples.sh': No such file or directory
config.status: creating test-samples.sh
config.status: executing depfiles commands
config.status: executing libtool commands
...

Are you seeing a different run order?

orlitzky:
Good catch, the `chmod +x` was being run after each entry in `AC_CONFIG_FILES`, and only the third entry created `test-samples.sh`. I force-pushed a fix for that.

john-boyer-phd:
That's a great fix, Michael, all set. Thank you for all the work you have done to make this improvement!

orlitzky:
Thank you for your patience and testing!

PR #23, opened 2024-03-04
Issue19  - Support conversion from formats including .g6 format to adjacency list format 

### LINKED ISSUE #19: Story: Support conversion from .g6 format to adjacency list format
Part of Epic #10 : Support the following command-line option configuration: `planarity -t [-q] -tga I O`.

wbkboyer:
Resolves #19 

## Added
* `c/g6-api-utilities.c/.h` - internal helper functions for `G6ReadIterator`
* `c/g6-read-iterator.c/.h` - "methods" operating over `G6ReadIterator` to handle reading `.g6` graph files
* `c/samples/`
  * `nauty_example.g6` - example graph from the [`.g6` format specification](https://users.cecs.anu.edu.au/~bdm/data/formats.txt) 
  * `nauty_example.g6.0-based.AdjList.out.txt` - expected output from running
  ```
  planarity -t -ta ./c/samples/nauty_example.g6 ./c/samples/nauty_example.g6.0-based.AdjList.out.txt
  ```
  * `K10.g6` - needed a sample where `gp_AddEdge` would return `NONEMBEDDABLE` to demonstrate the need for `gp_DynamicAddEdge`; generated using
  ```
  ./geng -h 10 45:0 > ../edge-addition-planarity-suite-fork/c/samples/K10.g6
  ```
  * `K10.g6.0-based.AdjList.out.txt` - expected output from running
  ```
   planarity -t -ta ./c/samples/K10.g6 ./c/samples/K10.g6.AdjList.expected.out.txt
  ```
* `c/planarityTestGraphFunctionality.c`
  * declares helper functions `transformFile` and `transformString`, which are called respectively by `TestGraphFunctionality` depending on whether or not `inputStr` is `NULL` (i.e. if `inputStr` is `NULL`, must mean we wish to `gp_Read` from the file specifed by `inputFile` rather than `gp_ReadFromString` on `inputStr`)
* `c/strOrFile.c/.h` - adds the `strOrFileP` struct, which either is assigned a `FILE` pointer or is assigned an `inputStr`; the `inputStrPos` keeps track of the current "start" of the "remainder" of the input string when performing `sf_getc()`, `sf_ungetc()`, and `sf_fgets()`

## Updated
* `.vscode/launch.json` - updated launch configurations to include sample `args`, including how to run planarity tests with repository samples directory
* `c/graph.h` - declared `gp_DynamicAddEdge`
* `c/graphIO.c`
  * initialize `N` to `-1` in `_ReadAdjMatrix`, `_ReadAdjList`, and `_ReadLEDAGraph` so that `gp_InitGraph` will fail if we were unable to read the order from the file (because it _could_ be a `.g6` file, or it could just be an invalid file.
  * `gp_Read` and `gp_ReadFromString` now try to `_ReadGraphFromG6FilePointer` or `_ReadGraphFromG6String` respectively when reading the file or string as other formats fails. For `.g6` format, we skip any post-processing that adds extra data to the graph datastructure
* `c/graphUtils.c` - defined `gp_DynamicAddEdge`, which doubles the arc-capacity when we get `NONEMBEDDABLE` from the first attempt to `gp_AddEdge`
* `c/planarity.h`
  * updating declaration of `RandomGraphs` to include parameter names to match style of other declarations
  * declaring `TestGraphFunctionality`, `GetTransformationName` and  `ConstructTransformationOutputFilename`
* `c/planarityCommandLine.c`
  * declared `callTestGraphFunctionality`, which is called by `commandLine` when the third command-line argument is `-t`. This function handles parsing the command-line arguments (most especially the `commandString`) and calling `TestGraphFunctionality`
  * declared `runGraphTransformationTest`, which is called by the smoke-test suite  `runSpecificGraphTests`, with the `inputInMemFlag` set to `TRUE` then `FALSE` to test how we handle receiving the graph input as a string vs. receiving a file pointer to a graph input file
* `c/planarityUtils.c` - defines `GetTransformationName`, `GetBaseName`, and `ConstructTransformationExpectedResultFilename`, which were declared in `c/planarity.h`

## Removed
* N/A


john-boyer-phd (changes requested):
Lots of good code in here. Mostly tweaks or no-change comments, but a few meaty ones that will make things better for the project than before G6 support.

c/planarity.c
```diff
@@ -47,6 +47,20 @@ int main(int argc, char *argv[])
...
 
+#ifdef DEBUG
+	// When one builds and runs the executable in an external console from an IDE
+	// such as VSCode, the external console window will close immediately upon
+	// exit 0 being returned. This means that one may miss Messages and
+	// ErrorMessages that are crucial to the debugging process. Hence, if we compile
+	// with the DDEBUG flag, this means that in appconst.h we #define DEBUG. That way,
+	// this prompt will appear only for debug builds, and will ensure the console
+	// window stays open until the user proceeds.
+	printf("\n\tPress return key to exit...\n");
```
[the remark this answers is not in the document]
wbkboyer: Confirmed with clang on MacOS, MinGW gcc on Windows, and MSVC cl on Windows (via internal console in VSCode), and internal build tools within Eclipse that one must eventually press the return key for `getc(stdin)` to execute (i.e. pressing any number of characters before that return key doesn't work).

c/planarityTestGraphFunctionality.c
```diff
@@ -0,0 +1,107 @@
...
+				ErrorMessage("Unable to transform file.\n");
+			}
+			else
+			{
+				if (outputStr != NULL)
+					Result = gp_WriteToString(theGraph, outputStr, outputFormat);
+				else
+					Result = gp_Write(theGraph, outfileName, outputFormat);
+				
+				if (Result == NOTOK)
```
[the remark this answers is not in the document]
wbkboyer: Nice, this and the above change means two less "error" messages when comparing with NOTOK.

c/graphIO.c
```diff
@@ -312,11 +316,17 @@ int  _ReadAdjList(graphP theGraph, FILE *Infile, strBufP inBuf)
...
-	int N, M, m, u, v, ErrorCode;
+	int N = -1;
+    int M, m, u, v, ErrorCode;
 	int zeroBasedOffset = gp_GetFirstVertex(theGraph)==0 ? 1 : 0;
 
     /* Skip the lines that say LEDA.GRAPH and give the node and edge types */
     fgets(Line, 255, Infile);
+
+    if (strncmp(Line, "LEDA.GRAPH", strlen("LEDA.GRAPH")) != 0) {
+        return NOTOK;
```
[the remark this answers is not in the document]
wbkboyer: I've removed this test due to the more refined machinery added to `gp_Read` and `gp_ReadFromString` to discern the input type and route appropriately.

c/g6-read-iterator.c
```diff
@@ -0,0 +1,795 @@
...
+			{
+				row = 0;
+				col++;
+			}
+
+			bitValue = ((currByte >> j) & 1u) ? 1 : 0;
+			if (bitValue == 1)
+			{
+				// Add gp_GetFirstVertex(pGraph), which is 1 if NIL == 0 (i.e. 1-based labelling) and 0 if NIL == -1 (0-based)
+				exitCode = gp_AddEdge(pGraph, row+gp_GetFirstVertex(pGraph), 0, col+gp_GetFirstVertex(pGraph), 0);
```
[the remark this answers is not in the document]
wbkboyer: Generated `K10.g6` with

```
% ./geng -h 10 45:0 > ../edge-addition-planarity-suite-fork/c/samples/K10.g6
>A ./geng -d0D9 n=10 e=45
>Z 1 graphs generated in 0.00 sec
```

And I've verified that the current mechanism to `_decodeGraph` flames out when `gp_AddEdge` returns `NONEMBEDDABLE` (defined in `graphstructures.h`) when we hit `M = 30` edges (I guess I'm kinda surprised an obstruction to planarity (i.e. a `K_{3,3}`)wasn't found after `M >= (3 * N - 5)`, since `N > 5`??)

As discussed 2024-03-06, the plan is to write a wrapper function for `gp_AddEdge`, i.e. `gp_DynamicAddEdge`, so that we don't have to duplicate this code in each of the subordinate functions of `gp_Read` and `gp_ReadFromString`.

c/g6-read-iterator.c
```diff
@@ -0,0 +1,795 @@
...
+		(*ppG6ReadIterator)->currGraph = NULL;
+
+		free((*ppG6ReadIterator));
+		(*ppG6ReadIterator) = NULL;
+	}
+
+	return exitCode;
+}
+
+int _ReadGraphFromG6String(graphP pGraphToRead, char *g6EncodedString)
```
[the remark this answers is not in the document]
wbkboyer: I've now added the `strOrFile` container, and have refactored `_ReadGraphFromG6String` to actually read from string (instead of using `fputs` to dump the string into a `tmpfile` and calling `_ReadGraphFromG6FilePointer`).

c/graphIO.c
```diff
@@ -383,34 +379,42 @@ int  _ReadLEDAGraph(graphP theGraph, FILE *Infile)
...
-    // Added second conjunct because we only want to try to read as .g6
-    // if trying to read as another format failed; NONEMBEDDABLE happens
-    // when we exceed the arc capacity, but that doesn't mean the graph
-    // read failed.
-    if (RetVal != OK && RetVal != NONEMBEDDABLE) {
-        RetVal = _ReadGraphFromG6FilePointer(theGraph, Infile);
+    fgets(lineBuff, 255, Infile);
+    // Reset file pointer to beginning of file
+    fseek(Infile, 0, SEEK_SET);
+    if (strncmp(lineBuff, "LEDA.GRAPH", strlen("LEDA.GRAPH")) == 0)
```
john-boyer-phd, line 394: Excellent refactor for detecting the file type from starting content.

c/g6-read-iterator.c
```diff
@@ -676,28 +705,51 @@ int _ReadGraphFromG6String(graphP pGraphToRead, char *g6EncodedString)
...
 	{
-		ErrorMessage("Unable to create temporary file to contain .g6 string contents.\n");
-		return NOTOK;
+		ErrorMessage("Unable to allocate G6ReadIterator.\n");
+		return exitCode;
 	}
 
-	int fputsCode = fputs(g6EncodedString, tmpG6Infile);
-	if (fputsCode == EOF)
+	strOrFileP g6Input = sf_New(NULL, g6EncodedString);
```
john-boyer-phd, line 717: I like the different beginG6ReadIteration methods with different type-based entry points. In C++, they could all have the same name and just accept different second parameters. In C, we use different names to compensate for absence of the function signature feature. Please add another method called `beginG6ReadIterationFromString()` to receive a char * as second parameter for a literal string. In its implemention, do the `sf_New()` code here then invoke `_beginG6ReadIteration()`. Then, at this point, invoke `beginG6ReadIterationFromString()`. This will do two things. First, you will have better consistency with the look of code between `_ReadGraphFromG6String()` and `_ReadGraphFromG6File()`. Second, the `sf_New()` will be at the same abstraction level as the `sf_Free()` (`endG6ReadIteration()` calls `sf_Free()` so a `beginG6ReadIteration ...  ()` method should be the thing that creates the string or file object).

c/planarityCommandLine.c
```diff
@@ -416,15 +447,32 @@ int runGraphTransformationTest(char *command, char *infileName, int inputInMemFl
...
+		// We want to output to string, so we pass in the address of the actualOutput string.
+		Result = TestGraphFunctionality(command, infileName, inputString, &zeroBasedOutputFlag, NULL, &actualOutput);
 		
 		if (Result != OK || actualOutput == NULL)
 		{
 			ErrorMessage("Failed to perform transformation to produce .g6 output.\n");
 		}
 		else
 		{
+			// Final arg is baseFlag, which is dependent on whether the FLAGS_ZEROBASEDIO is set in a graphP's internalFlags
```
john-boyer-phd, line 464: I think you have made an improvement such that this comment isn't needed anymore.

c/planarityTestGraphFunctionality.c
```diff
@@ -50,17 +50,22 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
...
-			if (Result == NOTOK) {
+			if (Result != OK) {
 				ErrorMessage("Unable to transform file.\n");
 			}
 			else
 			{
+				// Want to know whether the output is 0- or 1-based; will always be 
+				// 0-based for transformations of .g6 input
+				if (outputBase != NULL)
+					(*outputBase) = theGraph->internalFlags & FLAGS_ZEROBASEDIO;
```
john-boyer-phd, line 61: I think you need parens around the bitwise and, then use ternary conditional of ? 1 : 0.  This will give you a 1 regardless of which bit FLAGS_ZEROBASEDIO happens to be.

c/planarityUtils.c
```diff
@@ -548,29 +557,25 @@ int ConstructTransformationOutputFilename(char *infileName, char **outfileName,
 
 	if ((*outfileName) == NULL)
 	{
-		// If the primary output filename has not been given, then we use
-		// the input filename + the algorithm name + a simple suffix
-		// TODO: should I #define TRANSFORMATIONNAMEMAXLENGTH?
-		if ((transformationNameLen = strlen(transformationName)) <= ALGORITHMNAMEMAXLENGTH)
+		(*outfileName) = (char *) calloc(infileNameLen + 1 + strlen(baseName) + 1 + strlen(transformationName) + ((command == 'g') ? strlen(".g6") : strlen(".out.txt")) + 1, sizeof(char));
```
john-boyer-phd, line 560: I think the output .g6 file should probably have an ending of `".out.g6"`.

c/strOrFile.c
```diff
@@ -0,0 +1,168 @@
...
+		if (theStrOrFile->inputFile != NULL)
+			charToReturn = ungetc(theChar, theStrOrFile->inputFile);
+		// Don't want to ungetc to an index before theStrOrFile->inputStr start
+		else if (theStrOrFile->inputStr != NULL)
+		{
+			if (theStrOrFile->inputStrPos <= 0)
+				return EOF;
+
+			// Decrement inputStrPos, then replace the character in inputStr at that position with theChar
+			charToReturn = theStrOrFile->inputStr[--(theStrOrFile->inputStrPos)] = theChar;
```
john-boyer-phd, line 92: Concision achieved.

c/strOrFile.c
```diff
@@ -0,0 +1,168 @@
...
+
+	if (str == NULL || count < 0 || theStrOrFile == NULL)
+		return NULL;
+	
+	if (theStrOrFile->inputFile != NULL)
+	{
+		fgets(str, count, theStrOrFile->inputFile);
+		return str;
+	}
+	else if (theStrOrFile->inputStr != NULL)
```
john-boyer-phd, line 125: I think you need to add to this condition `&& theStrOrFile->inputStr[0] != '\0'` because we want `sf_fgets()` to return NULL when we reach the string version of EOF, which is the null terminator.

john-boyer-phd (changes requested):
Excellent, just a few tweaks left, and thank you for checking execution on Mac and Windows. Please also check execution on Linux.

PR #43, opened 2024-04-05
Issue 22 - Support testing an algorithm on all graphs in a .g6 file

### LINKED ISSUE #22: Story in Epic 10: Support testing an algorithm on all graphs in a .g6 file
Part of Epic #10 : Support the following command-line option configuration: `planarity -t [-q] C I O`.

wbkboyer:
Resolves #22 

## Updated
* `c/planarity.c` - updated `helpMessage()` with full `-t` options
* `c/planarityCommandLine.c` - updated docstring for `callTestGraphFunctionality()` with full `-t` options and description of functionality
* `c/planarityTestGraphFunctionality.c` - added `testAllGraphs()`, which leverages the `G6ReadIterator` to loop over the contents of a `.g6` file (`I`) and apply the algorithm corresponding to the `command` to each graph, then outputs the aggregate results to file `O`, e.g. 
  ```
   test.n11.m20.g6
   -p 15108047 3407101 11700946
  ```


c/planarityTestGraphFunctionality.c
```diff
@@ -72,10 +79,100 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
...
+					if (finalSlash == NULL)
+					{
+						finalSlash = strrchr(infileName, '\\');
+						if (finalSlash == NULL)
+						{
+							finalSlash = infileName;
+						}
+					}
+
+					infileBasename = finalSlash + 1;
```
john-boyer-phd, line 146: I think this should be written slightly differently for two reasons: 
1. You should be able to rely on FILE_DELIMITER in appconst.h to cut this down to one strrchr() call. Double-check that windows and mac are still working after this change. 
2. The current approach loses the first character when infileName contains no FILE_DELIMITER, so I'd then recommend getting rid of the whole if construct and just using a ternary conditional in the assignment to infileBasename. Specifically, `infileBasename = finalSlash ? finalSlash+1 : infileName;`.

c/planarityTestGraphFunctionality.c
```diff
@@ -72,10 +79,100 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
...
+
+					char *headerStr = (char *) malloc((strlen(infileBasename) + 2) * sizeof(char));
```
john-boyer-phd, line 148: Check malloc() for failure and also add another character because you may be slightly overwriting memory (on Windows only) if the sprintf() writes \r\n into your string in place of \n.

c/planarityTestGraphFunctionality.c
```diff
@@ -111,3 +212,130 @@ int transformString(graphP theGraph, char *inputStr)
...
+
+	exitCode = beginG6ReadIterationFromG6String(pG6ReadIterator, inputStr);
+
+	if (exitCode != OK)
+	{
+		ErrorMessage("Unable to begin .g6 read iteration.\n");
+
+		exitCode = freeG6ReadIterator(&pG6ReadIterator);
+
+		if (exitCode != OK)
```
john-boyer-phd, line 247: If the begin failed but then the free succeeds, this will return an OK exitCode when it shouldn't. It'd be better to just say `if (freeG6ReadIterator(...) != OK) then error message` because the exitCode is not reassigned from the original failure code produced by the begin iteration call. Or, just freeG6ReadIterator() but don't bother to check if it fails and print an error because the begin already failed.

c/planarityTestGraphFunctionality.c
```diff
@@ -111,3 +212,130 @@ int transformString(graphP theGraph, char *inputStr)
...
+		ErrorMessage("Unable to allocate graph to store copy of original graph before embedding.\n");
+		return NOTOK;
+	}
+
+	exitCode = gp_InitGraph(copyOfOrigGraph, pG6ReadIterator->graphOrder);
+	if (exitCode != OK)
+	{
+		ErrorMessage("Unable to initialize graph datastructure to store copy of original graph before embedding.\n");
+		gp_Free(&copyOfOrigGraph);
+		freeG6ReadIterator(&pG6ReadIterator);
```
john-boyer-phd, line 267: If you decide to keep checking the free iterator result above, then you have to check it here too. I think we're already in a failure state, so best just to free it without checking since we don't really get more info that could help with the earlier failure.

c/planarityTestGraphFunctionality.c
```diff
@@ -111,3 +212,130 @@ int transformString(graphP theGraph, char *inputStr)
...
+			ErrorMessage(Line);
+			break;
+		}
+
+		gp_ReinitializeGraph(copyOfOrigGraph);
+	}
+
+	if (exitCode == OK || exitCode == NONEMBEDDABLE)
+	{
+		// pG6ReadIterator->numGraphsRead is only incremented after successfully decoding
```
john-boyer-phd, line 311: Why is this comment here?

c/planarityTestGraphFunctionality.c
```diff
@@ -111,3 +212,130 @@ int transformString(graphP theGraph, char *inputStr)
...
+		char *resultsStr = (char *) malloc((3 +_getNumCharsToReprInt(numGraphsRead) +
+											1 + _getNumCharsToReprInt(numOK) +
+											1 + _getNumCharsToReprInt(numNONEMBEDDABLE) + 3) * sizeof(char));
+		sprintf(resultsStr, "-%c %d %d %d\n", command, numGraphsRead, numOK, numNONEMBEDDABLE);
+		sf_fputs(resultsStr, testOutput);
+		free(resultsStr);
+		resultsStr = NULL;
+	}
+
+	exitCode = endG6ReadIteration(pG6ReadIterator);
```
john-boyer-phd, line 322: The following two exitCode reassignments will also possibly hide a failure that occurred in the core loop. Better to just say `if (endG6ReadInteration(...) != OK) error message` and same for `freeG6ReadIterator()`.

john-boyer-phd (changes requested):
Looks great. Very exciting to see this working, even more so now that it is immediately showing how important this is given the problem we've found with the K_{3,3} results. Only a few minor adjustments.

john-boyer-phd (approved):
Looks great! This was the main story inspiring the epic, so congratulations on successfully finishing it!

PR #44, opened 2024-04-09
Issue 36 - Update Autotools Build Files

### LINKED ISSUE #36: Story in Epic 10: Update Autotools Build Files
This is part of Epic #10. 

Several files have been added as part of adding API support for `.g6` files in prior stories. These include read and write iterators, api utils, and new planarity app level files. 

- [ ] The `Makefile.am` should be updated to reflect the additional files.

- [ ]  The `Makefile.am` in the `samples` directory should be updated if there are any new sample files.

- [ ] We will wait to update `Configure.ac` until the story for preparing the release, at which point we will need to update the planarity version in AC_INIT and also the three `LT_*` changes to be made.

- [ ] On Windows, test the distribution make sequence (per the readme instructions, this starts with going into `bash` and `autogen.sh` and ends with `make distcheck`). This will include a successful smoke test after a successful build, proving that the correct files were all added to the two `Makefile.am` files above.

- [ ] Optionally, also test distribution make sequence on Linux.

wbkboyer:
Resolves #36

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,9 @@ compile
 
 # Ignore binary produced by compiling on MacOS
 planarity
+
+# Ignore make output
+**/m4/
+test-driver
+configure.ac~
+configure~
--- a/.gitignore
+++ b/.gitignore
@@ -90,6 +90,7 @@ ltmain.sh
 configure
 aclocal.m4
 compile
+test-driver
 
 # Ignoring configuration for VSCode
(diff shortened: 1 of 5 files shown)


john-boyer-phd (approved):
Looks great. Especially good you did the optional Linux component because it means for the next story about the man page, you'll be able to directly test the man page changes after installing from the compressed tarball you were able to make here.

PR #53, opened 2024-04-16
Issue 52 - Add platform time calls and update output to include timing info

### LINKED ISSUE #52: Issue 52 - Add platform time calls and update output to include timing info
Contributes to #49 

- [ ] Add platform time calls to detect how long it takes to test all graphs.
  * Take the start time before `ReadTextFileIntoString()` and take the end time at the end of the same code block. Before `ReadTextFileIntoString()`, emit a `Message("Starting to test all graphs\n)` and after taking the end time, emit a `Message()` to inform the user that the process is finished and how long it took.
- [ ] Header for file should be of the format
  ```
  FILENAME="<inputBasename>" DURATION="<duration>"
  ```
- [ ] Results should be displayed in the format `"-%c %d %d %d %s\n"`, populated with the command character, the total number of graphs processed, the number that returned `OK`, the number that returned `NONEMBEDDABLE`, and either `ERROR` or `SUCCESS` to indicate whether an error was encountered while processing graphs in the file.

wbkboyer:
Resolves #52 

Refactored `testAllGraphs` so that it only returns the results via `testAllStats` struct; `TestGraphFunctionality` encloses `testAllGraphs` call with timing calls and produces the output (either to string or to file via the `strOrFile` struct)

Currently, output is of the form:
```
FILENAME="test.n11.m20.g6" DURATION="87.134"
-p 15108047 3407101 11700946 SUCCESS
```

### DIFF
--- a/c/planarityTestGraphFunctionality.c
+++ b/c/planarityTestGraphFunctionality.c
@@ -174,12 +174,6 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
 						ErrorMessage(messageContents);
 						Result = NOTOK;
 					}
-
-					if (inputStr != NULL)
-					{
-						free(inputStr);
-						inputStr = NULL;
-					}
 				}
 			}
 		}
@@ -398,6 +392,8 @@ int outputTestAllGraphsResults(char command, testAllStatsP stats, char * infileN
 			{
 				fclose(outputFileP);
 				outputFileP = NULL;
+				// TODO: (#56) What happens to file that we opened for write, then immediately close? We won't be able to remove
+				// because testOutput failed to initialize
 			}
 		}
 	}
@@ -423,11 +419,6 @@ int outputTestAllGraphsResults(char command, testAllStatsP stats, char * infileN
 				{
 					(*outputStr)[0] = '\0';
 					testOutput = sf_New(NULL, (*outputStr));
-					if (testOutput == NULL)
-					{
-						free((*outputStr));
-						(*outputStr) = NULL;
-					}
 				}
 			}
 		}
@@ -442,6 +433,12 @@ int outputTestAllGraphsResults(char command, testAllStatsP stats, char * infileN
 
 		free(resultsStr);
 		resultsStr = NULL;
+
+		if (outputStr != NULL && (*outputStr) != NULL)
+		{
+			free((*outputStr));
+			(*outputStr) = NULL;
+		}
 
 		return NOTOK;
 	}
@@ -464,49 +461,7 @@ int outputTestAllGraphsResults(char command, testAllStatsP stats, char * infileN
 	if (Result == OK)
 	{
 		if (outputStr != NULL)
-			(*outputStr) = sf_getTheStr(testOutput);
-		else if (sf_getFile(testOutput) != NULL)
-		{
-			if (fclose(sf_getFile(testOutput)) != 0)
-			{
-				charsAvailForFilename = (int) (MAXLINE - strlen(outfileName));
-				messageFormat = "Unable to close output file \"%.*s\".\n";
-				sprintf(messageContents, messageFormat, charsAvailForFilename, outfileName);
-				ErrorMessage(messageContents);
-				Result = NOTOK;
-			}
-		}
-	}
-	else
-	{
-		if (outputStr != NULL && (*outputStr) != NULL)
-		{
-			free((*outputStr));
-			(*outputStr) = NULL;
-		}
-		else if (sf_getFile(testOutput) != NULL)
-		{
-			if (fclose(sf_getFile(testOutput)) != 0)
-			{
-				charsAvailForFilename = (int) (MAXLINE - strlen(outfileName));
-				messageFormat = "Unable to close output file \"%.*s\".\n";
-				sprintf(messageContents, messageFormat, charsAvailForFilename, outfileName);
-				ErrorMessage(messageContents);
-				Result = NOTOK;
-			}
-			else
-			{
-				if (remove(outfileName) != 0)
-				{
-					charsAvailForFilename = (int) (MAXLINE - strlen(outfileName));
-					messageFormat = "Unable to remove output file \"%.*s\" after error.\n";
-					sprintf(messageContents, messageFormat, charsAvailForFilename, outfileName);
-					ErrorMessage(messageContents);
-					Result = NOTOK;
-				}
-			}
-			
-		}
+			(*outputStr) = sf_takeTheStr(testOutput);
 	}
 
 	free(headerStr);
@@ -514,6 +469,15 @@ int outputTestAllGraphsResults(char command, testAllStatsP stats, char * infileN
 
 	free(resultsStr);
 	resultsStr = NULL;
+
+	if (sf_closeFile(testOutput) != OK)
+	{
+		charsAvailForFilename = (int) (MAXLINE - strlen(outfileName));
+		messageFormat = "Unable to close output file \"%.*s\".\n";
+		sprintf(messageContents, messageFormat, charsAvailForFilename, outfileName);
+		ErrorMessage(messageContents);
+		Result = NOTOK;
+	}
 
 	sf_Free(&testOutput);
 
--- a/c/planarityTestGraphFunctionality.c
+++ b/c/planarityTestGraphFunctionality.c
@@ -56,6 +56,7 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
 	int charsAvailForFilename = 0;
 	char *messageFormat = NULL;
 	char messageContents[MAXLINE + 1];
+	messageContents[MAXLINE] = '\0';
 
 	graphP theGraph;
 
@@ -151,12 +152,23 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
 					// Stop the timer
 					platform_GetTime(end);
 					stats.duration = platform_GetDuration(start, end);
-					sprintf(messageContents, "\nDone testing all graphs (%.3lf seconds).\n", stats.duration);
-					Message(messageContents);
+
+					if (Result != OK && Result != NONEMBEDDABLE)
+					{
+						messageFormat = "\nEncountered error while running command '%c' on all graphs in \"%.*s\".\n";
+						charsAvailForFilename = (int) (MAXLINE - strlen(messageFormat));
+						sprintf(messageContents, messageFormat, command, charsAvailForFilename, infileName);
+						ErrorMessage(messageContents);
+					}
+					else
+					{
+						sprintf(messageContents, "\nDone testing all graphs (%.3lf seconds).\n", stats.duration);
+						Message(messageContents);
+					}
 
 					if (outputTestAllGraphsResults(command, &stats, infileName, outfileName, outputStr) != OK)
 					{
-						messageFormat = "Error outputting results running command '%c' on graphs in \"%.*s\".\n";
+						messageFormat = "Error outputting results running command '%c' on all graphs in \"%.*s\".\n";
 						charsAvailForFilename = (int) (MAXLINE - strlen(messageFormat));
 						sprintf(messageContents, messageFormat, command, charsAvailForFilename, infileName);
 						ErrorMessage(messageContents);
@@ -223,6 +235,7 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
 
 	char *messageFormat = NULL;
 	char messageContents[MAXLINE + 1];
+	messageContents[MAXLINE] = '\0';
 
 	graphP copyOfOrigGraph = NULL;
 	int embedFlags = GetEmbedFlags(command);
@@ -347,7 +360,7 @@ int outputTestAllGraphsResults(char command, testAllStatsP stats, char * infileN
 
 	char *resultsStr = (char *) malloc(
 										(
-											3 +_getNumCharsToReprInt(stats->numGraphsRead) +
+											3 + _getNumCharsToReprInt(stats->numGraphsRead) +
 											1 + _getNumCharsToReprInt(stats->numOK) +
 											1 + _getNumCharsToReprInt(stats->numNONEMBEDDABLE)+
 											1 + 8 + // either ERROR or SUCCESS, so the longer of which is 7 + 1 chars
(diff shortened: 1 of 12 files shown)


c/planarityTestGraphFunctionality.c
```diff
@@ -109,8 +129,8 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
 						if (outputFileP == NULL)
 						{
 							charsAvailForFilename = (int) (MAXLINE - strlen(outfileName));
-							errorMessageFormatStr = "Unable to open file \"%.*s\" for output.\n";
-							sprintf(messageContents, errorMessageFormatStr, charsAvailForFilename, outfileName);
+							messageFormat = "Unable to open file \"%.*s\" for output.\n";
```
john-boyer-phd, line 132: starting with the `strOrFileP testOuptut` declaration and on down, all output code should be pull out of this function and put into a separate function that outputs the results of test all graphs. It should receive the stats object and the name of the output file, and do all the rest of the work related to outputting, testing for failures related to outputting, and returning NOTOK related to said failures, or OK on success.

c/planarityTestGraphFunctionality.c
```diff
@@ -140,36 +160,100 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
 					char *finalSlash = strrchr(infileName, FILE_DELIMITER);
 					char *infileBasename = finalSlash ? (finalSlash + 1) : infileName;
 
-					char *headerStr = (char *) malloc((strlen(infileBasename) + 3) * sizeof(char));
+					char *headerFormat = "FILENAME=\"%s\" DURATION=\"%.3lf\"\n";
```
john-boyer-phd, line 163: Move into the new output function.

c/planarityTestGraphFunctionality.c
```diff
@@ -140,36 +160,100 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
...
+					char command = commandString[1];
+					Result = testAllGraphs(theGraph, command, inputStr, stats);
+
+					// Stop the timer
+					platform_GetTime(end);
+					duration = platform_GetDuration(start, end);
+					sprintf(messageContents, "\nDone testing all graphs (%.3lf seconds).\n", duration);
+					Message(messageContents);
+
+					sprintf(headerStr, headerFormat, infileBasename, duration);
```
john-boyer-phd, line 212: Call the output function here. Be careful to preserve a NOTOK result from `testAllGraphs()`. In other words, call as `if (myOutputFunc(outfileName, stats) != OK) Result = NOTOK;` rather than just `Result = myOutputFunc(...)`.

c/planarityTestGraphFunctionality.c
```diff
@@ -140,36 +160,100 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
...
+															3
+														) * sizeof(char));
+					if (resultsStr == NULL)
 					{
-						errorMessageFormatStr = "Unable to perform algorithm corresponding to command '%c' to graph(s).\n";
-						sprintf(messageContents, errorMessageFormatStr, commandString[1]);
-						ErrorMessage(messageContents);
+						ErrorMessage("Unable allocate memory for results string.\n");
+
+						free(inputStr);
```
john-boyer-phd, line 229: This `free(inputStr)` and the `gp_Free`  will still be the responsibility of the `TestGraphFunctionality` not the new output function. You should just be able to preserve the result of the output func then do the free and gp_Free once, whether the result is or is not OK.

c/planarityTestGraphFunctionality.c
```diff
@@ -140,36 +160,100 @@ int TestGraphFunctionality(char *commandString, char *infileName, char *inputStr
...
+						inputStr = NULL;
+
+						free(stats);
+						stats = NULL;
+
+						gp_Free(&theGraph);
+
+						sf_Free(&testOutput);
+
+						return NOTOK;
```
john-boyer-phd, line 239: You have return NOTOK here, but below at the end there are some return -1 instances that seem like they should be return NOTOK as well?

c/planarityTestGraphFunctionality.c
```diff
@@ -226,16 +310,16 @@ int transformString(graphP theGraph, char *inputStr)
 	TEST ALL GRAPHS IN .G6
 */
 
-int testAllGraphs(graphP theGraph, char command, char *inputStr, strOrFileP testOutput)
+int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP stats)
```
john-boyer-phd, line 313: I really like that you did this as a struct so we can add other things later, if we choose to, like some of the extra stats we compute when testing randomly generated graphs.

c/planarityTestGraphFunctionality.c
```diff
@@ -226,16 +310,16 @@ int transformString(graphP theGraph, char *inputStr)
...
 	int exitCode = OK;
 
-	char *errorMessageFormatStr = NULL;
+	char *messageFormat = NULL;
 	char messageContents[MAXLINE + 1];
 
 	graphP copyOfOrigGraph = NULL;
 	int embedFlags = GetEmbedFlags(command);
-	int numGraphsRead = 0, numOK = 0, numNONEMBEDDABLE = 0;
+	int numGraphsRead = 0, numOK = 0, numNONEMBEDDABLE = 0, errorFlag = 0;
```
john-boyer-phd, line 322: errorFlag = FALSE

john-boyer-phd (changes requested):
Looks really good, some minor refactoring and tweaks.

c/planarityTestGraphFunctionality.c
wbkboyer: I think you're right, that all instances of `return -1` in `TestGraphFunctionality` should be changed to `return NOTOK`, since:
* `commandLine` of `planarityCommandLine.c` (which calls `callTestGraphFunctionality`, which calls `TestGraphFunctionality`) expects `OK`, `NONEMBEDDABLE`, or `NOTOK` from all the other functions it calls
* `runGraphTransformationTest` handles the conversion from `OK`, `NONEMBEDDABLE`, and `NOTOK` by `return (Result == OK) ? 0 : -1;` except in one case: when `ConstructTransformationExpectedResultFilename` doesn't return `OK` or `expectedOutfileName` is still `NULL`, I changed `return NOTOK` to `return -1`
  * This is because `runQuickRegressionTests` of `planarityCommandLine.c` calls `runSpecificGraphTests` and returns `NOTOK` if its return value is `< 0`; `runSpecificGraphTests` calls `runGraphTransformationTest`, and if its return value is `< 0`, sets `retVal` to `-1`
* `TransformMenu` and `TestMenu` of `planarity.c` expects `OK`, `NONEMBEDDABLE`, or `NOTOK`

c/planarityTestGraphFunctionality.c
wbkboyer: Whoops, almost forgot this one! Addressed in latest commit.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
 
-int _getNumCharsToReprInt(int theNum)
+int outputTestAllGraphsResults(char command, testAllStatsP stats, char * infileName, char *outfileName, char **outputStr)
 {
-	int numCharsRequired = 1;
+	int Result = OK;
 
-	while(theNum /= 10)
-		numCharsRequired++;
+	FILE *outputFileP;
```
john-boyer-phd, line 325: This caught my eye as an uninitialized pointer. Normally, one would assign to NULL, but it is better for you to just declare it later, as mentioned below.  So, just delete this line.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+
+		if (outputFileP != NULL)
+		{
+			fclose(outputFileP);
+			outputFileP = NULL;
+		}
+
+		if ((*outputStr) != NULL)
+		{
+			free(outputStr);
```
john-boyer-phd, line 373: This will crash because outputStr is a double pointer. It is necessary to instead to `free(*outputStr)` then assign *outputStr to NULL.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+
+	char *headerFormat = "FILENAME=\"%s\" DURATION=\"%.3lf\"\n";
+	char *headerStr = (char *) malloc(
+										(
+											strlen(headerFormat) +
+											strlen(infileBasename) +
+											strlen("-1.7976931348623158e+308") + // -DBL_MAX from float.h
+											3
+										) * sizeof(char));
+	if (headerStr == NULL)
```
john-boyer-phd, line 391: It is probably best (safest) to do the mallocs for headerStr and resultStr first. If either fail you can free them up and bolt. Then, when handling failure to create the testOutput strOrFile, you'd have to free these string, since you allocated them first. However, the benefit is that you would have fewer cases where you have to do the testOutput freeing logic. The stuff in the braces below lacks some code to deal with freeing the memory or the closing the file within testOutput.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+		if (outputFileP == NULL)
+		{
+			charsAvailForFilename = (int) (MAXLINE - strlen(outfileName));
+			messageFormat = "Unable to open file \"%.*s\" for output.\n";
+			sprintf(messageContents, messageFormat, charsAvailForFilename, outfileName);
+			ErrorMessage(messageContents);
+			return NOTOK;
+		}
+		
+		testOutput = sf_New(outputFileP, NULL);
```
john-boyer-phd, line 345: If testOutput is NULL after this line, you should fclose outputFileP right away.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+	if (outfileName != NULL)
+	{
+		outputFileP = fopen(outfileName, "w");
+		if (outputFileP == NULL)
+		{
+			charsAvailForFilename = (int) (MAXLINE - strlen(outfileName));
+			messageFormat = "Unable to open file \"%.*s\" for output.\n";
+			sprintf(messageContents, messageFormat, charsAvailForFilename, outfileName);
+			ErrorMessage(messageContents);
+			return NOTOK;
```
john-boyer-phd, line 342: Once you move the headerStr and resultStr mallocs up, you'd have to free them here.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+	}
+	else
+	{
+		if ((*outputStr) == NULL)
+			(*outputStr) = (char *) malloc(1 * sizeof(char));
+		
+		if ((*outputStr) == NULL)
+		{
+			ErrorMessage("Unable to allocate memory for outputStr.\n");
+			return NOTOK;
```
john-boyer-phd, line 355: Once you move the headerStr and resultStr mallocs up, you'd have to free them here, too.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+	{
+		ErrorMessage("Unable to set up string-or-file container for test output.\n");
+
+		if (outputFileP != NULL)
+		{
+			fclose(outputFileP);
+			outputFileP = NULL;
+		}
+
+		if ((*outputStr) != NULL)
```
john-boyer-phd, line 371: This seems like it would have the potential to crash because the double pointer outputStr could be NULL if an output file name was given. That being said, if you free the output string above (if the second sf_New fails), then you won't have this if test because you'll be in the right branch of code.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+		if ((*outputStr) == NULL)
+			(*outputStr) = (char *) malloc(1 * sizeof(char));
+		
+		if ((*outputStr) == NULL)
+		{
+			ErrorMessage("Unable to allocate memory for outputStr.\n");
+			return NOTOK;
+		}
+
+		testOutput = sf_New(NULL, (*outputStr));
```
john-boyer-phd, line 358: test next line after here whether testOutput is NULL, and if so then you can free the *outputStr right away

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+	}
+
+	if (testOutput == NULL)
+	{
+		ErrorMessage("Unable to set up string-or-file container for test output.\n");
+
+		if (outputFileP != NULL)
```
john-boyer-phd, line 365: This moves up to just after the sf_New call. That being said, this is now a place where the headerStr and resultStr would have to be freed, once you move their mallocs to the top.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
+											3 +_getNumCharsToReprInt(stats->numGraphsRead) +
+											1 + _getNumCharsToReprInt(stats->numOK) +
+											1 + _getNumCharsToReprInt(stats->numNONEMBEDDABLE)+
+											1 + 8 + // either ERROR or SUCCESS, so the longer of which is 7 + 1 chars
+											3
+										) * sizeof(char));
+	if (resultsStr == NULL)
+	{
+		ErrorMessage("Unable allocate memory for results string.\n");
+		sf_Free(&testOutput);
```
john-boyer-phd, line 415: This also doesn't fclose the output file or free the output string in the testOutput, but once the resultStr malloc and NULL test move up, the only thing you will have made yet is headerStr, so that would need to be freed here.

c/planarityTestGraphFunctionality.c
```diff
@@ -408,15 +315,120 @@ int testAllGraphs(graphP theGraph, char command, char *inputStr, testAllStatsP s
...
\ No newline at end of file
+	int charsAvailForFilename = 0;
+	char *messageFormat = NULL;
+	char messageContents[MAXLINE + 1];
+
+	strOrFileP testOutput = NULL;
+
+	if (outfileName != NULL)
+	{
+		outputFileP = fopen(outfileName, "w");
```
john-boyer-phd, line 335: Prepend `FILE *` here so that outputFileP doesn't exist outside this scope.

john-boyer-phd (changes requested):
Looks good, just a bit more refactoring to deal with error case memory leaks and possible crashes, and just to make the error cases shorter and easier to code.

c/planarityTestGraphFunctionality.c
wbkboyer: The problem I was trying to avoid was that if `sf_New` fails, I need to be able to close the `FILE` pointer... If it's declared inside the scope of the `if` statement, then if `testOutput` is `NULL`, I'm worried I can no longer access that pointer :(

c/planarityTestGraphFunctionality.c
wbkboyer: I have a silly question: what if someone passes the address to memory already allocated for `outputStr` (i.e. some other function "owns" the memory)? Should I error out if `outputStr != NULL && (*outputStr) != NULL`?

john-boyer-phd (approved):
This is looking great. Excellent improvements to the pre-existing code also, to get the best overall result.

Issue #62, opened 2024-04-27
Issue 62 - Write six separate scripts to sequentially invoke the release version of `planarity` on all 55 11-vertex graphs for each algorithm (pdo234)


wbkboyer:
Write six separate scripts for each algorithm (pdo234) to invoke the release version of `planarity` in sequence to test all graphs with the script's algorithm for each of the 55 11-vertex `graph6` files. Make the output file names indicate the input filenames followed by a suffix of `.C.out.txt` where C is the algorithm identifier (p, d, o, 2, 3, or 4). Check the six scripts into the `Test Support` directory. In six separate shells, run each of the six scripts on the 11-vertex graphs.

john-boyer-phd:
This was closed by PR #63 but my squash merge comment didn't correctly trigger the close operation, so closing manually.

PR #63, opened 2024-04-29
Issues 61 & 62 - Scripts to generate graphs and to execute release version of planarity

### LINKED ISSUE #61: Issue 61 - Write a script that invoke Nauty `geng` to generate all 11-vertex graphs using separate files for each of the possible numbers of edges
Write a script that invoke Nauty `geng` to generate all 11-vertex graphs using separate files for each of the possible numbers of edges (55). Name the files to include the number of vertices and to consistently with 2-digit values for the number of edges so that sorting files lexicographically will also put them in order by number of edges.  For example, `n11.m20.g6` could be the name of the `graph6` file containing all 11-vertex 20-edge graphs generated by `geng`. Check the script into a `Test Support` directory sibling of the `c` directory (just the script, not the 55 `graph6` files).

wbkboyer:
Resolves #61 and #62

## Added
* `TestSupport/generate_graphs.bat` - optionally accepts the graph order (defaults to 11) and uses `START` keyword to spawn processes to generate graphs of each edge count from 1 up until `(N * (N-1))/2` with output piped into files in a subdirectory of the `graphOutputDir` 
  * Run as follows (where `N` is the graph order):
  ```
  > cd cd git\edge-addition-planarity-suite-fork\TestSupport
  > .\generate_graphs.bat N
  ```
* `TestSupport/graphGenerationOrchestrator.py` - An alternate implementation in Python that leverages the [`multiprocessing`](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing) and [`subprocess`](https://docs.python.org/3/library/subprocess.html) modules to create a pool of threads (specifically the number of threads is how many cores the PC on which you are running the script has) and to run a blocking process on each thread to execute `geng` for the given graph order. Requires Python 3.5+, but tested using Python 3.12
  * Run as follows:
  ```
  > cd cd git\edge-addition-planarity-suite-fork\TestSupport
  > python graphGenerationOrchestrator.py -n 11 "C:\\Users\\wbkbo\\git\\nauty\\geng.exe" "E:\\graphs\\11"
  ```
* `TestSupport/planarity_runner.bat` - optionally accepts the graph order (defaults to 11), and optionally accepts the algorithm command to be run (defaults to `p`; no error checking), and sequentially executes the release version of `planarity` on the graphs in the `graphFilesDir`
  * Run as follows (where `N` is the graph order and `C` is one of `pdo234`):
  ```
  > cd cd git\edge-addition-planarity-suite-fork\TestSupport
  > .\planarity-runner.bat N C
  ```

## Updated
* `c/planarityTestGraphFunctionality.c` - Previously, if an error was encountered before running the algorithm on the graphs, the output file would still indicate `SUCCESS`. Now, the `testAllStats->errorFlag` is set if an error was encountered before actually running the tests.


c/planarityTestGraphFunctionality.c
```diff
@@ -240,6 +240,7 @@ int testAllGraphs(graphP theGraph, char command, FILE *infile, testAllStatsP sta
 	if (Result != OK)
 	{
 		ErrorMessage("Unable to allocate G6ReadIterator.\n");
+		stats->errorFlag = TRUE;
```
john-boyer-phd, line 243: Good catches.

john-boyer-phd (approved):
Looks great. Now that you've switched over to reading from file, it makes me wonder if you really need to use CALL rather than START (and, in the python version, defaulting to one thread per logical processor). We'll have to give it a test at some point.

PR #70, opened 2024-05-08
Issue 69 - Ensure edge returned by `_getNextEdge()` has endpoints in the upper-triangular order

### LINKED ISSUE #69: Issue 69 - `G6WriteIterator` - graph generated using `gp_CreateRandomGraphEx()` has incorrect edges in `.g6` format
When creating maximal planar graphs with `N = 11` (and therefore `M = DEFAULT_EDGE_LIMIT * N - 6 = 3 * 11 - 6 = 27`) to test `SpecificGraph()` (see #67 ), I chose to write to `.g6` so that I could cross-reference with the result returned by `testAllGraphs()`. However, I observed that the `.g6` output file reported as nonplanar (both by `SpecificGraph()` and `testAllGraphs()`). Then, I chose to write to Adjacency List, and the graph was reported as planar (as expected). I then converted that Adjacency List representation to `.g6`, but this representation didn't match the `.g6` initially produced.

<details>
<summary>Output from initial tests</summary>
The output to `.g6` format:
```
>>graph6<<J~}LEApoC?_
```

```
N=11
0: 1 2 3 4 5 6 7 8 9 10 -1
1: 0 2 3 4 7 9 -1
2: 0 1 3 4 6 8 -1
3: 0 1 2 4 8 -1
4: 0 1 2 3 5 -1
5: 0 4 -1
6: 0 2 -1
7: 0 1 8 -1
8: 0 2 3 7 -1
9: 0 1 10 -1
10: 0 9 -1
```

The Adjacency List format of the same graph:
```
N=11
1: 2 6 5 4 7 8 11 10 9 3 0
2: 3 4 5 6 1 0
3: 1 9 8 7 4 2 0
4: 1 5 2 3 7 0
5: 1 6 2 4 0
6: 1 2 5 0
7: 1 4 3 8 0
8: 1 7 3 9 10 11 0
9: 1 10 8 3 0
10: 1 11 8 9 0
11: 1 8 10 0
```

But when I transform the Adjacency list representation to `.g6` format, the resulting `.g6` output doesn't match:
```
>>graph6<<J~vLdE`_[A_
```

Output for `.g6` representation of the same graph:
```
The graph "C:\Users\wbkbo\git\edge-addition-planarity-suite-fork\maxPlanarG6WriterTests\n11.MAXPLANAR.g6" is not planar.
Algorithm 'PlanarEmbed' executed in 0.000 seconds.

...

FILENAME="n11.MAXPLANAR.g6" DURATION="0.000"
-p 1 0 1 SUCCESS
```

Output for Adjacency List:

```
The graph "C:\Users\wbkbo\git\edge-addition-planarity-suite-fork\maxPlanarG6WriterTests\n11.MAXPLANAR.AdjList.out.txt" is planar.
Algorithm 'PlanarEmbed' executed in 0.000 seconds.
```

</details>

At this point, I decided to see if I could reproduce the issue of the mismatched representations with a smaller `N`, which occurred with `N = 5` (and `M = DEFAULT_EDGE_LIMIT * N - 6 = 3 * 5 - 6 = 9`), and was able to reproduce the issue: I converted the initial `.g6` file to Adjacency List and compared it with the "good" Adjacency List, and noticed that there were edges missing for this specific example. I also manually confirmed that the `.g6` representation of the "good" Adjacency List had the expected edges, that converting the original Adjacency list to `.g6` back to Adjacency List produced the original representation (except that `.g6` format is 0-based, and so the vertex labelings were off by one from the original), and that there was no memory issue writing the original graph to `.g6` multiple times (same incorrect output, so the issue didn't appear to result from stomping over memory).

<details>
<summary>Reproducing issue for N = 5</summary>

The `.g6` file output:
```
>>graph6<<D~c
```

Converted to Adjacency List:
```
N=5
0: 1 2 3 4 -1
1: 0 2 3 -1
2: 0 1 3 -1
3: 0 1 2 4 -1
4: 0 3 -1
```

The Adjacency List representation of the same graph:
```
N=5
1: 5 4 2 3 0
2: 5 4 3 1 0
3: 4 2 1 0
4: 5 1 3 2 0
5: 2 1 4 0
```

Converted to `.g6`:
```
>>graph6<<D~s
```

Note that the edge `(2, 5)` is missing in the original `.g6` representation of the graph.
</details>

wbkboyer:
Resolves #69 

When we try to use the `columnOffsets` array to determine which byte and which bit in that byte must be set in the bit representation to indicate that an edge exists, the assumption is that the edges are in upper-triangular form (i.e. `u < v`). Since we're dealing with simple graphs, if `(u, v)` exists in the graph, then `(v, u)` exists, and so we can represent this edge with `(min(u, v), max(u, v))`.

When a graph is generated by `gp_CreateRandomGraphEx()`, the latter `(DEFAULT_EDGE_LIMIT * N - 6) - (N - 1)` edges `e` may be added in such a way that they appear in the edge list `E` with `E[e].neighbor` as the larger of the two endpoints of `e` and the twin arc `E[e + 1].neighbor` as the smaller. Therefore, in `_encodeAdjMatAsG6()`, when we `_getNextEdge()`, we need to ensure the endpoints are representative of the edge in the upper triangular matrix.


c/g6-write-iterator.c
```diff
@@ -338,6 +338,7 @@ int _encodeAdjMatAsG6(G6WriteIterator *pG6WriteIterator) {
 
 	int charOffset = 0;
 	int bitPositionPower = 0;
+	int tempVert = NIL;
```
john-boyer-phd, line 341: Remove this because you can just declare it inside the curly braces below.

c/g6-write-iterator.c
```diff
@@ -346,6 +347,17 @@ int _encodeAdjMatAsG6(G6WriteIterator *pG6WriteIterator) {
 		u -=  gp_GetFirstVertex(theGraph);
 		v -=  gp_GetFirstVertex(theGraph);
 
+		// The columnOffset machinery assumes that we are traversing the edges represented in 
+		// the upper-triangular matrix. Since we are dealing with simple graphs, if (v, u)
+		// exists, then (u, v) exists, and so the edge is indicated by a 1 in row = min(u, v)
+		// and col = max(u, v) in the upper-triangular adjacency matrix.
+		if (v < u)
+		{
+			tempVert = v;
```
john-boyer-phd, line 356: Just start this line with `int`

c/planarityCommandLine.c
```diff
@@ -18,6 +18,7 @@ int callSpecificGraph(int argc, char *argv[]);
 int callRandomMaxPlanarGraph(int argc, char *argv[]);
 int callRandomNonplanarGraph(int argc, char *argv[]);
 int callTestGraphFunctionality(int argc, char *argv[]);
+int callGenerateRandomMaxPlanarGraph(int argc, char *argv[]);
```
john-boyer-phd, line 21: Delete.

c/planarityCommandLine.c
```diff
@@ -58,6 +59,9 @@ int commandLine(int argc, char *argv[])
 	else if (strncmp(argv[1], "-t", 2) == 0)
 		Result = callTestGraphFunctionality(argc, argv);
 
+	else if (strcmp(argv[1], "-m") == 0)
```
john-boyer-phd, line 62: This can be deleted because it is adding a feature that we kind of already via use  of the command line. It'd be nicer to add the feature (along with another related one) at a different time. When this is added, some update of the help message, the man page, etc. has to happen.

c/planarityCommandLine.c
```diff
@@ -722,3 +726,31 @@ int callTestGraphFunctionality(int argc, char *argv[])
 	// We don't need to capture whether output is 0- or 1-based, so zeroBasedOutputFlag arg is NULL
 	return TestGraphFunctionality(commandString, infileName, NULL, NULL, outfileName, NULL);
 }
+
```
john-boyer-phd, line 729: This was very helpful for debugging but can be removed now.

c/planarityRandomGraphs.c
```diff
@@ -469,3 +469,268 @@ char saveEdgeListFormat;
 	FlushConsole(stdout);
 	return Result;
 }
+
```
john-boyer-phd, line 472: This was very helpful for debugging but can be removed now.

john-boyer-phd (changes requested):
Great bug fix, mainly just requests to remove the extra feature that was handy for finding and fixing the bug (and which could be added via a different command line at a future time).

Issue #42, opened 2024-04-05
EPIC: Investigate Operation of K_{3,3} Search Implementation


john-boyer-phd:
Comparing results of current code to results from version 2.0 code, there seem to be 100 11-vertex 20-edge graph that are now reported to be `K_{3,3}`-free but didn't used to be. For `K_{3,3}` graphs, the old table reported 3,499,929 but now we report 3,500,029 (the numbers containing `K_{3,3}` are similarly adjusted down by 100 in the current code). So, the code used to find a `K_{3,3}` and verify the result in those 100 graphs, and now it doesn't. 

This will involve several steps:

- [x] Get the v2.0 tag of the code into a separate private repo.
- [x] Add VS Code ability (for convenience)
- [x] Add the G6 file reading and writing and the Test All functionality.
- [x] See if this old code can reproduce the numbers from the V2.0 table.
- [x] If so, amend it to write the `OK` and `NONEMBEDDABLE` results as a string of 0 and 1 values for all 11-vertex 20-edge graphs.
- [x] In a branch of the current code, amend the test all function to read the results string from the old code and compare to results obtained by the current code.
- [x] For every OK (`K_{3,3}`-free) result in the new code that is not an OK in the results string, increment a counter. Also, if there are any `NONEMBEDDABLE` results in the new code that correlate with an OK in the results string, then increment a second counter.  See whether the counter differences match the total difference of 100. (The total differences matches but the separate differences between `OK` counts and `NONEMBEDDABLE` counts is much higher than expected because `makeg` and `geng` produce different graphs).
  * See [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/42#issuecomment-2125583338)
- [x] Determine whether `makeg` or `geng` is erroneously producing any duplicate graphs. (Both were found to be generating unique graphs, just some different graphs).
- [x] Develop a way to isolate the graphs that are different between two `g6` files, namely one representing the output of `makeg` and the other representing the output of `geng`.
  * See [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/42#issuecomment-2125623889)
  * See also https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/72
[...]

PR #74, opened 2024-05-24
Issue 73 - `TestSupport` Python script style overhaul

wbkboyer:
## Updates
* `TestSupport/generate_graphs.bat` now pipes output from `nauty`'s `geng` to files with names like `nXX.mXX.g6` 
* Renamed:
  * `TestSupport/graphGenerationOrchestrator.py` to `TestSupport/graph_generation_orchestrator.py`
  * `TestSupport/planarityOrchestrator.py` to `TestSupport/planarity_orchestrator.py`
  * `TestSupport/testTableGenerator.py` to `TestSupport/test_table_generator.py`
* `TestSupport/planarity_orchestrator.py` now correctly performs tests on `.g6` files starting with `M = 0` (previously was starting the `range()` for edge counts at 1)
* Added `#!/usr/bin/env python` shebang to all scripts so that if you `chmod u+x` these scripts and choose to directly run them from the command line, they'll run using the python interpreter that is currently active
* explicitly set `__all__ = []` to effectively make all members of these scripts private: if someone tries to `from MODULENAME import *`, nothing will be exported
* Updated all Python scripts' `argparse.ArgumentParser()` initialization to give a more suitable `usage` (I was previously overriding `prog` with the title of the program, which messed with the default `usage` message) and more informative `description` (using `formatter_class=argparse.RawTextHelpFormatter` to make sure newlines are respected in multi-line `description`)
* Added documentation to explain what is being done by these tools, most especially the workings of the `test_table_generator.py`
* Added custom exceptions in `test_table_generator.py` to differentiate between issues with paths to input files vs. issues with input file contents


TestSupport/graph_generation_orchestrator.py
```diff
@@ -0,0 +1,65 @@
...
+        geng_path: Path to the nauty geng executable
+        order: Desired number of vertices
+        num_edges: Desired number of edges
+        output_dir: Directory to which you wish to write the resulting .g6 file
+    """
+    filename = Path.joinpath(output_dir, f'n{order}.m{num_edges}.g6')
+    with open(filename, "w") as outfile:
+        subprocess.run(
+            [f'{geng_path}', f'{order}', f'{num_edges}:{num_edges}'],
+            stdout=outfile, stderr=subprocess.PIPE, shell=True)
```
wbkboyer, line 33: Happened to test this script on MacOS and found that the args weren't being passed as expected... However, when I allowed default `shell=False`, `subprocess.run()` passes the command line args as expected.

I'll have to test to see if default `shell=False` works on MacOS and Windows, and if not, then to use `shell=True` but to `[...].join()` the args into a string :(

TestSupport/planarity_orchestrator.py
```diff
@@ -0,0 +1,86 @@
...
+    outfile_path = Path.joinpath(
+        output_dir, f'{command}', f'n{order}.m{num_edges}.{command}.out.txt')
+    subprocess.run(
+        [
+            f'{planarity_path}',
+            '-t', f'-{command}',
+            f'{infile_path}',
+            f'{outfile_path}'
+        ],
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
```
wbkboyer, line 45: Same as in [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/74/files#r1614178925)

wbkboyer:
Received confirmation that this is safe to merge from @john-boyer-phd; this is a good starting point, since these scripts are currently in use to inform the investigation for #42. However, subsequent automation work has surfaced some necessary structural changes in some of these scripts, so it's best to merge these changes and then continue development.

PR #78, opened 2024-06-24
Issue 77 - Automate edge-deletion analysis

### LINKED ISSUE #77: Write a program to perform edge deletion `K_{3,3}` analyses on a given graph.
The desired program should attempt to find a `K_{3,3}` homeomorph in a given graph by calling on the `planarity` program's `K_{3,3}` search on the graph and on each subgraph that can be formed by deleting one edge from the original input graph. Although the program could be efficiently written using the graph planarity C API functions, the program should be written as a separate and independent Python program, to be consistent with the other test support programs and to defer to a future epic how to write programs that pull in and use `libPlanarity`.

The program should receive a `.g6` file on the command line. For each graph in the input file the program should:
- convert it to adjacency list format using the `planarity` program's transformer;
- invoke the `planarity` program's `K_{3,3}` search. Report whether the `K_{3,3}` search found (NONEMBEDDABLE) or did not find (OK) a subgraph homeomorphic to `K_{3,3}`
- if the result was an `OK` (did not find subgraph homeomorphic to `K_{3,3}`), then invoke `planarity` program's planar graph embedder/Kuratowski subgraph isolator.
- If the graph is reported to be planar, indicate that the graph does not contain a subgraph homeomorphic to `K_{3,3}` because it is planar, so the `K_{3,3}` search correctly reported no subgraph homeomorphic to `K_{3,3}`.
- if the graph is reported to be non-planar, and the output graph contains a subgraph homeomorphic to `K_{3,3}`, then report that the graph contains a subgraph homeomorphic to `K_{3,3}` that was not found by the `K_{3,3}` search.
- if the graph is reported to be non-planar, and the output graph contains a subgraph homeomorphic to `K_5` then perform the edge deletion analysis described below. 

To distinguish between a non-planar output that contains a subgraph homeomorphic to `K_{3,3}` versus a subgraph homeomorphic to `K_5` it suffices to read the output graph into an array of arrays of the adjacency lists, then determine whether the maximum adjacency list length (i.e., the maximum vertex degree) is 3 for a `K_{3,3}` homeomorph or 4 for a `K_5` homeomorph.

To perform the edge deletion analysis:
- read the adjacency list version of the input graph into an array of arrays representing the adjacency lists of the vertices
- For each array `v` representing a vertex, iterate through each array element `e`,
  - if `e` is greater than `v` then  
    - make a copy of the graph
    - remove the edge that `e` represents (by removing `e` from vertex array `v` and remove `v` from vertex array `e`)
    - output the graph copy less the deleted edge to a file with a name that includes the input filename and the edge deleted. 
    - Invoke the `planarity` program's `K_{3,3}` search on outputted graph
    - If the `K_{3,3}` search indicated that a `K_{3,3}` homeomorph was found, then report that a `K_{3,3}` homeomorph was found then give the output filename (you may report more than one `K_{3,3}` during this loop)   
- After the loop, if no `K_{3,3}` homeomorph was found, then report that no `K_{3,3}` homeomorph was found by the edge deletion analysis.

wbkboyer:
Resolves #77 

## Added
* `TestSupport/planaritytesting/`
  * `graph.py` - Implemented `Graph` class to wrap a graph's adjacency list representation. Since this adjacency list representation is a list of lists of integers, (i.e. a list of mutable types), I needed to implement the `__deepcopy__()` method so that during the edge-deletion analysis when we `delete_edge()` from a copy of a graph, it doesn't change the original graph's `graph_adj_list_repr`.
  * `edge_deletion_analysis.py` - when given a `.g6` input file, iterates over the lines of the input file and writes each graph to a new file, which is then transformed from `.g6` to adjacency list format. Then, if `K_{3, 3}` search reports the graph is `K_{3, 3}` free, checks whether core planarity reports that the graph is planar; if not and the obstruction found is homeomorphic to `K_{3, 3}`, then we've found one of the problematic graphs. If the obstruction found is homeomorphic to `K_5`, we perform the edge-deletion analysis steps to determine if the graph contains a `K_{3, 3}` that was missed by the `K_{3, 3}` search graph algorithm extension (with a high degree of certainty).
    * _**N.B.**_ It may be the case that the graph-minus-one-edge still has the same underlying structure that causes the issue with `K_{3, 3}` search on the original graph, so even this additional test is not a 100% guarantee of finding the `K_{3, 3}` that we should have found in the original graph.

## Updated
* `TestSupport/planaritytesting/planaritytesting_utils.py` - previously `planarity_constants.py`; now contains other helper methods and constants used throughout the package

Many additional changes were made to improve style and error handling throughout the project. While this means that this PR will not entail an atomic commit when it gets squash-merged, I chose to do this work to keep all modules aligned with the patterns introduced by the edge-deletion analysis tool.

These changes included:
* Adding docstrings to functions, classes, modules, and `__init__.py` for package
* Adding typehints for parameters and return values
* Using more descriptive exception names (i.e. not using `except BaseException as e` but rather `except BaseException as more_descriptive_name_error`)
* Using lazy formatting for `logger` messages (prevents evaluating strings that might never be used)
* Using [`typing.Optional`](https://docs.python.org/3/library/typing.html#typing.Optional) so that allowing optional types is explicit (see [pep-484](https://peps.python.org/pep-0484/))
* Explicitly giving `encoding="utf-8"` when `open()`ing a file (see [unspecified-encoding / W1514](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/unspecified-encoding.html))
* Explicitly specifying when `check=False` (rather than relying on default; see [subprocess-run-check / W1510](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/subprocess-run-check.html))
* For every module that may be run as a script except for `test_table_generator.py`, directly pass in `argparse` parsed `args` using keyword assignment instead of assigning to a variable with the same name (see [redefined-outer-name / W0621](https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/redefined-outer-name.html))
* For every module that may be run as a script, updated `argparse.ArgumentParser()` `description` argument to use implicit concatenation and explicit `\n` so that when the user uses `-h` argument, the text displayed flows nicely but still respects those explicit newlines (similarly for any `parser.add_argument()`'s `help` argument)

<details>
<summary>Pylint and Black Formatter VSCode setup</summary>

I installed the following extensions to VSCode:
* [Pylint](https://marketplace.visualstudio.com/items?itemName=ms-python.pylint)
  * had to open the extension settings to set `Pylint: Cwd` to `${workspaceFolder}/TestSupport/planaritytesting`. This is because with the default set to `${workspaceFolder}`, so linting caused Python import paths to be changed to the full path relative to the root of the workspace; this resulted in errors like:
  ```
  from TestSupport.planaritytesting.planaritytesting_utils import (
  ModuleNotFoundError: No module named 'TestSupport'
  ```
* [Black Formatter](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter)

And added the following to my `.vscode/settings.json`
```
{
    ...
    "[python]": {
        "editor.rulers": [
            79,
            99
        ],
        "editor.defaultFormatter": "ms-python.black-formatter",
        "editor.formatOnSave": true
    },
    "editor.formatOnSave": true,
    "editor.renderWhitespace": "all",
    "files.autoSave": "afterDelay",
    "python.analysis.typeCheckingMode": "basic",
    "black-formatter.cwd": "${workspaceFolder}/TestSupport/planaritytesting",
    "black-formatter.args": [
        "--line-length=79"
    ],
}
```
</details>


john-boyer-phd (approved):
Outstanding work. This tool is going to be heavily used for quite a while.

Issue #59, opened 2024-04-26
EPIC: Create Inventory of and then Fix Any Memory Leak and Memory Overrun Issues


john-boyer-phd:
- [x] Set up a memory checker tool such as `Leaks` on the Mac first (Windows and Linux may be done later, with a different toolset). (see #81 )
- [x] Investigate whether the memory checker tool can be meaningfully run with the release executable versus only with the debug version. (see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/59#issuecomment-2190017861))
- [x] Make any necessary changes to make and/or configuration (`.ini` file generated using `configparser` module) files to allow rapid inclusion/exclusion of various memory leak and overrun detection features.
- [ ] Add a page to the wiki to describe how to set up and run the `Leaks` tool with planarity and how to do the inclusion/exclusion of the various features.
- [x] From the menu or the command-line, run the following
  - [x] all algorithms (P, D, O, 2, 3, 4) using specific graph mode, 
  - [x] random graphs with one small graph, 
  - [x] the random max planar graph generator,
  - [x] the random nonplanar graph (maximal planar + edge) generator,
  - [x] the graph converter, and 
  - [x] test all graphs 
  to see if there are any memory leaks or memory overruns on a file containing a single small graph (e.g., a 6-vertex or an 8-vertex graph). For each memory issue, add a story item to this list right below this item to fix the issue.
- [x] Generate all 8-vertex graphs into a single `graph6` file. This size has been chosen to ensure we trigger the memory expansion code that `gp_Read()` invokes (via `gp_DynamicAddEdge()`; see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/64#issuecomment-2089249300)). 
- [x] For the given platform (Mac with Leaks), run the memory checker tool on the `planarity` test all graphs function for each of the six algorithms (pdo234) on the set of 8-vertex graphs. For any memory leak or memory overrun issue(s), add issue item(s) directly below this item to fix the issue(s).
- [x] Do some calculations to figure out how long it would take to run all 10-vertex graphs through the memory leak and overrun testing, then decide how many 10-vertex graphs to do and add an issue.
[...]

PR #86, opened 2024-07-11
Adding `planarity_leaks_config.ini` default config file to source control

wbkboyer:
Follow-up to #59

Adding `planarity_leaks_config.ini` default config file to source control so that users will not have to run `planarity_leaks_config_manager.py` locally to generate it before updating the default values with the parameters necessary for their run.


john-boyer-phd (approved):
This is very helpful to make it easier for retesting modifications to existing command-line functions. Nice to see also that we can comment-out/remove the gitignoring of this file, commit a change, then re-gitignore it in the same PR.

PR #87, opened 2024-07-15
Issue #79 - generalize edge deletion analysis tool

### LINKED ISSUE #79: EPIC: Generalize Edge Deletion Analytical Tool to Test Validity of `OK` results from `K_4` and `K_{2,3}` Searches
The Edge Deletion Analytical Tool originally developed in #77 was specific to testing the validity of OK results from the `K_{3,3}` search (in other words, results in which the search reports that no `K_{3,3}` was found).

Earlier, we discovered that `geng`, canonical `makeg`, and canonical `geng` generated three mostly different sets of alternative representations of the same graphs. Differences in the number of `OK` results across these sets clearly indicated that there were errors occurring in some of the graph based on the differences in values for the number of `OK`s (no `K_{3,3}` found results). If one set (such as from `geng`) had two more `OK`s than in another set (such as from canonical `geng`), then we knew that at least two graphs in the first set were causing an error in the `K_{3,3}` search because a `K_{3,3}` was found in the alternative representation of that graph in the second set.

Then, along came the program from #77, which we used on all `OK` results from the `K_{3,3}` search. We found that the discrepancies in `OK` results across the three generated sets did not represent all of the `K_{3,3}` search `OK` results that were invalid. For example, the number of `OK` results might vary by 2 or 4 out of 700, but `K_{3,3}` homeomorphs could be found in 30 more cases by systematic single edge deletion from the original graphs that were to be `K_{3,3}`-free (`OK`) by the `K_{3,3}` search.

The goal of this epic is to generalize that program from #77 so that it instead invokes the `K_{2,3}` search (`-2` on the command line) or the `K_4` search (`-4` on the command line). 
* In the place where the core planar graph embedder/Kuratowski subgraph isolator is performed (`-p` on the command line), the generalized program should instead run the outerplanar graph embedder/outerplanarity obstruction isolator (`-o` on the command line) when performing an analysis of the efficacy of the `K_{2,3}` search or the `K_4` search.
* In places where there is a test of algorithm output tests for `K_{3,3}` versus `K_5` homeomorphs, the tests should instead distinguish between `K_4` versus `K_{2,3}` homeomorphs. This can be done by examining the adjacency list output to determine whether all vertices are degree 2 except for either two vertices of degree three (for a `K_{2,3}` homeomorph) or four vertices of degree three (for a `K_4` homeomorph).

wbkboyer:
Resolves #79

## Updated
* `TestSupport/planaritytesting/edge_deletion_analysis.py` - Generalized mechanism from #77 so that user may specify whether they wish to check the results of `K_{2, 3}` search, `K_{3, 3}` search, or `K_4` search to see if, for each `.g6` encoded graph in a `.g6` input file, the homeomorph search erroneously returned `OK` when in fact the graph contains a subgraph homeomorphic to the target homeomorph.

### DIFF
--- a/TestSupport/planaritytesting/edge_deletion_analysis.py
+++ b/TestSupport/planaritytesting/edge_deletion_analysis.py
@@ -70,7 +70,8 @@ class EdgeDeletionAnalyzer:
             output_dir: Directory under which a new subdirectory will be
                 created to which results will be written.
             extension_choice: Specifies the graph algorithm extension you wish
-                to test, either 3 for K_{3, 3} search or 4 for K_4 search
+                to test, either 2  for K_{2, 3} search, 3 for K_{3, 3} search
+                or 4 for K_4 search
 
         Raises:
             argparse.ArgumentTypeError: If the planarity_path doesn't
@@ -98,11 +99,11 @@ class EdgeDeletionAnalyzer:
                 "which is not supported; please supply a .g6 file."
             )
 
-        if extension_choice not in (3, 4):
+        if extension_choice not in (2, 3, 4):
             raise ValueError(
                 f"'{extension_choice}' is not a valid graph algorithm "
-                "extension command specifier; expecting either 3 (K_{3, 3}) "
-                "4 (K_4)"
+                "extension command specifier; expecting either 2 (K_{2, 3}), "
+                "3 (K_{3, 3}), or 4 (K_4)."
             )
 
         if not output_dir:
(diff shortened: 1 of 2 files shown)


wbkboyer:
This does not cover testing `K_{2, 3}` search results, as required by #79 ; additional refactoring is required.

john-boyer-phd (approved):
Great work generalizing this, including the updated help strings.

PR #90, opened 2024-07-29
Issue #80 - Integrate edge-deletion analysis `numInvalidOK`s into `planarity` `testAllGraphs()` Test Tables

### LINKED ISSUE #80: EPIC: Automate validity testing of `OK` results from homeomorph search algorithm implementations
The test table generator program (from #65) creates a table that shows the number of `OK` and `NONEMBEDDABLE` results (and other summary data) across all of the edge-specific files for a given vertex order.

The goal of the desired program for this issue is to import and improve upon the logic of the original test table generator to:
* create a table with an extra column indicating the number of `invalid OK` results, in total and per edge count based on performing edge deletion analysis (from #77 and #79) on all graphs reported to be `OK` by the homeomorphic subgraph search algorithm being tests (`K_{3,3}`, `K_{2,3}`, and `K_4`).
* create an output `.g6` file per edge count that contains all graphs with `invalid OK` results.

This program will need to name the output table file and the `.g6` files according to the algorithm being run as well as the graph order (`N` value). 

This program will need to be run on the outputs of `geng`, canonical `geng`, and canonical `makeg`. 

This program should be initially run on vertex orders `N=`6 to 10 inclusive to get a baseline of current invalid OK results before we start fixing bugs. The contents of the output tables for each algorithm and for `N=`6 to 10 (15 tables) should be preserved, such as by checking them in.

After this epic is done and the baseline recorded, the bugs will be fixed, and the tables will be updated, until the number of `invalid OK` results is uniformly zero).

wbkboyer:
Resolves #80 

## Updated
In `TestSupport/planaritytesting/`:
* `__init__.py` - `test_table_generator.py` module has classes that are available for import, so must be added to the package index
* `edge_deletion_analysis.py
  * Changed `EdgeDeletionAnalyzer.__MAX_NUM_INVALID_OKs` default to 5, although the tables were generated initially with a much larger threshold (1000); this should decrease overall processing time, and should still give enough examples for debugging purposes as we drive the `numInvalidOK`s down to zero.
  * Extension choices are now `str` rather than `int` to match typing of `EDGE_DELETION_ANALYSIS_SPECIFIERS()` (which was changed to `tuple[str, ...]` to match `PLANARITY_ALGORITHM_SPECIFIERS()`)
  * `EdgeDeletionAnalyzer.analyze_graphs()` now returns the `numInvalidOK`s encountered when processing the `.g6` infile with `Path` `infile_path`
  * Now includes timestamp in log format
* `test_table_generator.py`
  * `TestTableGeneratorPathError`, `TestTableGeneratorFileProcessingError`, and `TestTableGenerator` added to `__all__`
  * `TestTableGenerator` now accepts optional `edge_deletion_analysis_results`, and integrates values for `numInvalidOK`, `proc_time_s` ([process time](https://docs.python.org/3/library/time.html#time.process_time)), and `tot_time_s` ([wall time](https://docs.python.org/3/library/time.html#time.perf_counter)) into the final test table.
* `g6_diff_finder.py` - Now includes timestamp in log format (for consistency with logs generated by `EdgeDeletionAnalyzer`)
* `g6_generation_and_comparison_driver.py` - now uses `Release/planarity` as default for `planarity` executable `Path` (for consistency with `TestTableGenerationWithInvalidOKs`) 
* `planarity_testAllGraphs_orchestrator.py` - removed `# pylint: disable=invalid-name` and unused `import shutil`
* `planarity_testAllGraphs_output_parsing.py` - removed `# pylint: disable=invalid-name` 
* `planaritytesting_utils.py` - removed `# pylint: disable=invalid-name` and added `EDGE_DELETION_ANALYSIS_SPECIFIERS()` to return immutable tuple of strings corresponding to the command specifiers for graph search extensions supported by edge-deletion analysis (i.e. currently `(2, 3, 4)` for `K_{2, 3}` search, `K_{3, 3}` search, and `K_4` search)

## Added
* `TestSupport/tables/` - Added tables for `N=[5, 10]` with `# Invalid OK` field, process time, and wall time for edge-deletion analysis
* `TestSupport/planaritytesting/test_table_generation_with_numInvalidOK.py`
  * Allows the user to specify whether the `planarity` `testAllGraphs()` results were generated using the following files with the combination of `-l` and `-m` flags:
    * `geng` `.g6`
    * `geng` canonical `.g6`
    * `makeg` `.g6`
    * `makeg` canonical `.g6`
  * Allows the user to specify for which graph algorithm command specifiers they wish to generate the Test Tables
    * Defaults to generating tables for all `PLANARITY_ALGORITHM_SPECIFIERS()`, so by default edge-deletion analysis will be done for all `EDGE_DELETION_ANALYSIS_SPECIFIERS()`.
    * If user chooses to override using `-c` flag, then EDA is run for only the `EDGE_DELETION_ANALYSIS_SPECIFIERS()` that appear in the list.
      * e.g. if the user specifies `-c p o`, then `self.perform_eda_for_commands` is empty and we skip EDA entirely before the tables for planarity and outerplanarity are generated
      * e.g. if the user specifies `-c 3 p o`, then `self.perform_eda_for_commands` is `("3",)` and we do EDA for `K_{3, 3}` search on all `.g6` infiles for each graph order before the tables for `K_{3, 3}` search, planarity, and outerplanarity are generated
  * For each of the graph orders for which one wishes to generate the Test Tables with `numInvalidOK` data, performs the edge-deletion analysis on each of the `.g6` files for that order for each of the commands in `self.perform_eda_for_commands`, then generates the tables for all commands in `self.generate_tables_for_commands`


TestSupport/planaritytesting/edge_deletion_analysis.py
```diff
@@ -43,14 +44,16 @@ def __init__(self, message):
 class EdgeDeletionAnalyzer:
     """Performs edge-deletion analysis on graph(s) in a .g6 input file"""
 
-    __MAX_NUM_ERRORS = 1000
+    # Cap this at a much lower value, since EDA for N=10 takes days if this
+    # threshold is high
+    __MAX_NUM_INVALID_OKs = 5
```
john-boyer-phd, line 49: This should be changed to 500 to match the tables being checked in. Also, I think there may need to be a code tweak based on the fact that the n10 tables are getting to 501, when the max should be 500.

TestSupport/planaritytesting/g6_diff_finder.py
```diff
@@ -126,7 +126,8 @@ def _setup_logger(self, log_path: Optional[Path] = None) -> None:
 
             # Create a Formatter for formatting the log messages
             logger_formatter = logging.Formatter(
-                "[%(levelname)s] - %(module)s.%(funcName)s - %(message)s"
+                "%(asctime)s - [%(levelname)s] - %(module)s.%(funcName)s - "
```
john-boyer-phd, line 129: good you added the timestamp

TestSupport/planaritytesting/g6_generation_and_comparison_driver.py
```diff
@@ -230,7 +249,8 @@ def _generate_makeg_g6_files_for_order_and_num_edges(
         """Generate makeg .g6 and makeg canonical .g6 files
 
         The makeg .g6 and makeg canonical .g6 files generated have paths:
-            {self.output_parent_dir}/{order}/n{order}.m{num_edges}.makeg(.canonical)?.g6
+            {self.output_parent_dir}/{order}/
```
john-boyer-phd, line 252: That pylint though!

TestSupport/planaritytesting/test_table_generation_with_numInvalidOK.py
```diff
@@ -0,0 +1,585 @@
...
+            raise argparse.ArgumentTypeError(
+                f"test_table_output_dir = '{test_table_output_dir}' "
+                "corresponds to a file."
+            )
+
+        self.test_table_output_dir = test_table_output_dir
+
+        self._numInvalidOKs = {}
+
+    def _input_dir_contains_expected_files(
```
john-boyer-phd, line 174: I appreciate your checking this.

TestSupport/tables/6/n6.mALL.3.out.txt
```diff
@@ -0,0 +1,20 @@
...
+|n6.m4.g6        |4        |9         |9     |0                |SUCCESS   |0.0       |0             |0.000          |0.193           |
+|n6.m5.g6        |5        |15        |15    |0                |SUCCESS   |0.0       |0             |0.094          |0.328           |
+|n6.m6.g6        |6        |21        |21    |0                |SUCCESS   |0.0       |0             |0.109          |0.433           |
+|n6.m7.g6        |7        |24        |24    |0                |SUCCESS   |0.016     |0             |0.094          |0.506           |
+|n6.m8.g6        |8        |24        |24    |0                |SUCCESS   |0.0       |0             |0.172          |0.516           |
+|n6.m9.g6        |9        |21        |20    |1                |SUCCESS   |0.0       |0             |0.125          |0.434           |
+|n6.m10.g6       |10       |15        |14    |1                |SUCCESS   |0.016     |0             |0.109          |0.381           |
+|n6.m11.g6       |11       |9         |7     |2                |SUCCESS   |0.0       |0             |0.062          |0.354           |
```
john-boyer-phd, line 14: I believe Asano had a theorem stating that a graph with no `K_{3,3}` had to have 3n-5 or fewer edges. So, starting with 3*6-4=14 edges, we need to solve bugs until the 7 OK results here becomes 0 (unless there is a problem with Asano's result, which seems doubtful).

john-boyer-phd (changes requested):
Fantastic stuff. Two little tweaks, from first comment, and we're good to go.

TestSupport/planaritytesting/edge_deletion_analysis.py
wbkboyer: We decided to bump this back up to 500 so that as we drive down the `numInvalidOK`s for the lower graph orders (currently, `N=9` has a maximum `numInvalidOK`s of 399 for `K_{3, 3}` search, whereas for `N=10` the threshold is reached for `M=17`), we can see that max value decreasing relative to the tables committed to the repository. For each bugfix, we can commit the new version of the tables with the new (preferably lower!) `numInvalidOK`s.

TestSupport/planaritytesting/g6_generation_and_comparison_driver.py
wbkboyer: Yeah, in this case I included the path to the parent directory on one line and the filename on the subsequent line because the naming convention is OUTTA CONTROLLL lol!

john-boyer-phd (approved):
Excellent!

PR #92, opened 2024-08-08
Issue #91 - Source code reorganization

### LINKED ISSUE #91: EPIC: Source Code Reorganization
The source code for `planarity` needs to be reorganized to make easier to navigate and to reuse the relevant APIs in other projects in other repositories. At each step:

- various include files and include directives will need to be amended to restore compilation ability.  
- it will be necessary to amend `Makefile.am` and the developer project files for VS Code and Eclipse to restore and test all compile, run, `make dist`, `make install`, and `man` capabilities. 
- The template development environment files will need to be updated, if they exist (per solution to #89).

- [x] `planarity.c` should be reduced to just containing `main()`. All other code should be moved out to new files. Make a new file `planarityMenu.c` for all the methods below `main()` that are about the menu, and make `planarityHelp.c` for all the methods above `main()`. 
- [x] Move all files starting with `planarity` into a subdirectory called `planarityApp`. Ideally this would even include `planarity.c` and `planarity.1`
- [x] Move all files starting with `graph`, `g6`, `ap`, `li`, `st`, and `platform` into a `graphLib` directory. 
- [x] Within `graphLib`, create an `io` directory for `graphIO.c`, all `g6` files, and files starting with `str`.
- [x] Within `graphLib`, create a `lowlevelUtils` directory for files starting with `ap`, `stack`, `li`, and `platform`.
- [x] Within `graphLib`, create a `extensionSystem` directory for files starting with `graphExtensions` and `graphFunctionTable.h` 
- [x] Within `graphLib`, create a `homeomorphSearch` directory for all files starting with `graphK`
- [x] Within `graphLib`, create a `planarityRelated` directory for `graphEmbed.c`, `graphIsolator.c`, `graphNonplanar.c`, `graphOuterplanarObstruction.c`, `graphTests.c` and all files starting with `graphDrawPlanar`
- [x] In `Makefile.am`, pull the planarityApp files out of libPlanarity sources and put them into planarity sources (we will need to deal with a different version number update to libPlanarity later, as a separate issue).
- [x] Removed wiki update task because it would be better to have the code reorg merged to main before editing the wiki to talk about the structure, so separate issue, #97.
- [x] Within `graphLib` add a file called `graphLib.h` and use the copyright message, `#ifndef` protection, and `extern C` encapsulation. Inside the `extern C` envelope, do the includes currently shown on lines 14 through 24 of `planarity.h`.
- [x] Within `planarity.h`, change the includes on lines 14-24 to instead include `graphLib.h`

wbkboyer:
Resolves #91 

## Notable changes
* `.gitignore` - Changed so that any directories named `Debug/`, `Release/`, `random/`, `embedded/`, `adjlist/`, `obstructed/`, `error/`, or `.deps/` are ignored, no matter their parent directory. Also added updated to ignore `autogen` build artifacts (i.e. `planarity-\d+.\d+.\d+.tar.gz` tarball and `planarity-\d+.\d+.\d+/` directory)
* `.vscode/`
  * `launch.json` - Added Python debug launch configuration with example demonstrating test parameterization for `TestSupport/planaritytesting/test_table_generation_with_numInvalidOK.py`
  * `tasks.json` - Updated all configurations so that they include all the source files in compilation
* `Makefile.am` - Updated `libplanarity_la_SOURCES`, `pkginclude_HEADERS`, `planarity_SOURCES`, and `man1_MANS` to reflect new structure.
  * _**N.B.**_ `planarity_SOURCES` now includes `c/planarityApp/planarityHelp.c` and `c/planarityApp/planarityMenu.c` due to functions moved out of `c/planarityApp/planarity.c`. More discussion is required as to what is the minimal set of `.c` files required to build the `planarity` executable (i.e. since `planarity.c`'s `main()` calls `planarityMenu.c`'s `menu` and `planarityCommandLine.c`'s `commandLine()` and `legacyCommandLine()`
* Updated `#include` statements throughout the project using **relative paths**
  * _**N.B.**_ There is significant debate as to whether or not this is the best practice, and might require further consideration. For example, one might change the `tasks.json` to use the `-I` compiler directive to set the root to `${workspaceFolder}` so that import statements would be of the form `#include "path/to/foo.h"` rather than a relative path. This would also need to be done in the `Makefile.am` using [`AM_CFLAGS`](https://www.gnu.org/software/automake/manual/1.7.7/automake.html#index-AM_005fCFLAGS) or `planarity_CFLAGS`


john-boyer-phd (changes requested):
This is really looking a lot nicer and cleaner. A couple of minor tweaks:
1) list collection belongs in lowLevelUtils (revised original spec)
2) The planarity sources, not libplanarity, is what should contain the planarity test all graphs c file (contrary to my prior advisement when you were doing that feature). The problem is that planaritySpecificGraph.c and planarityRandomGraphs.c were inadvertently placed in libplanarity long ago, but we may as well not make that mistake worse.
3) Pull the other three planarityApp files out of libPlanarity sources and into planarity sources (we will need to deal with a different version number update to libPlanarity later as a separate issue).
4) Update the wiki page to reflect the new structure.

john-boyer-phd (changes requested):
This is looking great. Our discussions caused me to add two bullet points to the end of the issue to request the creation and use of a `graphLib.h` header. This will facilitate doing the `graphLib submodule work for #95. Also, please update the wiki page with this reorg then request review so that the approval and squash merge can also close the issue #91.

c/planarityApp/planarityHelp.c
```diff
@@ -28,55 +28,6 @@ char *GetProjectTitle()
            "\n==================================================\n";
 }
 
-/****************************************************************************
```
john-boyer-phd: Great catch noting that these don't really belong in the new `planarityHelp.c` file.

john-boyer-phd (approved):
I've moved the wiki change to a separate issue (#97) because I'd like this reorg in main before the wiki talks about them.

PR #101, opened 2024-09-24
Issue #94 - Separate Transform Graph and Test All Graphs functionality

wbkboyer:
Resolves #94 

## Removed
- `c/planarityApp/planarityTestGraphFunctionality.c` - split out into `planarityTransformGraph.c` and `planarityTestAllGraphs.c`

## Added
- `c/planarityApp/`
  - `planarityTransformGraph.c` - Contains `TransformGraph()`, which is comprised of the first branch of the old `TestGraphFunctionality()` function
  - `planarityTestAllGraphs.c` - Contains `TestAllGraphs()`, which is comprised of the second branch of the old `TestGraphFunctionality()` function. (note change in function signature, since Test All Graphs branch in old `TestGraphFunctionality()` function disallowed `inputStr`)

## Updated
- `.vscode/launch.json` - Corrected Transform Graph example launch configuration command-line parameterization (i.e. `planarity -x -(gam) infileName outfileName` rather than `planarity -t -t(gam) infileName outfileName`)
- `Makefile.am` - Updated `planarity_SOURCES` to include `c/planarityApp/planarityTransformGraph.c` and `c/planarityApp/planarityTestAllGraphs.c`
- `c/planarityApp/`
  - `planarity.1` - Updated manpage to split up old `-t` flag into transform graph (`-x`) versus test all graphs (`-t`)
  - `planarityHelp.c` - Updated `helpMessage()` to split up old `-t` flag into transform graph (`-x`) versus test all graphs (`-t`)
  - `planarity.h` - Added declarations for `TransformGraph()` and `TestAllGraphs()` (note change in function signature, since Test All Graphs branch in old `TestGraphFunctionality()` function disallowed `inputStr`)
  - `planarityCommandLine.c` - replaced `callTestGraphFunctionality()` with two new functions, `callTransformGraph()` (note that we now expect `-(gam)` rather than `-t(gam)`, so expected length of `command` is reduced within both `callTransformGraph()` and `runGraphTransformationTest()`) and `callTestAllGraphs()`. Also updated the arguments of the calls to `runGraphTransformationTest()` within `runSpecificGraphTests()` so that they are of the form `-(gam)` rather than `-t(gam)`.
  - `planarityMenu.c` - Removed `TransformOrTestMenu()` layer so that `menu()` directly leads to the transform graph and test all graphs menus; renamed `TransformMenu()` to `TransformGraphMenu()` and `TestMenu()` to `TestAllGraphsMenu()`

PR #107, opened 2024-10-21
Issue 105 - Updating test tables to verify `numInvalidOK`s determined by EDA driven to 0 for N=6,10

wbkboyer:
Additional verification of fix #106

## Updated
* `TestSupport/planaritytesting/edge_deletion_analysis.py` - `EdgeDeletionAnalyzer`'s `transform_input()` method calls planarity executable with correct command-line parameters to transform graph to adjacency list (see #101)
* `TestSupport/tables/` - regenerated tables for `K_{3, 3}` search for `N=5,10` (even though `N=5` can't contain a `K_{3, 3}`, I had previously included those tables so I wanted to make sure they were kept in-sync)


john-boyer-phd (approved):
It's really great to see from your automated retesting that the bug fix resolved all invalid `K_{3,3}-free` results for N=6,10.

PR #114, opened 2024-11-15
Issue #108 - add save state mechanism to test table generation with edge-deletion analysis results

### LINKED ISSUE #108: Update `test_table_generation_with_numInvalidOK.py` to save EDA progress with option to resume analysis
## Background

The Edge-Deletion Analyzer originally developed in #77 was generalized in #79 to check the results of `K_{2, 3}` search, `K_{3, 3}` search, or `K_4` search to see if, for each `.g6`-encoded graph in a `.g6` input file, the homeomorph search erroneously returned `OK` when in fact the graph contains a subgraph homeomorphic to the target homeomorph. Then, #80 introduced the `test_table_generation_with_numInvalidOK.py`, which runs edge-deletion analysis on all `.g6` files for each each edge-count of each graph order for which we wish to generate the Test Tables (#65). 

## Problem statement

Initially, I kicked off the `test_table_generation_with_numInvalidOK.py` for `N=5,11`, but discovered after a few days that the program had terminated, only producing the updated test tables for `N=5,9`; I thought this was due to Windows updating and then rebooting, since I had to log back into my workstation and observed the terminal window had closed, so I restarted the process only for for `N=10,11`.

The `test_table_generation_with_numInvalidOK.py` successfully produced the test table for `N=10` indicating the `numInvalidOKs` had been driven down to 0 (see #107 ). Sometime during this processing for `N=10`, I noted the application was using roughly 6GB of RAM. However, during the processing of `N=11`, I noticed that the memory footprint continued to increase at a greater pace, ballooning up to 18 GB while working through the graphs with `M=19`, which alone took 47 hours to process:

```
2024-10-19 10:48:51,288 - [INFO] - test_table_generation_with_numInvalidOK._call_edge_deletion_analysis - NUMBER OF INVALID OKs in 'n11.m19.g6' is 0; took 42263.218750s process time and 170562.578561s total time
```

I decided to terminate the program while processing `N=11` `M=20` when it reached 22GB RAM, since even though it likely could have continued (with Windows handling paging, up to an additional 64GB for the max pagefile size), there was no guarantee for whether we'd make it over processing all graphs in each of the `.g6` infiles for the "fat middle" of edge-counts.

This means that if we want to determine the `numInvalidOKs` for `N=11` using edge-deletion analysis, we have to start from scratch running the `test_table_generation_with_numInvalidOK.py` and hope that we don't run out of memory + pagefile, since there's no way to resume after execution has been halted.

## Possible causes

When examining `test_table_generation_with_numInvalidOK.py`, I noted the following:
1. I am using `with open()` Python context manager to create a new file containing the adjacency list representation for each graph in the input file; however, this is still being done for a *lot* of graphs e.g. 8,640,134 graphs for `N=11, M=19`, as verified by `geng` with `-u` (generate + count without output):
    ```
    % ./geng 11 19:19 -u
    >A ./geng -d0D10 n=11 e=19
    >Z 8640134 graphs generated in 2.19 sec
    ```
    So even though context managers are meant to safeguard against memory leaks, there might be a bug that's only apparent with such a large number of file writes.
1. For each graph, if the graph is determined to be free of the target homeomorph but is non(outer)planar, I'm calling `subprocess.run()` for each graph-minus-one-edge for however many edges are in that graph to see if we can find the target homeomorph. This means we're spawning an INSANE number of processes, which according to [this](https://www.reddit.com/r/learnpython/comments/10qokch/subprocesspopen_causes_memory_leak) Reddit post might be the source of a memory leak:
    > I have determined that the memory leak is due to a Page Table Entry(PTE) leak of processes that have been killed. While the processes spawned by Popen() have correctly executed and terminated, Windows, for some reason, keeps their PTEs in RAM, resulting in 32K~36K loss per instance that adds up to tens of GBs over a few hours. C++ showed the same symptom. This is a Windows problem.

## Proposed solution

Since the cause is likely an OS-level issue rather than even one with the Python `subprocess` module, it would be fruitful to have some way to save state after finishing processing an edge-count for a given graph order, and then to have an option to restore from that save state and resume processing; then, if the OS runs out of memory and the system crashes, we at least will have the intermediate results from lower edge-counts. Furthermore, if the user is able to specify an edge-count limit, processing may be done up to and including that edge-count and preserved in the resume file so that the user may gracefully reboot the system and resume up to the next highest edge-count limit.

wbkboyer:
Resolves #108 

_**N.B.**_ Updated [3. Test Support](https://github.com/graph-algorithms/edge-addition-planarity-suite/wiki/3.-Test-Support) with details pertaining to this PR.

## Updated
* `TestSupport/planaritytesting/planaritytesting_utils.py` - added `raise_on_duplicates()` to be used as `object_pairs_hook` for `json.load()` to populate `resume_file_contents` in `ResumableTTGwithEDA._initialize_numInvalidOKs()`, since although the JSON specs indicate that keys **should** be unique, this is not enforced (see [this](https://stackoverflow.com/a/23195243/5456879) StackOverflow discussion)
* `TestSupport/planaritytesting/test_table_generation_with_numInvalidOK.py` - split the functionality of `TestTableGenerationWithInvalidOKs` class into an orchestrator and driver pattern to simplify command line parameter validation and business logic
  * Added the `TTGWithEDAPathConfig` `dataclass` as a container for paths relevant to EDA and test table generation; this was primarily to reduce the number of parameters when initializing other classes, but the `dataclass` pattern has other benefits: it's easier to explicitly indicate typing, the `__post_init()` allows us to separate initialization from setting defaults and validation, and it's immediately apparent when you've made a typo (vs. storing related attributes in a `dict`)
  * Added `TTGwithEDAOrchestrator` class to handle initializing the correct driver (i.e. concrete class of ABC `TTGwithEDA`) based on the command line parameterization and calling the instance's concrete methods (implementations of the abstract methods dictated by the ABC)
  * Added `TTGwithEDA` abstract base class to indicate what attributes and functionality are required by a driver for running EDA and generating the corresponding table; using the `abstractmethod` decorator ensures that one may not create an instance of `TTGwithEDA`, and that any concrete class that implements this ABC must provide an implementation of `determine_numInvalidOKs()` and `generate_tables_with_numInvalidOKs()` that do some work before calling `_determine_numInvalidOKs()` and `_generate_table_with_numInvalidOKs_for_order_and_command()` respectively
  * Added `MultiOrderOrCommandTTGwithEDA` and `ResumableTTGwithEDA` concrete classes of the abstract base class `TTGwithEDA`, where the former is instantiated by the `TTGwithEDAOrchestrator` and assigned to `self.ttg_with_eda_driver` if multiple graph orders (provided the max doesn't exceed 9) and/or command specifiers are indicated by the command line parameters, and the latter is instantiated and assigned if a single graph order and command specifier are given (which is also the only case in which the user providing an `edgecount_limit` makes sense). For the `ResumableTTGwithEDA`, we check if there is a resume file for the given order and command, and if so, then restore `self._numInvalidOKs` from the `.json` file and proceed with performing EDA (if applicable) for edge-counts starting from the one following the highest previously processed up to and including the `edgecount_limit` (defaults to $\frac{N\times\(N - 1\)}{2}$).


john-boyer-phd (approved):
Looks great, very handy for crash recovery.

PR #115, opened 2024-11-18
Issue #109 - Always display "# Invalid OK", "EDA Proc Time", and "EDA Total Time" columns if EDA has been run

### LINKED ISSUE #109: Always include "# Invalid OK", "EDA Proc Time", and "EDA Total Time" in test tables
## Problem statement

When we run the `test_table_generation_with_numInvalidOK.py` script, if the `numInvalidOK` is zero for graphs of a given order and algorithm command specifier, we don't render the columns for "# Invalid OK", "EDA Proc Time" ([process time](https://docs.python.org/3/library/time.html#time.process_time)) , and "EDA Total Time" ([wall time](https://docs.python.org/3/library/time.html#time.perf_counter)) in the final test tables. This was initially as-planned, as this process was meant to highlight when something went wrong with a graph algorithm search extension (when `numInvalidOK` is nonzero; see #42). However, for larger `N`, the edge-deletion analysis spawns so many processes that Windows has a Bad Time™ (see #108). We want to be able to measure speed improvements to the `EdgeDeletionAnalyzer` even when there are no `numInvalidOK`, and so we always want to include those three columns in the generated test tables.

<details>
<summary>Overview of how TestTableGenerationWithInvalidOKs leverages the EdgeDeletionAnalyzer and sends those results to the TestTableGenerator</summary>

Currently, the `test_table_generation_with_numInvalidOK.py` script creates an instance of the `TestTableGenerationWithInvalidOKs` class and calls its member method `generate_test_tables_with_invalidOKs()`. This function calls `_determine_numInvalidOKs()`, which, for each edge-count of that order, calls `_call_edge_deletion_analysis()`. This creates an instance of the `EdgeDeletionAnalyzer` to determine the `numInvalidOK_for_infile`, the `proc_time_s`, and `tot_time_s`, and assigns these to the sub-dict `self._numInvalidOKs[order][command][infile_name]`.

Once this is done for each edge-count for that order, we call `_generate_test_table_with_numInvalidOK()`, which creates an instance of the `TestTableGenerator` with `edge_deletion_analysis_results` assigned the appropriate sub-dict, i.e. 

```
    edge_deletion_analysis_results=self._numInvalidOKs.get(
        order, {}
    ).get(command, {})
```

Then calls its `process_files()` method to build the `self._processed_data` dict by parsing the `planarity` `TestAllGraphs()` output files for the given algorithm command specifier, as well as extracting the `numInvalidOK`, `proc_time_s`, and `tot_time_s` from the `self._edge_deletion_analysis_results` dict passed in at initialization and to finally aggregate those results in the `self._totals` dict.

Finally, when we call `write_table_formatted_data_to_file()`, we render the columns for "# Invalid OK", "EDA Proc Time", and "EDA Total Time" *only if* the `self._totals` dict has a nonzero value for `numInvalidOK`.

</details>

## Proposed solution

The logic determining when the `TestTableGenerator` should render the columns for "# Invalid OK", "EDA Proc Time", and "EDA Total Time" must be updated so that:
* When the `test_table_generator.py` script is run standalone (i.e. `edge_deletion_analysis_results` is `{}`), these columns will *not* exist
* When we run the `test_table_generation_with_numInvalidOK.py` script, the columns are always rendered so that even if `numInvalidOK` is 0, we still have the `proc_time_s` and `tot_time_s`.

wbkboyer:
Resolves #109 

_**N.B.**_ Updated [3. Test Support](https://github.com/graph-algorithms/edge-addition-planarity-suite/wiki/3.-Test-Support) with details pertaining to this PR.

## Updated
* `TestSupport/tables/` - Re-ran TTG with EDA for `N=5,10`
* `TestSupport/planaritytesting/test_table_generator.py` - Previously, `write_table_formatted_data_to_file()` would only show the "# Invalid OK", "EDA Proc Time", and "EDA Total Time" columns if `self._totals.get("numInvalidOK")` was truthy; now we condition on the truthiness of `self._edge_deletion_analysis_results`, which is only a nonempty dict if `TestTableGenerator` is initialized by `TestSupport/planaritytesting/test_table_generation_with_numInvalidOK.py` with the EDA results for an order and command


john-boyer-phd (approved):
Very glad to have these updated tables that visually show no invalid OKs.

Issue #46, opened 2024-04-09
Resolve suggestions+warnings from make distribution process


wbkboyer:
When one uses the autotools build system on both Windows (autotools packaged with MinGW) and Debian Linux (after `sudo apt install build-essential dkms linux-headers-$(uname -r) gdb libtool autoconf autotools autotools-dev`; documenting the Debian Linux dev environment setup steps will be the subject of an upcoming task), warnings are emitted:

* On Windows using MinGW (which includes autotools), 
  ```
  $ ./autogen.sh
  libtoolize: putting auxiliary files in `.'.
  libtoolize: copying file `./ltmain.sh'
  libtoolize: Consider adding `AC_CONFIG_MACRO_DIR([m4])' to configure.ac and
  libtoolize: rerunning libtoolize, to keep the correct libtool macros in-tree.
  libtoolize: Consider adding `-I m4' to ACLOCAL_AMFLAGS in Makefile.am.
  configure.ac:24: installing `./config.guess'
  configure.ac:24: installing `./config.sub'
  configure.ac:2: installing `./install-sh'
  configure.ac:2: installing `./missing'
  Makefile.am: installing `./depcomp'
  ```
  **N.B.** When this issue was written, the Windows development environment included MinGW32, whose version of `autotools` doesn't support `AC_CONFIG_MACRO_DIRS`, but rather `AC_CONFIG_MACRO_DIR`; however, this is not a problem as there is only one macro dir (i.e. `m4`) that the process needs to know about, and `AC_CONFIG_MACRO_DIR` is supported by the autotools build system on Debian Linux (see `autoconf` 2.70 docs section[4.4 Configure Input: Source Code, Macros, and Auxiliary Files](https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Input.html)). the `autotools` one installs within MSYS2+MinGW-64 supports the correct macro and doesn’t throw this error.
* On Debian using 
  ```
  $ ./autogen.sh 
  configure.ac:24: warning: The macro `AC_PROG_LIBTOOL' is obsolete.
  configure.ac:24: You should run autoupdate.
  aclocal.m4:122: AC_PROG_LIBTOOL is expanded from...
  configure.ac:24: the top level
  libtoolize: putting auxiliary files in '.'.
  libtoolize: copying file './ltmain.sh'
  libtoolize: Consider adding 'AC_CONFIG_MACRO_DIRS([m4])' to configure.ac,
[...]

PR #117, opened 2024-11-25
configure.ac: run autoupdate

orlitzky:
Two minor changes that the "autoupdate" program suggests:

* Quote the arguments to `AC_INIT`
* Replace `AC_PROG_LIBTOOL` with `LT_INIT`

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,4 @@
-AC_INIT(planarity, 3.0.2.0, jboyer@acm.org)
+AC_INIT([planarity],[3.0.2.0],[jboyer@acm.org])
 AM_INIT_AUTOMAKE([subdir-objects] [foreign])
 AC_CONFIG_SRCDIR([c/])
 
@@ -21,7 +21,7 @@ AC_SUBST(LT_REVISION)
 AC_SUBST(LT_AGE)
 
 AC_PROG_CC
-AC_PROG_LIBTOOL
+LT_INIT
 AC_PROG_INSTALL
 
 AC_CHECK_HEADERS([ctype.h stdio.h stdlib.h string.h time.h unistd.h])


wbkboyer:
I'll take a look at this; I need to double-check if this resolves the warnings from #46 !

orlitzky:
It should resolve the warning about `AC_PROG_LIBTOOL` at least. Square brackets are quotes in m4, so adding those should be harmless, and according to the documentation, `LT_INIT` is simply the new name for the `AC_PROG_LIBTOOL` macro: https://www.gnu.org/software/libtool/manual/html_node/LT_005fINIT.html#index-LT_005fINIT

orlitzky:
> 1. Rearrange commands in [autogen.sh](http://autogen.sh/) so libtool --copy would come first and we have the m4 directory populated before calling aclocal and autoconf (resolve warnings on Windows)
> 3. Additional sources indicate it might be a good practice to add [autoreconf -ivf](https://stackoverflow.com/a/32003575/5456879) to autogen.sh.

Unless you are doing something weird (and you aren't), it should suffice to put **only** `autoreconf -fi` inside `autogen.sh`. The autoreconf program is a wrapper around all of the other stuff in there, so you don't have to remember five separate programs to run and the correct order to run them in. But I also have no experience on Windows, so who knows.

> 2. Add AC_CONFIG_MACRO_DIRS([m4]) to configure.ac and ACLOCAL_AMFLAGS = -I m4 to Makefile.am

This isn't a warning _per se_. If you are using any local macros, normally they get concatenated into an `aclocal.m4` that is shipped with your program.  If you add those two lines, though, your local macros will instead get placed in the `m4` directory and shipped separately (`aclocal.m4` will `m4_include` them). It's a little nicer for people who want to edit the macros, but you don't have any local macros besides the libtool boilerplate.

So it's up to you. Adding them makes the suggestion go away, but leaves you with an empty `m4` directory in your source tree that technically doesn't need to be there. Which is more annoying?

wbkboyer (approved):
Heck yeah, this is great; I just double checked, and replacing the existing contents of `autogen.sh` with only the command `autoreconf -fi` works as expected on Windows 10 (MinGW32) and Debian 12.7, but I'll leave that for a follow-up PR! 💪🏼 

Thank you as always for your contributions and knowledge-share, @orlitzky; you're an absolute rockstar!! 🚀

Issue #118, opened 2024-11-25
Build fails with GCC 15 due to argument mismatches


orlitzky:
Original report: https://bugs.gentoo.org/944844

GCC 15 is not happy that you have declared

```C
int  (*fpSortVertices)();
```

but then call it with arguments:

```C
int gp_SortVertices(graphP theGraph)
{
    return theGraph->functions.fpSortVertices(theGraph);
}
```

I was able to fix(?) this one problem by declaring,

```C
typedef struct baseGraphStructure baseGraphStructure;
```

at the top of `c/graphLib/extensionSystem/graphFunctionTable.h` and then updating the signature,

```C
int  (*fpSortVertices)(baseGraphStructure *);
```

...But, then it looks like some of the other function table entries have the same problem:

```
$ make 2>&1 | grep 'too many'
c/graphLib/graphUtils.c:200:12: error: too many arguments to function 'theGraph->functions.fpInitGraph'
c/graphLib/graphUtils.c:319:5: error: too many arguments to function 'theGraph->functions.fpReinitializeGraph'
c/graphLib/graphUtils.c:421:12: error: too many arguments to function 'theGraph->functions.fpEnsureArcCapacity'
c/graphLib/graphUtils.c:1822:9: error: too many arguments to function 'theGraph->functions.fpHideEdge'
c/graphLib/graphUtils.c:1850:9: error: too many arguments to function 'theGraph->functions.fpRestoreEdge'
c/graphLib/graphUtils.c:1949:16: error: too many arguments to function 'theGraph->functions.fpHideVertex'
c/graphLib/graphUtils.c:1992:16: error: too many arguments to function 'theGraph->functions.fpContractEdge'
c/graphLib/graphUtils.c:2049:16: error: too many arguments to function 'theGraph->functions.fpIdentifyVertices'
c/graphLib/graphUtils.c:2229:16: error: too many arguments to function 'theGraph->functions.fpRestoreVertex'
```

GCC 15 won't be released until around April 2025, but we are trying to get ahead of the inevitable chaos.

wbkboyer:
Thank you for surfacing this, @orlitzky; I believe this is related to the warnings in #40 !

[...]

PR #119, opened 2024-11-26
Issues 40 and 118 - function pointer param fix

### LINKED ISSUE #40: Remediate Function Pointer Warnings from Mac clang Compiler
This project provides algorithmic extensibility of the planarity-related algorithms by using functions pointers to make virtual those functions that may need to be overloaded. Ideally, we would like to simply include the parameter lists of the virtual functions in their function pointer declarations. However, this causes compiler errors for many compilers including _gcc_. 

Recently, we have been adding support for _clang_ on the Mac. However, it produces 92 compiler warnings, one per invocation of a function via function pointer. The warnings indicate that such function invocations without parameter lists is deprecated. Online documentation suggests that support has been removed in C23. Despite producing the warnings, _clang_ also produces compiler errors if the parameter lists are added to the function pointer declarations. 

We need a method to eliminate the _clang_ warnngs while not introducing any compilation errors for previously supported compilers including _gcc_. Preferably, the new declaration method will also be compatible with C23 compilation.

wbkboyer:
Resolves #40 and #118

Extrapolated fix recommended by @orlitzky and verified semantics with @john-boyer-phd

## Updated
* `c/graphLib/graphStructures.h` - now declare named struct `baseGraphStructure` and `typedef` that struct so that the `typedef` of `graphP` works (since `graphUtils.c` `#include`s `graphStructures.h`, there's no need to add the same`typedef` statements)
* `c/graphLib/extensionSystem/graphFunctionTable.h` - updated so that we can use `graphP` in parameter list (external linkage due to `typedef`)

Verified Debug and Release builds on all supported platforms, i.e. MacOS `clang`, Windows 10 (MinGW32 `gcc` and MSVC `cl`), and Debian 12.7 (`gcc` 12.2.0)

_**N.B.**_ I tried cloning the `gcc` [repository](https://gcc.gnu.org/git.html), but there's no branch yet for `gcc-15` (which is slated to be released in April 2025); however, I am going to try to compile the [`gcc-15` snapshot](https://gcc.gnu.org/pub/gcc/snapshots/LATEST-15/) on Debian linux using the provided [instructions](https://gcc.gnu.org/install/index.html) and test compilation of this branch of `planarity` again to ensure there's no warnings or errors

### DIFF
--- a/c/graphLib/extensionSystem/graphFunctionTable.h
+++ b/c/graphLib/extensionSystem/graphFunctionTable.h
@@ -8,53 +8,56 @@ See the LICENSE.TXT file for licensing information.
 */
 
 #ifdef __cplusplus
-extern "C" {
+extern "C"
+{
 #endif
 
-/*
- NOTE: If you add any FUNCTION POINTERS to this function table, then you must
-       also initialize them in _InitFunctionTable() in graphUtils.c.
-*/
+    /*
+     NOTE: If you add any FUNCTION POINTERS to this function table, then you must
+           also initialize them in _InitFunctionTable() in graphUtils.c.
+    */
+    typedef struct baseGraphStructure baseGraphStructure;
+    typedef baseGraphStructure *graphP;
 
-typedef struct
-{
+    typedef struct
+    {
         // These function pointers allow extension modules to overload some of
         // the behaviors of protected functions.  Only advanced applications
         // will overload these functions
-    	int  (*fpEmbeddingInitialize)();
-        void (*fpEmbedBackEdgeToDescendant)();
-        void (*fpWalkUp)();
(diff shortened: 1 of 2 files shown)


orlitzky:
I don't mean to spoil your fun, but it builds fine with gcc-15 for me and the tests pass.

john-boyer-phd:
> I don't mean to spoil your fun, but it builds fine with gcc-15 for me and the tests pass.

Thank you @orlitzky We're glad to know this works for your gcc-15 compiler also, and your checking has saved quite a bit of work for @wbkboyer :-)

john-boyer-phd (approved):
This is great, and Michael confirmed it also compiles on the gcc compiler to be released next year.

Issue #123, opened 2024-12-04
Resolve warnings emitted by `gcc` and `clang` with `-Wall -Wextra` and by MSVC `cl` with `/Wall` 


wbkboyer:
## Problem statement

While working on #96 , I noticed that MSVC `cl` with the [`/Wall`](https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170) flag produces warnings not produced by `gcc` nor `clang` when the `-Wall` flag is used.

From the documentation on MSVC `cl`'s warning flags:

>  /W4 displays level 1, level 2, and level 3 warnings, and all level 4 (informational) warnings that aren't off by default. We recommend that you use this option to provide lint-like warnings. For a new project, it may be best to use /W4 in all compilations. This option helps ensure the fewest possible hard-to-find code defects.
> /Wall  Displays all warnings displayed by /W4 and all other warnings that /W4 doesn't include—for example, warnings that are off by default. For more information, see Compiler warnings that are off by default.

Looking at the documentation for `gcc`, this seems to correspond to the [`-Wextra`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-W) flag (similarly for `clang`: [`-Wextra`](https://clang.llvm.org/docs/DiagnosticsReference.html#wextra))

When `-Wall -Wextra` is used, a lot of the same warnings are emitted as by MSVC `cl`; in particular, there's a whole bunch of ["unreferenced formal parameter"](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/96#issuecomment-2513208166) warnings that either should be resolved or suppressed on a case-by-case basis.

For example, if we wanted to ignore the warning

* MinGW `gcc`:
```
C:\Users\wbkbo\git\edge-addition-planarity-suite-fork\c\planarityApp\planaritySpecificGraph.c:206:35: warning: unused parameter 'theGraph' [-Wunused-parameter]
  206 | void WriteAlgorithmResults(graphP theGraph, int Result, char command, platform_time start, platform_time end, char *infileName)
      |     
```

* Linux `gcc`:
```
/home/wbkboyer/git/edge-addition-planarity-suite/c/planarityApp/planaritySpecificGraph.c:206:35: warning: unused parameter ‘theGraph’ [-Wunused-parameter]

```

* MSVC `cl`:
```
[...]

PR #124, opened 2024-12-04
Issue #96 - Resolve warnings from setting `-Wall` on `clang` and `gcc`

### LINKED ISSUE #96: Resolve warnings from `-Wall` setting in release and debug configurations for `gcc` and `clang`
There are some additional compiler warnings that occur when `-Wall` is added to the compiler flags. Please add `-Wall` to VSCode, Eclipse, and make compilation options, release and debug, then resolve warnings that pop up due to the added compiler flag.

wbkboyer:
Resolves #96 

## Updated
* `devEnvSetupAndDefaults/.vscode/tasks.json` - Enable comprehensive warnings within all build commands; adding [`-Wextra`](https://clang.llvm.org/docs/DiagnosticsReference.html#wextra) to `clang` and [`-Wextra`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#index-W) to `gcc` (both Linux and MinGW32) build commands now yields largely the same warnings as those produced by MSVC `cl`, but are currently commented out and will be the subject of #123 . Additionally, I've chosen to ignore the following warnings issued by MSVC `cl` using the [`/wd`](https://learn.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=msvc-170) [compiler option](https://learn.microsoft.com/en-us/cpp/build/reference/compiler-options-listed-alphabetically?view=msvc-170):
    * [C4464](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4464?view=msvc-170) - Ignore warnings about relative paths
    * [C4710](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4710?view=msvc-170) - Ignore warnings about not inlining scanf, sprintf, printf, fprintf, etc.
    * [C4711](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4711?view=msvc-170) - Ignore warnings about automatic inline expansion, as we are not concerned with code size
    * C4774 (see [Microsoft C/C++ compiler warnings C4600 through C4799](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warnings-c4600-through-c4799?view=msvc-170)) - Ignore warnings about format string not being a string literal; requires developer caution!
    * [C4996](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=msvc-170) - Ignore safety warnings about sprintf, scanf, strcpy, strcat, fopen, etc.
    * [C5045](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/c5045?view=msvc-170) - Ignore Spectre mitigation warning 
* `c/graphLib/graphUtils.c` - `gp_New()` signature now includes `void` to get rid of warning `no function prototype given: converting '()' to '(void)'`
* `c/graphLib/io/graphIO.c` - Removed unused variable `Ch` from `gp_ReadFromString()`
* `c/graphLib/io/strOrFile.c` - Removed unnecessary check `(strlen(theStr) >= 0)` from conditional statements (see [here](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/96#issuecomment-2512807922) for justification)
* `c/graphLib/lowLevelUtils/apiutils.c` - `getQuietModeSetting()` signature now includes `void` to get rid of warning `no function prototype given: converting '()' to '(void)'`
* `c/graphLib/lowLevelUtils/appconst.h` - `debugNOTOK()` signature now includes `void` to get rid of warning `no function prototype given: converting '()' to '(void)'`
* `c/planarityApp/planarity.h` - removed unused parameter `outputBase` from `TestAllGraphs()` declaration. Signatures for `Reconfigure()` and `menu()` now include `void` to get rid of warning `no function prototype given: converting '()' to '(void)'`
* `c/planarityApp/planarityCommandLine.c` - Resolved warning from MSVC about `(chdir|getcwd) undefined; assuming extern returning int` (see [here](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/96#issuecomment-2512496440)), and updated args list for call to `TestAllGraphs()` from `callTestAllGraphs()` since we no longer need to pass `NULL` for `outputBase` (missed cleanup in #101 )
* `c/planarityApp/planarityHelp.c` - `GetProjectTitle()` signature now includes `void` to get rid of warning `no function prototype given: converting '()' to '(void)'`
* `c/planarityApp/planarityMenu.c` - `menu()`, `TransformGraphMenu()`, and `TestAllGraphsMenu()` signatures now include `void` to get rid of warning `no function prototype given: converting '()' to '(void)'`. Updated call to `TestAllGraphs()` from `TestAllGraphsMenu()` since we no longer need to pass `NULL` for `outputBase` ( missed cleanup in #101 )
* `c/planarityApp/planarityTestAllGraphs.c` - Removed `outputBase` parameter, and also updated docstring to remove nonexistent parameters (missed cleanup in #101 ). Also removed unused variable `numGraphsRead` from `testAllGraphs()`, since we use the `testAllStatsP` to hold the `numGraphsRead` (along with other data to output)
* `c/planarityApp/planarityTransformGraph.c` - Removed error message code, since none of them are dynamic (e.g. include filename in `ErrorMessage()` when an error is encountered). Also removed timing variables, since those are only used when we `TestAllGraphs()` (i.e. missed cleanup in #101 )
* `c/planarityApp/planarityUtils.c` - Function definition for `Reconfigure()` now includes `void` to get rid of warning `no function prototype given: converting '()' to '(void)'`. Also fixed minor bug in `SaveAsciiGraph()` error message: should include `filename`, not `outfile` (which is a `FILE *` and would be `NULL` at this point if we hit this error state)

### DIFF
--- a/c/planarityApp/planarityCommandLine.c
+++ b/c/planarityApp/planarityCommandLine.c
@@ -771,6 +771,5 @@ int callTestAllGraphs(int argc, char *argv[])
 	outfileName = argv[4 + offset];
 
 	// We don't want to write to string, so outputStr is NULL
-	// We don't need to capture whether output is 0- or 1-based, so zeroBasedOutputFlag arg is NULL
-	return TestAllGraphs(commandString, infileName, NULL, outfileName, NULL);
+	return TestAllGraphs(commandString, infileName, outfileName, NULL);
 }
--- a/c/planarityApp/planarityCommandLine.c
+++ b/c/planarityApp/planarityCommandLine.c
@@ -6,8 +6,11 @@ See the LICENSE.TXT file for licensing information.
 
 #include "planarity.h"
 
-#if defined(_MSC_VER) && !defined(__INTEL_COMPILER)
+#if defined(_MSC_VER) && !defined(__llvm__) && !defined(__INTEL_COMPILER)
 // MSVC under Windows doesn't have unistd.h, but does define functions like getcwd and chdir
+#include <direct.h>
+#define getcwd _getcwd
+#define chdir _chdir
 #else
 #include <unistd.h>
 #endif
--- a/c/graphLib/graphUtils.c
+++ b/c/graphLib/graphUtils.c
@@ -11,30 +11,30 @@ See the LICENSE.TXT file for licensing information.
 
 /* Imported functions for FUNCTION POINTERS */
 
-extern int  _EmbeddingInitialize(graphP theGraph);
-extern int  _SortVertices(graphP theGraph);
+extern int _EmbeddingInitialize(graphP theGraph);
+extern int _SortVertices(graphP theGraph);
 extern void _EmbedBackEdgeToDescendant(graphP theGraph, int RootSide, int RootVertex, int W, int WPrevLink);
 extern void _WalkUp(graphP theGraph, int v, int e);
-extern int  _WalkDown(graphP theGraph, int v, int RootVertex);
-extern int  _MergeBicomps(graphP theGraph, int v, int RootVertex, int W, int WPrevLink);
+extern int _WalkDown(graphP theGraph, int v, int RootVertex);
+extern int _MergeBicomps(graphP theGraph, int v, int RootVertex, int W, int WPrevLink);
 extern void _MergeVertex(graphP theGraph, int W, int WPrevLink, int R);
-extern int  _HandleBlockedBicomp(graphP theGraph, int v, int RootVertex, int R);
-extern int  _HandleInactiveVertex(graphP theGraph, int BicompRoot, int *pW, int *pWPrevLink);
-extern int  _MarkDFSPath(graphP theGraph, int ancestor, int descendant);
-extern int  _EmbedPostprocess(graphP theGraph, int v, int edgeEmbeddingResult);
-extern int  _CheckEmbeddingIntegrity(graphP theGraph, graphP origGraph);
-extern int  _CheckObstructionIntegrity(graphP theGraph, graphP origGraph);
-extern int  _ReadPostprocess(graphP theGraph, void *extraData, long extraDataSize);
-extern int  _WritePostprocess(graphP theGraph, void **pExtraData, long *pExtraDataSize);
+extern int _HandleBlockedBicomp(graphP theGraph, int v, int RootVertex, int R);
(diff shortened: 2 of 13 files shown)


c/planarityApp/planarityCommandLine.c
```diff
@@ -6,8 +6,11 @@ See the LICENSE.TXT file for licensing information.
 
 #include "planarity.h"
 
-#if defined(_MSC_VER) && !defined(__INTEL_COMPILER)
+#if defined(_MSC_VER) && !defined(__llvm__) && !defined(__INTEL_COMPILER)
```
john-boyer-phd, line 9: The change to add "!defined(__llvm__)" triggers two questions. 

The first is just why was it necessary to add it? In other  words, do we ever have _MSC_VER defined and __INTEL COMPILER undefined for any compiler other than the Windows Microsoft cl compiler? 

The second is a question I should've asked earlier, which is why must __INTEL COMPILER be undefined? In other words, is _MSC_VER ever defined for any compiler other than the Windows Microsoft cl compiler? In other other words, is _MSC_VER being defined sufficient to uniquely detect the Microsoft cl compiler, or does testing only for _MSC_VER being defined lead to the `#if` part for any other compiler (gcc on any platform, with or without VSCode, or clang with VSCode)?

john-boyer-phd (commented):
Some really great catches. One small question, probably nothing, and then we should be able to merge this.

c/planarityApp/planarityCommandLine.c
wbkboyer: According to the [`cpredef`](https://github.com/cpredef/predef/tree/master) [`Compilers.md`](https://github.com/cpredef/predef/blob/master/Compilers.md) documentation, we need to make sure `!defined(__INTEL_COMPILER)` because the Intel compiler defines `_MSC_VER`. Similarly, if `clang` is run on Windows as `clang.exe`, then `_MSC_VER` is defined, so we need `!defined(__llvm__) && !defined(__INTEL_COMPILER)` to make sure that block is only being run when we're building using MSVC `cl`. There might be a more streamlined test, but I figured this was the most comprehensive!!

PR #126, opened 2024-12-12
Issue #102 - Add support for MinGW-w64 and deprecate MinGW 32 

### LINKED ISSUE #102: Add support for MinGW 64 and deprecate MinGW 32
While writing the [2. Dev Setup](https://github.com/graph-algorithms/edge-addition-planarity-suite/wiki/2.-Dev-Setup) wiki page (#99), I noticed that the VSCode setup steps for [Using GCC with MinGW](https://code.visualstudio.com/docs/cpp/config-mingw) explicitly instruct users to install MinGW 64, and grew concerned that this would contradict the current instructions to install MinGW 32 (see `README.md` from commit [b4063b](https://github.com/graph-algorithms/edge-addition-planarity-suite/blob/b4063bf68d2ac6788f0c1eec42e617d4d269dbb5/README.md), to be integrated into the Dev Setup wiki page)

There's a few benefits to moving support to MinGW 64 and deprecating MinGW 32
1. The MinGW 32 installation source site, https://osdn.net/projects/mingw/releases/ , seems to be very flakey (e.g. sometimes get 502 Bad Gateway errors, exceedingly unresponsive website; got notification when downloading installer on 2024/10/05 that HTTPS cert had been expired for 780 days)
2. The most recent release of MinGW32 is from 2021, and various Reddit, StackOverflow, and GitHub issue posts seem to indicate that MinGW64 has a better track record of fixing bugs
3. MinGW 64 will include a more recent version of `gcc`
4. MinGW 64 is backwards compatible, and can produce 32-bit executables (although I don't think that's a hangup)

More info for what differentiates MinGW 64 vs. 32 is available on the [MinGW-w64](https://en.wikipedia.org/wiki/Mingw-w64) wiki page, but one point in particular stands out for our purposes:
> Better [C99](https://en.wikipedia.org/wiki/C99) support

Another [StackOverflow reply](https://stackoverflow.com/a/14580724/5456879) points to the [Feature list of mingw-w64 runtime](https://sourceforge.net/p/mingw-w64/wiki2/Feature%20list/), which possibly gives a bit more insight into the statement on the Wikipedia page:
> Headers are mostly C99 compatible.

wbkboyer:
Resolves #102 

_**N.B.**_ One must follow the updated [2. Dev Setup](https://github.com/graph-algorithms/edge-addition-planarity-suite/wiki/2.-Dev-Setup) steps to install MSYS2 and then install the MinGW-w64 toolchain.

## Updated
* `devEnvSetupAndDefaults/.vscode/tasks.json` - update `command` to use the `gcc.exe` installed as part of the MinGW-w64 toolchain
* `devEnvSetupAndDefaults/.vscode/launch.json` - update `miDebuggerPath` to use the `gdb.exe` installed as part of the MinGW-w64 toolchain 
* `README.md` - Instruct users to perform steps to build the distribution using the MSYS2 UCRT64 terminal app.

---

Confirmed that the distribution build process outlined in `README.md` successfully produces the expected `.tar.gz`.

Additionally, successfully hit breakpoints (e.g. within `planarityTestAllGraphs.c`) with the Debug executable and was able to run the Release executable from VSCode.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,9 +32,9 @@ Please refer to the [2. Dev Setup](https://github.com/graph-algorithms/edge-addi
 
 Once one has set up the development environment and is able to work with the code in the development environment, it is possible to make the distribution with the following additional steps:
 
-1. Ensure that the `autotools`, `configure`, and `make` are available on the command-line (e.g. add `C:\MinGW\msys\1.0\bin` to the system `PATH` before Windows Program Files to ensure that the `find` program is the one from `MSYS` rather than the one from Windows (e.g. adjust the `PATH` variable as needed)). 
-2. Navigate to the root of the `edge-addition-planarity-suite` repository (i.e. the directory containing `configure.ac` and the `c` subdirectory)
-3. Get into `bash` (e.g., type `bash` in the Windows command-line), then enter the following commands:
+1. Ensure that the `autotools`, `configure`, and `make` are available on the command-line (e.g. add `C:\msys64\usr\bin` to the system `PATH` before Windows Program Files to ensure that the `find` program is the one from `MSYS2` rather than the one from Windows (e.g. adjust the `PATH` variable as needed)). 
+2. Open the MSYS2 UCRT64 terminal app and navigate to the root of the `edge-addition-planarity-suite` repository (i.e. the directory containing `configure.ac` and the `c` subdirectory)
+3. Enter the following commands:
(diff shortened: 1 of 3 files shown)


john-boyer-phd (approved):
This looks great, and less painful than when I tried to do this quite a few years ago. Maybe it was just that you absorbed the pain ^_^

PR #127, opened 2024-12-12
Issue #120 -  Fixing how we determine whether the OS is Windows so that `FILE_DELIMETER` is correctly defined for MSVC `cl`

### LINKED ISSUE #120: Compiling `planarity` with MSVC `cl` on Windows causes `FILENAME` in output files to include full path rather than only the filename
## Problem statement

While undertaking #116, I observed that when I went to run the `test_table_generator.py` on the output files produced by `planarity_testAllGraphs_orchestrator.py` with the executable being one which had been compiled with MSVC `cl`, I would fail to parse the headers of the planarity output files. This is because instead of 
```
FILENAME="n{N}.m{M}.g6" DURATION="X.XXX"
...
```
The header included the full path to the input file, i.e.
```
FILENAME="{input_dir}\n{N}.m{M}.g6" DURATION="X.XXX"
...
```

## Possible cause
In `c\planarityApp\planarityTestAllGraphs.c` function `outputTestAllGraphsResults()`, we determine the index of the final slash; if it exists, then we set the `infileBaseName` pointer to the next index, otherwise just set `infileBaseName` to the entire `infileName`:

```
char *finalSlash = strrchr(infileName, FILE_DELIMITER);
char *infileBasename = finalSlash ? (finalSlash + 1) : infileName;
```

It's likely the case that `strrchr()` behaves differently on MSVC than it does with the other compiler toolchains.

## In the meanwhile...
Either use VSCode's search and replace feature to get rid of the parent directory or some other command-line `grep` magic and retry running `test_table_generator.py` on the planarity output files.

## References
* [MSVC docs - `strrchr`, `wcsrchr`, `_mbsrchr`, `_mbsrchr_l`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strrchr-wcsrchr-mbsrchr-mbsrchr-l?view=msvc-170)

wbkboyer:
Resolves #120 

## Summary

`FILE_DELIMETER` was being defined incorrectly due to incorrect OS check, which caused `outputTestAllGraphsResults()` in `planarityTestAllGraphs.c` to fail in finding the `finalSlash` within `infileName`.

## Updated
* `c/graphLib/lowLevelUtils/appconst.h` - Should check if `_WIN32` is defined to determine if the OS is Windows according to the Pre-defined Compiler Macros Wiki page on [Operating Systems](https://sourceforge.net/p/predef/wiki/OperatingSystems/), which itself links to [64-Bit Compiler - Predefined macros](https://learn.microsoft.com/en-us/windows-hardware/drivers/kernel/64-bit-compiler?redirectedfrom=MSDN#predefined-macros).
* `c/graphLib/lowLevelUtils/platformTime.h` - should check against `WINDOWS` rather than `WIN32` for the above reason, as well as the fact that `platformTime.h` is only `#include`'d by `appconst.h`, which `#define`'s `WINDOWS` before including `platformTime.h`
* `devEnvSetupAndDefaults/.vscode/tasks.json` - added warning switches to ignore warnings thrown by external headers

### DIFF
--- a/c/graphLib/lowLevelUtils/appconst.h
+++ b/c/graphLib/lowLevelUtils/appconst.h
@@ -11,7 +11,7 @@ See the LICENSE.TXT file for licensing information.
 
 // NOTE: This is defined on 32- and 64-bit Windows systems; see
 // https://sourceforge.net/p/predef/wiki/OperatingSystems
-#ifdef _WIN32
+#if defined(WIN32) || defined(_WIN32)
 #define WINDOWS
 #endif
 
--- a/c/graphLib/lowLevelUtils/appconst.h
+++ b/c/graphLib/lowLevelUtils/appconst.h
@@ -9,7 +9,9 @@ See the LICENSE.TXT file for licensing information.
 
 #include "apiutils.h"
 
-#ifdef WIN32
+// NOTE: This is defined on 32- and 64-bit Windows systems; see
+// https://sourceforge.net/p/predef/wiki/OperatingSystems
+#ifdef _WIN32
 #define WINDOWS
 #endif
 
--- a/c/graphLib/lowLevelUtils/platformTime.h
+++ b/c/graphLib/lowLevelUtils/platformTime.h
@@ -7,20 +7,23 @@ All rights reserved.
 See the LICENSE.TXT file for licensing information.
 */
 
-#ifdef WIN32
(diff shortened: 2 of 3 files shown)


john-boyer-phd (approved):
This looks great, esp. checking for both WIN32 (for legacy compile reasons) and not just _WIN32, as well as also turning of the MS cl.exe warnings for external headers that we have no control over.

PR #128, opened 2024-12-13
Issue #71 - Add platform-appropriate line terminator between `.g6` encoded graphs in `.g6` output file

### LINKED ISSUE #71: `G6WriteIterator` - When writing multiple graphs to a single `.g6` file, no newline is inserted between graph encodings
Proposed solution: in `_printEncodedGraph()`, add a call to `sf_fputs()` so that a newline is added to the `strOrFile` contents (either `FILE *` or to the `char *`).

wbkboyer:
Resolves #71 

## Updated
* `c/graphLib/io/g6-write-iterator.c` - auto-formatting fixed up inconsistencies in whitespace etc.; in `_printEncodedGraph()`, after writing the `pG6WriteIterator->currGraphBuff` to `pG6WriteIterator->g6Output`, added another call to `sf_fputs()` to write `"\n"` to the `strOrFile` so that only one `.g6` encoded graph appears per line.

---

Verified that underlying `fputs()` call to write `"\n"` to file outputs the correct line terminator per-platform:
* On MacOS, each line was terminated with `\n`
* On Windows (both MinGW-w64 `gcc` and MSVC `cl`), each line was terminated with `\r\n`
* On Debian Linux, each line was terminated with `\n`

### DIFF
--- a/c/graphLib/io/g6-write-iterator.c
+++ b/c/graphLib/io/g6-write-iterator.c
@@ -10,7 +10,6 @@ See the LICENSE.TXT file for licensing information.
 #include "g6-write-iterator.h"
 #include "g6-api-utilities.h"
 
-
 int allocateG6WriteIterator(G6WriteIterator **ppG6WriteIterator, graphP pGraph)
 {
 	int exitCode = OK;
@@ -23,7 +22,7 @@ int allocateG6WriteIterator(G6WriteIterator **ppG6WriteIterator, graphP pGraph)
 
 	// numGraphsWritten, graphOrder, numCharsForGraphOrder,
 	// numCharsForGraphEncoding, and currGraphBuffSize all set to 0
-	(*ppG6WriteIterator) = (G6WriteIterator *) calloc(1, sizeof(G6WriteIterator));
+	(*ppG6WriteIterator) = (G6WriteIterator *)calloc(1, sizeof(G6WriteIterator));
 
 	if ((*ppG6WriteIterator) == NULL)
 	{
@@ -99,7 +98,7 @@ int getPointerToGraphToWrite(G6WriteIterator *pG6WriteIterator, graphP *ppGraph)
 	return OK;
 }
 
-int getGraphBuff(G6WriteIterator *pG6WriteIterator, char**ppCurrGraphBuff)
+int getGraphBuff(G6WriteIterator *pG6WriteIterator, char **ppCurrGraphBuff)
 {
 	if (pG6WriteIterator == NULL)
 	{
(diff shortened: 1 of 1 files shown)


c/graphLib/io/g6-write-iterator.c
```diff
@@ -461,6 +467,12 @@ int _printEncodedGraph(G6WriteIterator *pG6WriteIterator)
 		exitCode = NOTOK;
 	}
 
+	if (sf_fputs("\n", pG6WriteIterator->g6Output) < 0)
```
john-boyer-phd, line 470: This is a good add. One ask is if you could run an experiment on Windows to see if instead using "\r\n" will write "\r\n" (correct) or "\r\r\n" (incorrect). If the output is correct ("\r\n" produces "\r\n" in the file output), then it would be better to amend this PR as follows:
1. add a LINE_DELIMITER macro in appconst.h just below FILE_DELIMITER and set equal to "\r\n" for WINDOWS and "\n" otherwise.
2. Use LINE_DELIMITER in this call instead of just "\n"

The advantage of this approach, if the output is correct, is that it will write "\r\n" when you are writing to string on Windows, and not just when writing to file.

john-boyer-phd (changes requested):
Just one thing to experiment with...

c/graphLib/io/g6-write-iterator.c
wbkboyer: When I wrote

```
sf_fputs("\r\n", pG6WriteIterator->g6Output)
```
when `pG6WriteIterator->g6Output` contains a `FILE *`, I observed that the resulting `.g6` output file converted both characters to `CRLF` (i.e. two `CRLF between each graph).

When writing strings *only* (i.e. if writing to file, then allow `fputs()` to continue doing it's thing), should I try to replace any line terminator within the string (e.g. `\r`, `\n`, or `\r\n`) with the `LINE_DELIMITER` character that's based on whether `#ifdef WINDOWS`, or should I leave as-is? 🎉

john-boyer-phd: Ah, ok, thanks for checking. I've been reluctant for many years to add this LINE_DELIMITER business, and even though it's disappointing that `fputs()` did what I consider to be the wrong thing (writing two CRLFs in response to having one CRLF in the string), it is kind of a relief to not have to add LINE_DELIMITER now. The right answer in your `sf_fputs()` call is to use `"\n"`, as you have done, because the methods we're calling (like the real `fputs()` and `printf()`) are turning `\n` into a CRLF on windows, as we need them to do. 

It's a good question whether the string side of `sf_fputs()` should be converting `\n` to CRLF. The right answer is to leave it as is, which seems counterintuitive from the standpoint of wanting to be consistent with methods of the language, like `fputs()` and `printf()`. However, we're still on the outside of those language methods, so if we think of how the strings we write will get used, they will go into those language methods, which will then turn our CRLFs into double CRLFs. 

So now the only question is how I can turn off the "Change requested" flag on this PR because I should have just submitted it as a review comment without requesting a change yet, since the ask was for an experiment first, and the net result of that experiment is that we should merge this PR without a change.

Issue #130, opened 2024-12-17
The inclusion of `graphLib.h` should include all `graphLib` headers


john-boyer-phd:
While continuing development of the `planarity` app that consumes and demonstrates the `graphLib` capabilities, we noticed the need to include some headers from lower levels of the `graphLib`. This included graph read and write iterators as well as certain other new lower-level functionality. It's best to add these headers so that including `graphLib.h` ensures that everything in the `graphLib` that could be called from an app layer like `planarity` is available.

Issue #104, opened 2024-10-04
Ensure consistent formatting of C source files


wbkboyer:
With VSCode + the C/C++ Extension, at some point `formatOnSave` was enabled in `.vscode/settings.json` (see [Editing and Navigating C++ CodeCode formatting](https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting)) and so I noticed that there were whitespace changes being made throughout files that I touched. These changes do make the code easier to read, and ensure that the files are rendered consistently on GitHub, so it is desirous that all files should be formatted consistently.

wbkboyer:
Can likely resolve alongside #27

john-boyer-phd:
Now that we have VS Code in our pockets for over a year now, we're not going back! So lets focus on VS Code and not worry about the Eclipse side of this issue.

wbkboyer:
One possible solution would be to install the VSCode extension [Format All Files in Workspace](https://marketplace.visualstudio.com/items?itemName=alexr00.formatallfilesinworkspace), which formats each file with the relevant formatter extension; for C files, this would mean using [`clang-format`](https://clang.llvm.org/docs/ClangFormat.html) (the default used by the [VSCode C/C++ Extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools), according to [Code formatting](https://code.visualstudio.com/docs/cpp/cpp-ide#_code-formatting)), and the [`black`](https://marketplace.visualstudio.com/items?itemName=ms-python.black-formatter) for Python files (configuration provided in default `settings.json` to be added with #112; see #100 for background). Then, the format operation would be performed as a one-off (i.e. this extension won't need to be added to the [2. Dev Setup - Installing Extensions](https://github.com/graph-algorithms/edge-addition-planarity-suite/wiki/2.-Dev-Setup#installing-extensions) section).

There are some available configuration options that one may add to `settings.json`, i.e. `formatAll.includeFileExtensions` and `formatAll.excludeFolders`, but we might as well force format all files.

john-boyer-phd:
We have sunset Eclipse-based dev, so I believe this can be closed now. Do you agree @wbkboyer ? If so, you can have the pleasure of doing so, since you did all the work of getting us over to VSCode.

wbkboyer:
[...]

PR #131, opened 2024-12-19
Issue129 write random graphs to g6

### LINKED ISSUE #129: Write graphs generated by `RandomGraphs()` to `.g6` output file
## Background

During the undertaking of #71 , since there's no endpoint in `edge-additon-planarity-suite` that writes multiple `.g6` encoded graphs to the same file (i.e. `TransformGraph()` is the only place where we're writing a `.g6` output file, which takes a single graph from an input file), I had to work within the old `planarity-backup` code to which I had added the `G6WriteIterator` infrastructure to write out the graphs generated by the embedded `makeg` (see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/42#issuecomment-2125583338) on #42 ; side-note: this is the point where I had actually discovered and fixed #71).

This means that I have not tested the code for #128 within the modern `edge-addition-planarity-suite`. Additionally, since the `planarity-backup` code only deals with writing to `FILE` rather than to string, I haven't been able to verify what happens when we try to write multiple graphs to a string output (i.e. `fputs()` converts `"\n"` to the correct line terminator for the current platform: `\n` for MacOS and Debian Linux, `\r\n` for Windows (for both MinGW-w64 `gcc` and MSVC `cl`), but if the `strOrFile` contains a `char *`, then we'll be writing `\n` to the string regardless of platform...)

## Problem statement
Add additional optional parameters to `RandomGraphs()` so that `callRandomGraphs()` outputs the generated graphs to a `.g6` output file. This will demonstrate that the fix to #71 from #125 adds the expected intervening line terminator between graphs in the `.g6` output.

wbkboyer:
Resolves #129 and #130

## Updated
* `.gitignore` - Comment informing users that 
    ```
    **/random/
    **/embedded/
    **/adjlist/
    **/obstructed/
    **/error/
    ```
    Are directories to be ignored, as they are the default output directories for `RandomGraphs()`
* `devEnvSetupAndDefaults/devEnvSetup.sh` - Now creates the default output directories for `RandomGraphs()` at the root of the project; verified in MSYS2 UCRT shell on Windows 10 (22H2), MacOS `zsh`, and Debian Linux `bash`
* `c/samples/` - Added `\n` line terminator to `.g6` samples (both those used as input for transformation from `.g6` to other graph formats, and those which are the expected output for transforming from other formats to `.g6`)
* `c/graphLib/graph.h` - now include `io` headers so that they will be included where they're needed in the `planarityApp` layer
* `c/graphLib/graphLib.h` - added whitespace for readability
* `c/graphLib/graphStructures.h` - added whitespace for readability, moved `#include`s statements so as to be alphabetical, and removed `#include` of `strbuf.h` (`strBufP` is not used as a parameter to any functions or macros, nor does it appear as a member of the `struct`s defined in this header. Also verified the project compiles on all supported platforms and successfully ran through manual spot tests so I think this is safe)
* `c/graphLib/io/graphIO.c` - removed `#include`s of `io` headers, since we get those through `graph.h`
* `c/graphLib/lowLevelUtils/apiutils.c` - added newline at end of source file ([justification](https://dev.to/documendous/why-you-should-end-your-source-files-with-a-new-line-156g)); originally only touched this file because I had intended to move a helper function here, but I decided to keep this change in this PR.
* `c/graphLib/lowLevelUtils/apiutils.h` - added whitespace for readability (done automatically by VSCode; see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/104#issuecomment-2442674739) on #104 ); originally had only touched this because I was going to add a helper function declaration here, but decided to keep this change in the PR.
* `c/planarityApp/planarity.1` - updated docs for the parameterization of `planarity` that allows one to `callRandomGraphs()` to include the optional `OUTPUT` file, which is corresponds to the `outfileName` parameter
* `c/planarityApp/planarity.h` - updated the `RandomGraphs()` declaration to include `char *outfileName`, added `extern` for `OrigOutFormat` so that `menu()` of `planarityMenu.c` has access to this value that's set when we `Reconfigure()` (along with the original options), and added declaration for helper function `GetNumCharsToReprInt()` (moved from `planarityTestAllGraphs.c` as I as originally going to call this in a calculation in `planarityRandomGraphs.c`, but it proved unnecessary at this time; however, it may prove useful in the future for other parts of the application, and therefore belongs in a central location)
* `c/planarityApp/planarityCommandLine.c` - updated docstring for `callRandomGraphs()` to include the optional output file parameter in square brackets, added check that no more than 7 command-line args are expected by `callRandomGraphs()`, and that we should only extract the optional output file parameter if `argc` indicates that that additional parameter exists
    * _**N.B.**_ [This](https://stackoverflow.com/a/16419021) StackOverflow post (and [another](https://stackoverflow.com/a/18547129); see [Programming languages — C](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n3047.pdf)) indicates that the C standard requires `argv[argc]` to be `NULL`, which I verified in the debugger (i.e. `outfileName = argv[5 + offset];` resulted in `outfileName` being assigned`NULL`). I feel like it doesn't hurt to check, but maybe because this is part of the standard I should revert it...
* `c/planarityApp/planarityHelp.c` - added optional `[O]` parameter to Random Graphs help message
* `c/planarityApp/planarityMenu.c` - added `NULL` for `outfileName` arg in call to `RandomGraphs()` so that the output filename will be generated for the user based on their input values.
* `c/planarityApp/planarityRandomGraphs.c` - whitespace automatically added by VSCode automatic code formatting. Changed stack-allocated `theFilename` to have size `MAXLINE` (_**N.B.**_ only did this to get rid of magic number in code, but maybe we should create a different `#define` for this :S). Added hooks into `G6WriteIterator` framework that only occur:
    1. If the user is interacting with the `planarity` Menu interface, they must first choose `r` to `Reconfigure()`, then:
        1. Choose `r` corresponding to the option "Randomly generate graphs (r)"
        2. Choose `y` when prompted "Do you want original graphs in directory 'random'? (y/n)"
        3. Choose `g` when prompted ""Do you want to output generated graphs to Adjacency List (last 10 only) or to G6 (all)? (a/g)"
        4. Choose arbitrary options for other configuration until you return to main Menu
        5. Choose one of the graph algorithm command specifiers
        6. Enter number of graphs to generate
        7. Enter size of graphs
        Observe the summary statistics output to terminal, and the `.g6` file containing the random graphs generated will be available at `${workspaceFolder}/random/n<SizeOfGraphs>.k<NumGraphs>.g6`; these are the graphs used to produce the summary statistics.
    2. If the user supplies the optional `[O]` planarity random graphs command line parameter, `callRandomGraphs()` passes non-`NULL` `outfileName` extracted from `argv` to `RandomGraphs()`; after the summary statistics are output to terminal, the `.g6` file containing the random graphs generated will be available at this path (as long as you had write permissions and nothing errored out).
* `c/planarityApp/planarityTestAllGraphs.c` - removed unnecessary `#include`s. In `outputTestAllGraphsResults()`, now call `GetNumCharsToReprInt()` helper function that was moved out of this `.c` source file. Fixed a couple of instances where the calculation of the characters available to display the output filename in error messages was incorrect.
* `c/planarityApp/planarityTransformGraph.c` - removed unnecessary `#include`s (tested transform functionality works as expected by running `-test` and fixing the `.g6` input/output files; see above)
* `c/planarityApp/planarityUtils.c` - Updated `Reconfigure()` machinery to expose option to write all generated graphs to `.g6` output file


c/planarityApp/planarityCommandLine.c
```diff
@@ -598,13 +598,14 @@ int runGraphTransformationTest(char *command, char *infileName, int inputInMemFl
...
-// 'planarity -r [-q] C K N': Random graphs
+// 'planarity -r [-q] C K N [O]': Random graphs
 int callRandomGraphs(int argc, char *argv[])
 {
 	char Choice = 0;
 	int offset = 0, NumGraphs, SizeOfGraphs;
+	char *outfileName = NULL;
 
-	if (argc < 5)
+	if (argc < 5 || argc > 7)
```
john-boyer-phd, line 608: Good catch

c/planarityApp/planarityRandomGraphs.c
```diff
@@ -19,22 +19,22 @@ graphP MakeGraph(int Size, char command);
...
-int  ObstructionMinorFreqs[NUM_MINORS];
-graphP theGraph=NULL, origGraph=NULL;
-platform_time start, end;
-int embedFlags = GetEmbedFlags(command);
-int ReuseGraphs = TRUE;
-int writeResult;
-int writeErrorReported_Random=FALSE, writeErrorReported_Embedded=FALSE,
-	writeErrorReported_AdjList=FALSE, writeErrorReported_Obstructed=FALSE,
-	writeErrorReported_Error=FALSE;
+	char theFileName[MAXLINE + 1];
```
john-boyer-phd, line 26: Excellent catch (MAXLINE)

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -329,8 +315,8 @@ int outputTestAllGraphsResults(char command, testAllStatsP stats, char *infileNa
 		FILE *outputFileP = fopen(outfileName, "w");
 		if (outputFileP == NULL)
 		{
-			charsAvailForFilename = (int)(MAXLINE - strlen(outfileName));
 			messageFormat = "Unable to open file \"%.*s\" for output.\n";
+			charsAvailForFilename = (int)(MAXLINE - strlen(messageFormat));
```
john-boyer-phd, line 319: Good catch on the strlen arg (above and below, too).

john-boyer-phd (approved):
This all looks great!

PR #132, opened 2024-12-19
Issue #98 - Build format string to use with `scanf()` when prompting user for filenames in `TransformGraphMenu()` and `TestAllGraphsMenu()`

### LINKED ISSUE #98: On Debian 12 with gcc (Debian 12.2.0-14) 12.2.0,`fflush(stdin)` before `fgets()` doesn't get rid of `\n` left in input buffer by prior `scanf()`
When testing changes from #92 on Debian 12 compiled using `gcc` (Debian 12.2.0-14) 12.2.0 (either via VSCode `tasks.json` or `automake` build system), I observed that when using the command-line menu system (i.e. running `planarity` executable without any arguments), if you choose `T` for "Test graph functionality", then either `T` (transform) or `A` (test all graphs), the prompt for the `infileName` is immediately dismissed and `infileName` only contains `\n` (i.e. `fgets()` gets a `\n` that was not discarded by `fflush(stdin)`). This did not manifest when running the `planarity` executable produced by `clang` on MacOS, by MinGW `gcc` nor MSVC `cl` on Windows, nor by Eclipse CDT on Windows.

It turns out that on some platforms, `fflush(stdin)` is a no-op (see [here](https://c-faq.com/stdio/stdinflush2.html)). One possible (portable) solution is to add:

```
while((c = getchar()) != '\n' && c != EOF)
    /* discard */ ;
```

at some point before subsequent `fgets()` or `getc()` calls (from [this](https://stackoverflow.com/a/2187514/5456879) StackOverflow response). One possible location would be within `Prompt()` of `c/planarityApp/planarityUtils.c`, since this should always be called before we try to obtain user input.

[This](https://stackoverflow.com/a/59119187/5456879) StackOverflow response helped to complete the picture: because `TransformOrTestMenu()` uses `scanf()` to get the user choice before calling either `TransformMenu()` or `TestMenu()`, there's a `\n` left in the `stdin` buffer which, on certain platforms, is not discarded when we try to `fflush(stdin)` before `fgets()` in `TransformMenu()` or `TestMenu()`.

wbkboyer:
Resolves #98 

## Updated
* `c/planarityApp/planarityMenu.c` - When prompting users for filenames, I first construct a format string that limits the number of characters being read using `scanf()`, then pass this constructed format string to `scanf()` when trying to populate `infileName` and `outfileName` in both `TransformGraphMenu()` and `TestAllGraphsMenu()`

_**N.B.**_ The changes to `c/planarityApp/planarity.h` and `c/planarityApp/planarityUtils.c` are changes that will come from merging #131 

---

Verified both Transform Graph and Test All Graphs menu interactions on MacOS (`clang`), Windows (MinGW-w64 `gcc` and MSVC `cl`), and Debian Linux (`gcc`).

### DIFF
--- a/c/planarityApp/planarity.h
+++ b/c/planarityApp/planarity.h
@@ -68,6 +68,8 @@ extern "C"
     int ConstructTransformationExpectedResultFilename(char *infileName, char **outfileName, char command, int actualOrExpectedFlag);
     void WriteAlgorithmResults(graphP theGraph, int Result, char command, platform_time start, platform_time end, char *infileName);
 
+    int GetNumCharsToReprInt(int theNum);
+
 #ifdef __cplusplus
 }
 #endif
--- a/c/planarityApp/planarityMenu.c
+++ b/c/planarityApp/planarityMenu.c
@@ -109,11 +109,15 @@ void TransformGraphMenu(void)
     char commandStr[4];
     commandStr[0] = '\0';
 
+    char *fileNameFormatFormat = " %%%d[^\r\n]";
+    char *fileNameFormat = (char *)malloc((strlen(fileNameFormatFormat) + GetNumCharsToReprInt(MAXLINE) + 1) * sizeof(char));
+    sprintf(fileNameFormat, fileNameFormatFormat, MAXLINE);
+
     do
     {
         Prompt("Enter input filename:\n");
         fflush(stdin);
-        fgets(infileName, MAXLINE, stdin);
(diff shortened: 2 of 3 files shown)


john-boyer-phd (approved):
It's crazy we can't reliably `fflush()` on `stdin` and even crazier that there is no way to easily soak up the trailing line delimiter of one input at the beginning of the next by using variants of `fget(stdin)`. Given that, this is a very good fix.

PR #133, opened 2024-12-19
Epic #42 - Updated test tables for `N=11`

### LINKED ISSUE #42: EPIC: Investigate Operation of K_{3,3} Search Implementation

wbkboyer:
Resolves #42 

## Updates
* Re-ran `TestSupport/planaritytesting/planarity_testAllGraphs_orchestrator.py` and then `TestSupport/planaritytesting/test_table_generator.py` for each of the graph algorithm commands to regenerate the tables from the `HEAD` of `master`. All tables save for `K_{3,3}` search have the same `OK`/`NONEMBEDDABLE` counts as was produced by `planarity` as of 2.X, and it looks like we're catching a heck of a lot more `K_{3,3}`s now!


john-boyer-phd (approved):
Looks great! How about those new K_{3,3} numbers !

PR #140, opened 2025-02-10
Issue #57 -  Recover responsiveness to stdin and upgrade strOrFile

### LINKED ISSUE #57: EPIC - Recover responsiveness to stdin and upgrade strOrFile
Prior to adding `graph6` support, the command-line (and ultimately `gp_Read()`) were responsive to piping in a graph file on standard input. When input filename of `stdin` could be used in the `I` command-line parameter, a `FILE` pointer is assigned to `stdin` rather than the result of `fopen()`. 

When the `graph6` support was added, it was necessary to interrogate the input more deeply to determine whether the file was a   `graph6` file or one of the earlier supported formats. The interrogation is incompatible with `stdin`. Specifically, it used to be possible to call `getc()` and determine the file format based on the first character, then use `ungetc()` to allow the reader for the format to read the whole file contents including the first character. Now, it is necessary to read a first line of content (we read up to 255 characters even though we currently need only 5 to distinguish a LEDA file from a `graph6` file). Once we have read the line, we cannot be guaranteed on all platforms and compilers that multiple `unget()` invocations will be successful, so we now use `fseek()` to return to the beginning of the file. However, the `fseek` does not work with `stdin`. 

To recover the capability to input from `stdin`, it will first be necessary to upgrade the new `strOrFile` object with a special `ungetContent()` method that can place a given string of characters into a memory buffer. This would overcome the limitation of `ungetc()` that only one invocation is guaranteed to work. However, the other reading methods in `strOrFile` would have to be taught to use the memory buffer before resorting to `FILE` operations. Also, the many reading methods of `strBuf` would undoubtedly need to be merged into `strOrFile`. Then, all graph readers in `graphIO.c` must then be taught to use the `strOrFile` object rather than a `FILE` object. Furthermore, the updated `strOrFile` should be used preferentially to `strBuf`. This should remove separate `FILE`/`strBuf` code for doing the same things in the graph readers. Ultimately, `strBuf` should be removed.

While upgrading `strOrFile` as above, also upgrade it to remove construction with a given `FILE` pointer. Instead, those cases should just receive a file name and file mode (reading or writing) and then assign an internal `FILE` pointer to the result of `fopen` (or stdin, stdout, or stderr). There should no occurrences of `FILE *` except possibly in some non-API code in `planarityUtils.c`.  Then, it should be easy to improve `strOrFile` in the way described in Issue #56.

wbkboyer:
Resolves #57 

## Added

* `c/samples/nauty_example.LEDA` - handwritten [LEDA format](https://www.algorithmic-solutions.info/leda_guide/graphs/leda_native_graph_fileformat.html) translation of the `nauty` example graph from section 20 (p.82) of the [manual](https://users.cecs.anu.edu.au/~bdm/nauty/nug28.pdf), which was used to manually test `_ReadLEDAGraph()`

### Changes to the `strOrFile` container and supporting `sf_[a-zA-Z]+()` functions

<details>
<summary>c/graphLib/io/strOrFile.h</summary>

* Updated `strOrFile` `struct` to include `strBufP` for `theStr`, which eliminates the need for `theStrPos`. Also adds `containerType`, which allows us to swiftly check whether the container is an `INPUT_CONTAINER` or `OUTPUT_CONTAINER` when validating whether a particular operation is applicable to the container type. Finally, the `ungetBuf`, a `stackP`, was added to manage unget operations on `INPUT_CONTAINER`s
* Updated the declaration of `sf_New()` to indicate it now accepts a string for the filename to open, if applicable, and a string representing the `ioMode`, in addition to the optional `theStr`
* Added declarations for helper functions required by the `_Read[a-zA-Z]+()` and `_Write[a-zA-Z]+()` functions
</details>

<details>
<summary>c/graphLib/io/strOrFile.c</summary>

* `sf_New()` - If the `strOrFile` container is meant to contain a `FILE *`, it now handles opening the file corresponding to `fileName` with the correct `ioMode` (either `READTEXT` or `WRITETEXT`) and assigns to `pFile`. This logic also handles when the user specifies some input/output stream (i.e. `stdin`, `stdout`, `stderr`). Otherwise, if the container is meant to hold a string, it initializes a `strBufP` and appends the initial `theStr` (as a follow-up to [this question](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/140/files#r1951740625), if `theStr` is not `NULL` and the `containerType` is `OUTPUT_CONTAINER`, then we error out). Regardless of the container type, an `ungetBuf` `stackP` is allocated, although only in the case of `INPUT_CONTAINER`s do we interact with it: it is used to contain characters that are ungotten (either by `sf_ungetc()` or `sf_ungets()`) and is consumed first when we `sf_getc()` or `sf_fgets()` (in [this question](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/140/files#r1951749135), I asked whether the `ungetBuf` should only be allocated for `INPUT_CONTAINER` (i.e. when `ioMode` is `READTEXT`). Upon further reflection, even if it's not being used, we should allocate the `ungetBuf` so that there's no illegal memory access issues).
* `sf_getc()` - if the `ungetBuf` `stackP` is nonempty, `sp_Pop()`s a `char`; otherwise, `getc()` from the `FILE *` or get the character at the current `ReadPos` of `theStr` (a `strBufP`).
* `sf_ReadSkipChar()`, `sf_ReadSkipWhitespace()`, `sf_ReadSingleDigit()`, `sf_ReadInteger()`, and `sf_ReadSkipInteger()` all use `sf_getc()` to advance through the `strOrFile` container
* `sf_ReadSkipLineRemainder()` uses `sf_fgets()` to advance through the file container and consume up to and including the next `\n`
* `sf_ungetc()` -ensures the container is a valid `INPUT_CONTAINER` before `sp_Push()`ing `theChar` onto the `ungetBuf` `stackP`
* `sf_ungets()` - ensures the container is a valid `INPUT_CONTAINER` before `sp_Push()`ing each character of `strToUnget` onto the `ungetBuf` `stackP` *in reverse order* so that they can be fetched from the `ungetBuf` in the order of the original string.
* `sf_fgets()` - ensures the container is a valid `INPUT_CONTAINER` before it obtains as many characters as possible from the `ungetBuf` before switching over to the `FILE *` or `strBufP` in the `strOrFile` container (if no `\n` encountered and there's still characters desired by the caller)
* `sf_fputs()` - ensures the container is a valid `OUTPUT_CONTAINER` before it calls `fputs()` if the container has a `FILE *` or `sb_ConcatStr()` if it contains `theStr`.
* `sf_takeTheStr` - now uses `sb_TakeString()`, since `theStr` is a `strBufP`
* `sf_closeFile()` - if the `strOrFile` contains a `FILE *`, it's only appropriate to `sp_Free()` the `ungetBuf`, since it's no longer valid.
* `sf_Free()` - Now calls `sb_Free()` on `theStr` (a `strBufP`) rather than `free()` (since `theStr` used to correspond to a `char *`), and frees the `ungetBuf` `stackP` using `sp_Free()` if it has not yet been freed.

</details>

### Changes to `strOrFile` dependencies

* `c/graphLib/io/strbuf.h` - added new macro to `sb_GetUnreadCharCount()`: we only want to see how many characters have not yet been read (e.g. used to ensure there's at least one character available to read in `sf_getc()` and `sf_fgets()`)
* `c/graphLib/lowLevelUtils/apiutils.h` - added `MAXCHARSFOR32BITINT` so that, for example, `sf_ReadInteger()` can allocate sufficient memory for the `intCandidateStr`. Other functions needing to represent 32-bit integers as strings should use this same constant.

### Using the `strOrFile` container for input and output

The original intent of this refactor was to restore being able to read from `stdin`, which was lost when we changed how we interrogated the input files (reading an entire line from file vs. reading a single character). As an added bonus of this refactor, the logic of `gp_Read(FromString)?()` and `gp_Write(ToString)?()` functions has been simplified and unified in the respective `_(Read|Write)Graph()` functions and the logic of the `_(Read|Write)[a-zA-Z]+()` functions no longer require branching logic to handle whether or not the `(input|output)Str` or `(in|out)fileName` was `NULL`.

<details>
<summary>c/graphLib/io/graphIO.c</summary>

* `gp_Read()` - Now creates a `strOrFileP` to handle opening the file corresponding to `FileName` for `READTEXT` and calls `_ReadGraph()`
* `gp_ReadFromString()` - Now creates a `strOrFileP` to contain the `inputStr` and calls `_ReadGraph()`
* `_ReadGraph()` - Applies the relevant `sf_[a-zA-Z]+()` functions to interrogate the contents of the `inputContainer` to determine the input filetype and to call the corresponding `_Read[a-zA-Z]+()`. If extra data is allowed for the input type (i.e. only after successfully reading Adjacency List or Adjacency Matrix), create a `strBufP` to contain the `extraData`, which is read from the `inputContainer` line-by-line using `sf_fgets()` and concatenated using `sb_ConcatString()`.
* `gp_Write()` - Now creates a `strOrFileP` to handle opening the file corresponding to `FileName` for `WRITETEXT` and calls `_WriteGraph()`
* `gp_WriteToString()` - Now creates a `strOrFileP` with `theStr` initially empty to send to `_WriteGraph()` alongside the `pOutputStr`; if writing to the `outputContainer` was successful *AND* if the `outputContainer` has *not* yet been cleaned up (as is the case for writing Adjacency List and Adjacency Matrix), then `sf_takeTheStr()` and assign the pointer to `(*pOutputStr)`
* `_WriteGraph()` - Based on the `Mode`, determines which `_Write[a-zA-Z]+()` function should be called and passes in the (dereferenced) `outputContainer`. If we're writing to `.g6`, then `_WriteGraphToG6StrOrFile()` will have already taken the string from the `outputContainer` and assigned it to `pOutputStr`, so we null out the reference to `outputContainer` to prevent `gp_WriteToString()` from trying to take the string again from the `outputContainer`, or to double-free the `outputContainer`. Then, if extra data is allowed to be written (i.e. assuming we have run `gp_Embed()` on `theGraph` with a graph algorithm extension attached for which `fpWritePostprocess()` has been overridden, e.g. `_DrawPlanar_WritePostprocess()`), generates the extra data as a `char *` and uses `sf_fputs()` to write it to the `outputContainer`.
* Function signatures for all of `_ReadAdjMatrix()`, `_ReadAdjList()`, `_ReadLEDAGraph()`, `_WriteAdjList()`, `_WriteAdjMatrix()`, and `_WriteDebugInfo()` all updated so that they accept only the input/output `strOrFile` container and perform the relevant `sf_[a-zA-Z]+()` functions for read/write operations.

</details>

<details>
<summary>c/graphLib/io/g6-read-iterator.c/.h</summary>

* Deprecated `beginG6ReadIterationFromG6FilePath()`, `beginG6ReadIterationFromG6FilePointer()`, and `beginG6ReadIterationFromG6String()` in favour of the new `beginG6ReadIterationFromG6StrOrFile()`
* Deprecated `_ReadGraphFromG6FilePointer()` and updated `_ReadGraphFromG6FilePath()` and `_ReadGraphFromG6String()` helper functions to call the new `_ReadGraphFromG6StrOrFile()` helper function.
* The new `beginG6ReadIterationFromG6StrOrFile()` function validates the `g6InputContainer` and assigns it to the `pG6ReadIterator` before calling `_beginG6ReadIteration()`
* `_beginG6ReadIteration()` now initializes `charConfirmation` to `EOF` (since value of `EOF` is implementation-dependent, although it's commonly set to `-1`)
* `endG6ReadIteration()` now no longer "knows" whether the `g6Input` `strOrFileP` holds a string or a file, so remove the explicit call to `sf_closeFile()` and allow `sf_Free()` to clean that up.
* `_ReadGraphFromG6FilePath()` now creates an `inputContainer` to handle opening the file corresponding to the `pathToG6File` before calling `_ReadGraphFromG6StrOrFile()`
* `_ReadGraphFromG6String()` now creates an `inputContainer` to handle reading the graph from the `g6EncodedString` before calling `_ReadGraphFromG6StrOrFile()`
* The new `_ReadGraphFromG6StrOrFile()` helper function accepts the `g6InputContainer`, ensures it's valid, and then allocates the `G6ReadIterator` before calling `beginG6ReadIterationFromG6StrOrFile()` and proceeding with the remaining flow (i.e. `readGraphUsingG6ReadIterator()`, `endG6ReadIteration()`, and `freeG6ReadIterator()`)

</details>

<details>
<summary>c/graphLib/io/g6-write-iterator.c/.h</summary>

* Deprecated `beginG6WriteIterationToG6FilePath()`, `beginG6WriteIterationToG6FilePointer()`, and `beginG6WriteIterationToG6String()` in favour of the new `beginG6WriteIterationToG6StrOrFile()`
* Deprecated `_WriteGraphToG6FilePointer()` and updated `_WriteGraphToG6FilePath()` and `_WriteGraphToG6String()` helper functions to call the new `_WriteGraphToG6StrOrFile()` helper function.
* The new `beginG6WriteIterationToG6StrOrFile()` function validates the `outputContainer` and assigns it to the `pG6WriteIterator ` before calling `_beginG6WriteIteration()`
* `endG6WriteIteration()` now no longer "knows" whether the `g6Output` `strOrFileP` holds a string or a file, so remove the explicit call to `sf_closeFile()` and allow `sf_Free()` to clean that up.
    * __**N.B.**__ If the `g6Output` contains a string, it's expected that you must have `sf_TakeTheString()` before ending write iteration, otherwise `theStr` gets cleaned up.
* `_WriteGraphToG6FilePath()` now creates an `outputContainer` to handle opening the file corresponding to the `g6OutputFilename ` before calling `_WriteGraphToG6StrOrFile()` (which now owns that memory)
* `_WriteGraphToG6String()` now creates an `outputContainer` to pass into  `_WriteGraphToG6StrOrFile()` (owns that memory) alongside the pointer-pointer `g6OutputStr`
* The new `_WriteGraphToG6StrOrFile()` helper function accepts the `outputContainer`, ensures it's valid, and then allocates the `G6WriteIterator` before calling `beginG6WriteIterationToG6StrOrFile()` and proceeding with the remaining flow (i.e. `writeGraphUsingG6WriteIterator()`, __**taking the string from the `outputContainer` if `outputStr` is not `NULL`**__, `endG6WriteIteration()`, and `freeG6WriteIterator()`)

</details>

<details>
<summary>c/planarityApp/planarityRandomGraphs.c</summary>

This refactor was primarily due to the fact that `beginG6WriteIterationToG6FilePath()` was deprecated. So then the two branches that allow us to write to `.g6` had to be updated:
* If `RandomGraphs()` is called with non-NULL `outfileName`, it must have been called by `callRandomGraphs()` where the user specified the optional `O` parameter via the command-line. The `outputContainer` is initialized with this `outfileName` to handle opening the file object for this path for the command-line-driven branch.
* If `outfileName` is `NULL`, then we check to see if `RandomGraph()` was called via the menu-driven system after choosing to `Reconfigure()` with `r` ("Randomly generate graphs"), `y` ("Do you want original graphs in directory 'random'?"), and `g` ("Do you want to output generated graphs to ... G6 (all)?"). If so, then we construct `theFileName` based on the order of the graphs and the number to be generated and initialize the `outputContainer` with `theFileName`.
 
Afterwards, if the `G6WriteIterator` and `outputContainer` have been allocated, then we `beginG6WriteIterationToG6StrOrFile()` and will `writeGraphUsingG6WriteIterator()` during the random graph generation loop.

</details>

<details>
<summary>c/planarityApp/planarityTestAllGraphs.c</summary>

* In `TestAllGraphs()`, we create the `inputContainer` from the `infileName` and allow the `strOrFile` machinery to handle opening the file and creating the file pointer. Then, this is passed to `testAllGraphs()`, which sets up `g6` read iteration from this `strOrFileP` using `beginG6ReadIterationFromG6StrOrFile()`. This means that one may pipe `.g6`-encoded graphs in from `stdin` to test:
    ```
    % ./Release/planarity -t -p stdin n8.m15.g6.stdin_test.t.p.out.txt < ./TestSupport/results/graph_generation_orchestrator/8/n8.m15.g6 
    Start testing all graphs in "stdin".

    Done testing all graphs (0.006 seconds).

    % cat n8.m15.g6.stdin_test.t.p.out.txt
    FILENAME="stdin" DURATION="0.006"
    -p 1557 631 926 SUCCESS
    ```
* In `outputTestAllGraphsResults()`, we create the `testOutput` `strOrFileP` to simplify handling whether we're outputting to a file with path `outfileName` or to `outputStr`. Once the `headerStr` and `resultsStr` are written to whatever output via `sf_fputs()`, only if `outputStr` was non-NULL do we try to `sf_takeTheString()` from the `testOutput` and assign that pointer to the memory to which `outputStr` points. This means that one may still output the results of `TestAllGraphs()` to file or to string. 

</details>

### Changing the type of `extraData`

@john-boyer-phd explains the original motivation behind `extraData` having `void *` type and the choice to move to using `char *` in [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/140#discussion_r1953545412):
> When first designed, the thought was to let people do any binary thing they needed to do in extensions, but it eventually became apparent that extensions would only work together if we treated the extra data as a string, so all the code ended up relying on string-ness, so best to clean this up and make it be a string instead of a binary blob.

* `c/graphLib/extensionSystem/graphFunctionTable.h` - updated `graphFunctionTable` struct's function pointers' signatures for  `fpReadPostprocess()` and `fpWritePostprocess()` to correspond to `extraData` being a null-terminated `char *` (i.e. `extraDataLength` is no longer required)
* `c/graphLib/graphUtils.c` - updated default graph algorithm extension functions `_ReadPostprocess()` and `_WritePostprocess()` function signatures to correspond to `extraData` being a null-terminated `char *`
* `c/graphLib/planarityRelated/graphDrawPlanar_Extensions.c`
    * updated declarations and definitions of function overrides for `_ReadPostprocess()` (i.e. `_DrawPlanar_ReadPostprocess()`) and `_WritePostprocess()` (i.e. `_DrawPlanar_WritePostprocess()`) to correspond to `extraData` being a null-terminated `char *`
    * In `_DrawPlanar_ReadPostprocess()`, we can now take `strlen(extraData)`, since it must be a null-terminated `char *`, and no longer need to rely on supplementary variable `extraDataSize`. No longer need to cast `extraData` as a `char *` to advance the pointer beyond the textual `<DrawPlanar>` start tag and re-cast to `void *` (`extraData`'s original type), since we are guaranteed that `extraData` is a `char *`
    * In `_DrawPlanar_WritePostprocess()`, the `extraData` is now `calloc`'ed instead of `malloc`'ed to guarantee the memory allocated is zeroed out (since `\0` is a control character with value zero, this means any string we construct is guaranteed to be null-terminated)

### Resolving MSVC compiler warnings
* `c/planarityApp/planarityMenu.c` and `c/planarityApp/planarityUtils.c` - since `tolower()` returns an `int` rather than a `char`, explicitly cast the result to `char`
* `c/planarityApp/planarityRandomGraphs.c` - in `RandomGraphs()`, explicitly assign `outputContainer` `NULL` when declared
* `c/graphLib/io/g6-read-iterator.c` - Explicitly cast result of `strlen()` to `int` (from `size_t`) in comparisons, and cast `firstChar` and `graphChar` to `char` when we `sf_ungetc()`
* `c/graphLib/io/g6-write-iterator.c` - explicitly cast conversion of `graphOrder` to an `ASCII` character


c/graphLib/io/strOrFile.c
```diff
@@ -4,148 +4,536 @@ All rights reserved.
...
-            theStrOrFile->theStr = theStr;
-            theStrOrFile->theStrPos = 0;
+            strBufP strBufToAssign = sb_New(0);
+            if (strBufToAssign == NULL)
+            {
+                sf_Free(&theStrOrFile);
+                return NULL;
+            }
+
+            if (theStr != NULL && sb_ConcatString(strBufToAssign, theStr) != OK)
```
wbkboyer, line 117: @john-boyer-phd Should I add additional validation before this point to error out when the `strOrFile` is an `OUTPUT_CONTAINER` (i.e. `ioMode` is `WRITETEXT`) and a non-`NULL` `theStr` is passed in, or do you think appending some initial output content to the `theStr` is a valid use-case (and therefore I *shouldn't* error out)?  😨

c/graphLib/io/strOrFile.c
```diff
@@ -4,148 +4,536 @@ All rights reserved.
...
+                containerType = INPUT_CONTAINER;
+            else if (strncmp(ioMode, WRITETEXT, strlen(WRITETEXT)) == 0)
+                containerType = OUTPUT_CONTAINER;
+
+            theStrOrFile->theStr = strBufToAssign;
+        }
+
+        theStrOrFile->containerType = containerType;
+
+        theStrOrFile->ungetBuf = sp_New(MAXLINE);
```
wbkboyer, line 134: @john-boyer-phd Should I only allocate the `stackP` and assign to `ungetBuf` if the `containerType` is `INPUT_CONTAINER`?

I already added checks to error out otherwise in the functions that interact with the `ungetBuf`, i.e. `sf_getc()`, `sf_ungetc()`, `sf_ungets()`, and `sf_fgets()`.

c/graphLib/io/strOrFile.c
wbkboyer: @john-boyer-phd agreed that we should error-out if `theStr` is not `NULL` when we're initializing an `OUTPUT_CONTAINER`; the expected pattern is that one should create a new `OUTPUT_CONTAINER` using `sf_New(NULL, NULL, WRITETEXT)`, write to it using the relevant `sf_[a-zA-Z]+()` functions, and then `sf_takeTheStr()` before calling `sf_Free()`.

c/graphLib/io/strOrFile.c
wbkboyer: @john-boyer-phd agreed that we should only allocate the `ungetBuf` for `INPUT_CONTAINER`s (especially since any operations that interact with the `ungetBuf` will throw an error if `containerType` is anything other than `INPUT_CONTAINER`).

john-boyer-phd (changes requested):
Most comments are just going ga-ga over how great this change set is going to be. A few minor tweaky tweakies mixed in.

c/samples/nauty_example.LEDA
```diff
@@ -0,0 +1,17 @@
+LEDA.GRAPH
```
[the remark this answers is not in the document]
wbkboyer: Right on! Since I've included this hand-translation in the `c/samples/`, should I add `-test`s for:
* Transforming from LEDA to at least one other format
* `SpecificGraph()` from LEDA

c/graphLib/io/strOrFile.c
```diff
@@ -4,148 +4,542 @@ All rights reserved.
...
+    char intCandidateStr[MAXCHARSFOR32BITINT];
+    memset(intCandidateStr, '\0', MAXCHARSFOR32BITINT * sizeof(char));
+
+    int intCandidate = 0;
+    char currChar, nextChar = '\0';
+    bool startedReadingInt = FALSE;
+    int intCandidateIndex = 0;
+    while ((currChar = sf_getc(theStrOrFile)) != EOF)
+    {
+        if (intCandidateIndex > (MAXCHARSFOR32BITINT - 2))
```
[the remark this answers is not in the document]
wbkboyer: Since I use `%d` in the `sscanf()` in `sf_ReadInteger()`, I believe it will try to still read the contents of `intCandidateStr` as a signed integer and will produce an incorrect result; I was thinking that the updated comment in `apiutils.h` for `MAXCHARSFOR32BITINT` should be:

```
// N.B. Every time you're trying to read a 32-bit int from a string,
// you should only need to read this many characters: an optional '-',
// followed by 10 digits (max signed int value is 2,147,483,647). One
// must always allocate an additional byte for the null-terminator!
```

Does that work? 🎉

john-boyer-phd (changes requested):
Great bug fixes on stdout! Just a few tweaks on the tweaks.

c/graphLib/io/g6-read-iterator.c
```diff
@@ -704,14 +706,17 @@ int _ReadGraphFromG6StrOrFile(graphP pGraphToRead, strOrFileP g6InputContainer)
         return NOTOK;
     }
 
-    if (readGraphUsingG6ReadIterator(pG6ReadIterator) != OK)
+    exitCode = readGraphUsingG6ReadIterator(pG6ReadIterator);
+    if (exitCode != OK)
         ErrorMessage("Unable to read graph from .g6 read iterator.\n");
 
-    if (endG6ReadIteration(pG6ReadIterator) != OK)
+    exitCode = endG6ReadIteration(pG6ReadIterator);
```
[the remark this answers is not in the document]
wbkboyer: I decided to follow the same pattern that's in `g6-write-iterator.c`'s `_WriteGraphToG6StrOrFile()`, i.e. using intermediate variables to store the exit codes from ending and freeing the `G6ReadIterator`. Is that an ok solution, or should I change both to use your method of returning `NOTOK` on nonzero `errorCount`? 🎉

c/graphLib/io/graphIO.c
```diff
@@ -632,24 +620,21 @@ int _WriteAdjList(graphP theGraph, FILE *Outfile, strBufP outBuf)
  returns OK for success, NOTOK for failure
  ********************************************************************/
 
-int _WriteAdjMatrix(graphP theGraph, FILE *Outfile, strBufP outBuf)
+int _WriteAdjMatrix(graphP theGraph, strOrFileP outputContainer)
 {
     int v, e, K;
     char *Row = NULL;
     char numberStr[128];
```
wbkboyer, line 627: @john-boyer-phd Same issue as with `_WriteAdjList()`: should this be set to `MAXCHARSFOR32BITINT`?

c/graphLib/io/g6-read-iterator.c
```diff
@@ -710,13 +710,19 @@ int _ReadGraphFromG6StrOrFile(graphP pGraphToRead, strOrFileP g6InputContainer)
...
+    int endG6ReadIterationCode = endG6ReadIteration(pG6ReadIterator);
+    if (endG6ReadIterationCode != OK)
+    {
         ErrorMessage("Unable to end G6ReadIterator.\n");
+        exitCode = endG6ReadIterationCode;
+    }
 
-    exitCode = freeG6ReadIterator(&pG6ReadIterator);
+    int freeG6ReadIteratorCode = freeG6ReadIterator(&pG6ReadIterator);
     if (exitCode != OK)
```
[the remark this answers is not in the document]
wbkboyer: OH GOSH, YEP! Thank you for catching that!!!

c/samples/nauty_example.LEDA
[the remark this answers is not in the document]
wbkboyer: After discussion with @john-boyer-phd , we've decided to postpone adding LEDA transformation test until such time as we have implemented the LEDA writer.

c/graphLib/lowLevelUtils/apiutils.c
```diff
@@ -38,3 +41,62 @@ void ErrorMessage(char *message)
         fflush(stderr);
     }
 }
+
+int GetNumCharsToReprInt(int theNum, int *numCharsRequired)
```
john-boyer-phd, line 45: Good job noticing the architecture point that this needed to move to the graphLib level from the planarity wrapper app level.

c/graphLib/lowLevelUtils/apiutils.c
```diff
@@ -38,3 +41,62 @@ void ErrorMessage(char *message)
...
+        theNum /= 10;
+        charCount++;
+    }
+
+    (*numCharsRequired) = charCount;
+
+    return OK;
+}
+
+int ChopNumDigitsFromInt(int *intToTruncate, int numDigitsToChop)
```
john-boyer-phd, line 75: I think you should delete this because you can avoid using malloc() and free() by instead using integer division by 10 at the call point (which only needs to shave the LSB).

c/graphLib/lowLevelUtils/apiutils.h
```diff
@@ -32,6 +32,9 @@ extern "C"
     extern void Message(char *message);
     extern void ErrorMessage(char *message);
 
+    int GetNumCharsToReprInt(int theNum, int *numCharsRequired);
+    int ChopNumDigitsFromInt(int *intToTruncate, int numDigitsToChop);
```
john-boyer-phd, line 36: This line should also be deleted since the function it declares will be deleted.

c/graphLib/io/strOrFile.c
```diff
@@ -346,22 +342,64 @@ int sf_ReadInteger(int *intToRead, strOrFileP theStrOrFile)
...
+        {
+            if (sscanf(intCandidateStr, "%d", &intCandidate) != 1)
+                exitCode = NOTOK;
+            else
+            {
+                // N.B. The only way we could be here is if `i` is less than the max possible
+                // length of the string representation of a signed 32-bit integer (10 for
+                // positive and 11 for negative)
+                int underflowThreshold = INT32_MIN, overflowThreshold = INT32_MAX;
+                if (ChopNumDigitsFromInt((&underflowThreshold), 1) != OK || ChopNumDigitsFromInt((&overflowThreshold), 1) != OK)
```
john-boyer-phd, line 364: The ChopNumDigitsFromInt(..., 1) calls should be replaced with integer division by 10.

john-boyer-phd (changes requested):
Great set of changes, nearly there. Just a tweak to how ReadInteger() detects the edge cases (and retest).

john-boyer-phd (approved):
Looks great on all fronts of this substantial and much-needed set of changes.

PR #141, opened 2025-02-26
Issue #24 - Use compiler preprocessor option to Signal 1-based versus 0-based Array Indexing

### LINKED ISSUE #24: Use Dedicated Macro to Signal 1-based versus 0-based Array Indexing
The original implementation stored vertices and edges using 0-based array indexing. This resulted in a definition of -1 for the `NIL` index (indicating no vertex or index). A significant performance increase was achieved by changing `NIL` to 0 because then every single test for `NIL` became an equality comparison with 0.

Hence, the value of the `NIL` macro is used in `#ifdef` tests throughout the project to decide whether to compile 1-based or 0-based code. 

The goal in this issue is to define a macro, such as `#define ARRAY_BASE_INDEX 1`, or `0`, and then let that setting drive the decisions about what code to compile, including what value is chosen for `NIL`.

wbkboyer:
Resolves #24

Updated [1. Project Overview - Optimization of 1-Based Array Indexing](https://github.com/graph-algorithms/edge-addition-planarity-suite/wiki/1.-Project-Overview#optimization-of-1-based-array-indexing) to reflect the change to how we manage 0-based vs. 1-based array indexing using the `-DUSE_0BASEDARRAYS`/`/D USE_0BASEDARRAYS` compiler option controlling the preprocessor.

### Necessary bugfix
* `c/graphLib/io/strOrFile.c` - need to `#include <stdint.h>` for the definitions for `INT32_MIN`/`MAX`; this was not discovered during #140, although tests were performed on a MacBook pro with `clang`. I found a [StackOverflow response](https://stackoverflow.com/a/19440917) suggesting that perhaps `clang`'s `stdlib.h` includes `stdint.h` for internal purposes, which *might* have been why compilation/running didn't fail on MacOS but *did* on Windows without this include, but I couldn't manually divine the appropriate chain of `#include`s to have given `stdint.h` for free while surfing through the `clang` `stdlib.h` and `string.h` headers :(

### Changes pertaining to Issue #24 
* `c/graphLib/`
    * `graphStructures.h` - instead of conditioning on the value of `NIL`, now check whether `#ifdef USE_FASTER_1BASEDARRAYS`, and no longer need the `#else #error`.
    * `graphUtils.c` - instead of conditioning on the value of `NIL`, now check whether `#ifdef USE_FASTER_1BASEDARRAYS` for how to `_InitVertices()`; the legacy API-based approach to initialize vertices has been commented-out for future reference.
    * `homeomorphSearch/`
        * `graphK33Search_Extensions.c` - Previously, when `NIL` was either 0 or 1, we always performed the same `memset()`s, so the `#if` guard has been removed. The legacy API-based approach to initializing the structures required for K<sub>3, 3</sub> search has been commented-out for future reference.
        * `graphK4Search_Extensions.c` -  Previously, when `NIL` was either 0 or 1, we always performed the same `memset()`s, so the `#if` guard has been removed. The legacy API-based approach to initializing the structures required for K<sub>4</sub> search has been commented-out for future reference.
    * `lowLevelUtils/appconst.h` - definition of `NIL` and `NIL_CHAR` now depend on whether or not `USE_FASTER_1BASEDARRAYS` is defined. If the user compiles with the `-DUSE_0BASEDARRAYS`/`/DUSE_0BASEDARRAYS` compiler option to control the preprocessor, then we `#undef USE_FASTER_1BASEDARRAYS`, which means that `NIL` is set to `-1` and `NIL_CHAR` is set to `0xFF`. Otherwise, if this compiler option is not set, then the default 1-based values for `NIL` (i.e. `0`) and `NIL_CHAR` (i.e. `0x00`) are used.
    * `planarityRelated/graphDrawPlanar_Extensions.c` - now condition on whether `#ifdef USE_FASTER_1BASEDARRAYS` rather than `#ifdef NIL == 0` when we `_DrawPlanar_InitStructures()`
 * `c/planarityApp/planarityCommandLine.c` - Updated `Message()` contents to better indicate these are 1-based array index tests
* `devEnvSetupAndDefaults/.vscode/tasks.json` - added (commented out) compiler preprocessor option for defining `USE_0BASEDARRAYS` to all build configurations.

### DIFF
--- a/c/graphLib/graphStructures.h
+++ b/c/graphLib/graphStructures.h
@@ -64,14 +64,14 @@ extern "C"
 
     typedef edgeRec *edgeRecP;
 
-#ifdef ZEROBASED
-#define gp_IsArc(e) ((e) != NIL)
-#define gp_IsNotArc(e) ((e) == NIL)
-#define gp_GetFirstEdge(theGraph) (0)
-#else
+#ifdef USE_FASTER_1BASEDARRAYS
 #define gp_IsArc(e) (e)
 #define gp_IsNotArc(e) (!(e))
 #define gp_GetFirstEdge(theGraph) (2)
+#else
+#define gp_IsArc(e) ((e) != NIL)
+#define gp_IsNotArc(e) ((e) == NIL)
+#define gp_GetFirstEdge(theGraph) (0)
 #endif
 
 #define gp_EdgeInUse(theGraph, e) (gp_IsVertex(gp_GetNeighbor(theGraph, e)))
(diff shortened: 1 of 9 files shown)


john-boyer-phd (changes requested):
Now that I see it as diffs, driving this off of "ZEROBASED" is troublematic because it reverses the code around to make the programmer first read the old slower code (which was the competitive code at the Graph Drawing Conference, but still the current code is faster and so preferred). We should name this and arrange the code so that the reader has no misunderstanding about what option they should preferably use. So, let's please make some changes based on the following steps:

1. In `appconst.h`, do a `#define USE_FASTER_1BASEDARRAYS`
2. Immediately below that `#fdef USE_0BASEDARRAYS`, `#undefine USE_FASTER_1BASEDARRAYS`, `#endif`
3. Change all the `#else` clauses that contain the 1-based code to be first and to be brought into compilation by `#ifdef USE_FASTER_1BASEDARRAYS`
4. Put all the zero-based code in second place, within the `#else` clauses
5. Change tasks.json from `ZEROBASED` to `USE_0BASEDARRAYS`

john-boyer-phd (approved):
All good.

Issue #147, opened 2025-04-08
Fix compile warnings


james-d-mitchell:
Firstly thanks for the excellent library. We are wrapping it for the GAP package digraphs at:

https://github.com/digraphs/Digraphs/

(we've been doing this for a number of years). I've been trying to update the package to use version 4.0.0.0, and one of the issues I've encountered is that there are a very large number of compiler warnings issues:

[compile-warning.txt](https://github.com/user-attachments/files/19650545/compile-warning.txt)

It seems to me that there are a number of issues, a number of features of c99 (namely, missing prototypes, and "mixing declarations and code is incompatible with standards before C99") and c11 (redeclaring typedefs) are used without this being specified in the build system. A good starting point for fixing this would be to decide upon which c standard you are using, and then to fix the remaining warnings. I have started fixing these in the Digraphs package, and would be happy to complete this and submit a PR here if it is likely to be acceptable to you?

I'll be happy to hear your thoughts, positive or otherwise...

john-boyer-phd:
Hi James,

The high order bits are that we are very interested in addressing the warnings you are seeing and, yes, one or more PRs from you would be most welcome and tremendously helpful. Thank you for your kind offer. We’ll also cut a 4.0.0.1 release when we get these warnings resolved, in the hope that it will make your life easier going forward.

For 4.0.0.0, we had to address a particular type of warning that Michael Orlitzky made us aware of (#118), so we decided to kill off all warnings produced by `-Wall` across Debian Linux gcc (v12.2.0), Mac clang (Homebrew 19.1.7), and Windows with the gcc from MinGW 64 (13.2.0). We also have been doing our best to still be able to compile with MSVC cl, which seems to be an embrace-and-extend of C89 (the 8 is not a typo). 

Are you generating these warnings using `-Wextra` or by something else? Could you share any other platform and compiler version info?

Scanning through the compiler warnings file that you attached, I see that a great number of the warnings can be eliminated by adding a line feed at the end of `apiutils.h` and, more importantly, by addressing the many calls to `ErrorMessage()` corresponding to all the additional error checking that went into 4.0, especially the new G6 file format support.

How would you feel about doing a first PR just to kill those warnings and then creating a smaller file showing what's left?

[...]

PR #148, opened 2025-04-09
Add `--enable-compile-warnings` configure flag

james-d-mitchell:
This PR adds the flag `--enable-compile-warnings` which enables a slew of compiler warnings. These are the compiler flags used by the Digraphs package for GAP as reported in #147. 

I thought this would be a good place to start with resolving #147 so that it is possible to recreate the warnings here.

### DIFF
--- a/m4/ax_require_defined.m4
+++ b/m4/ax_require_defined.m4
@@ -1,0 +1,37 @@
+# ===========================================================================
+#    https://www.gnu.org/software/autoconf-archive/ax_require_defined.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_REQUIRE_DEFINED(MACRO)
+#
+# DESCRIPTION
+#
+#   AX_REQUIRE_DEFINED is a simple helper for making sure other macros have
+#   been defined and thus are available for use.  This avoids random issues
+#   where a macro isn't expanded.  Instead the configure script emits a
+#   non-fatal:
+#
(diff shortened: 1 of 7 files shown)


james-d-mitchell:
It seems that I forgot to add the m4 file `ax_require_defined.m4` in the original pull request, which meant that the changes in this PR originally would have led to unhelpful errors from autoconf. The PR should now be complete (missing file added), and everything required by the autotools should now be included. Sorry about that!

m4/ax_require_defined.m4
```diff
@@ -0,0 +1,37 @@
+# ===========================================================================
+#    https://www.gnu.org/software/autoconf-archive/ax_require_defined.html
```
wbkboyer, line 2: Very cool! Thank you very much for adding this file; I’ll test this change on Monday on Windows (MSYS2 UCRT with MinGW-w64 gcc and autotools) and Debian Linux (gcc and autotools) to make sure we’re still able to build the distribution without installing autoconf-archive to the build environment (as I mentioned on #149 )!

wbkboyer (approved):
Right on, with the addition of [`m4/ax_require_defined.m4`](https://www.gnu.org/software/autoconf-archive/ax_require_defined.html), I'm able to build on Debian Linux and Windows (using MSYS2 UCRT shell) without installing `autoconf-archive`, so I think this is good to merge! Thank you again, @james-d-mitchell !

PR #149, opened 2025-04-09
Fix compile warnings 1

james-d-mitchell:
This PR builds on #148, by fixing the compile warnings:

* `-Wnewline-eof errors` this was pretty straightforward.
* `-Wincompatible-pointer-types-discards-qualifiers` this required pretty extensive changes to add `const` to string literals, variables representing them, and to function parameters. I'd like to check that I did this correctly again, but thought I'd post this for visibility first.


james-d-mitchell:
I looked again, and I'm happy with this now.

wbkboyer (changes requested):
Thank you so much for your excellent contributions, @james-d-mitchell ! @john-boyer-phd and I have been eagerly discussing your PRs, and I've been doing some experimentation behind the scenes. What I discovered is that in order to accept PR #148, we'd have to require that the `autoconf-archive` package be installed in the build environment (otherwise running the `./configure` script generated from `configure.ac` after running `autoreconf -fi` blows up because `AX_COMPILER_FLAGS_CFLAGS` is unrecognized); this requirement might not be available in the build environments of other downstream consumers, so if you're amenable, please can you remove the PR #148 code from this PR, as we're able to test all of your PRs that are along these lines by temporarily adding the appropriate warning flags (e.g. `-Wincompatible-pointer-types-discards-qualifiers`) to the VSCode `tasks.json` build configuration file.

james-d-mitchell:
Thanks for your comments @wbkboyer! I'm not sure that 

> we'd have to require that the autoconf-archive package be installed in the build environment (otherwise running the ./configure 

is correct, but it seems I forgot to include one of the required m4 files in the commit [aa3e789](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/149/commits/aa3e789065c8a6b031a499298348816a15603245) which would have led to the warning you've seen. 

In any case you are right I shouldn't mix up the two pull requests, I'll remove the first commit!

john-boyer-phd (approved):
These look great, James, thank you very much for contributing them. I also asked Wanda to run a bunch of tests to make sure I didn't accidentally ask for the wrong thing, and all the tests were good to go, too. (I had worried about accidentally changing function signatures, even though the const additions shouldn't do that; sure enough, it didn't (across platforms)).

wbkboyer (approved):
Thank you again, @james-d-mitchell ! I've verified that the 4.0 `planarity` wrapper executable is able to call the functions in the `.libs` with these updated function signatures 🎉

PR #150, opened 2025-04-11
Specify `pkgconfigdir` in `Makefile.am`

Joseph-Edwards:
When compiling on certain architectures, the following error message is produced:
```
Makefile.am:93: error: 'pkgconfig_DATA' is used but 'pkgconfigdir' is undefined
```

This pull request specifies the `pkgconfigdir` to address that error.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -90,4 +90,5 @@ TESTS = test-samples.sh
 docdir = @docdir@
 dist_doc_DATA = README.md LICENSE.TXT
 
-pkgconfig_DATA = libplanarity.pc
+pkgconfigdir	= $(libdir)/pkgconfig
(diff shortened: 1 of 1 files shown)


Joseph-Edwards:
This error is the same error causing the tests to fail in https://github.com/conda-forge/planarity-feedstock/pull/22. Hopefully, by merging this pull request, the checks in the feedstock will pass.

wbkboyer:
This looks great! Thank you for this fix, @Joseph-Edwards 🎉  I confirmed that adding this line to `Makefile.am` doesn't change where the `libplanarity.pc` shows up (i.e. `/usr/local/lib/pkgconfig/`) when one `sudo make install` the distribution build on Debian Linux.

Issue #152, opened 2025-04-11
Enable CI tests


wbkboyer:
In #151 , @Joseph-Edwards introduced some GitHub actions to test for compiler warnings, to ensure the tests run in various environments, and to check for memory leaks using `valgrind`. He advises in [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/151#issuecomment-2796388004):

> At the moment, workflows are not enabled for this repository, so no checks are run. To enable the workflows, see the guide [Disabling and enabling a workflow](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow). To see what a pull request would look like with these workflows enabled, see https://github.com/Joseph-Edwards/edge-addition-planarity-suite/pull/1 on my fork.

As a follow-up, 

1. Ensure that these workflows are enabled, and that merging PRs are blocked until these tests pass
2. Add environments (e.g. Debian Linux, Windows with MinGW-w64) to check the code builds and tests pass
3. Investigate whether we should add additional actions to improve code quality

PR #151, opened 2025-04-11
Add some continuous integration checks

Joseph-Edwards:
This pull request adds some continuous integration checks that will be triggered when new pull requests are made. In particular, the tests:
* check for compiler warnings;
* check that the tests pass using a variety of different clang and gcc versions;
* check that the tests pass on the systems Alpine, Cygwin, macOS, Ubuntu with aarch64, and Ubuntu with arm;
* run an address sanitizer, undefined behaviour sanitizer and valgrind; and
* checks for spelling mistakes.

PR #157, opened 2025-04-16
Add workflow jobs for Windows and Debian 

Joseph-Edwards:
In response to @john-boyer-phd's comment in #151, and to #152, this pull request adds workflow jobs that checks the tests pass on various windows systems using `MSYS2`, and on Debian. It also gets rid of the the dependency on `gap-actions/setup-cygwin`.


john-boyer-phd (approved):
Thank you for adding these, Joe!

PR #159, opened 2025-06-04
Issue #147 - Update VSCode `tasks.json` with explicit list of platform-dependent compiler flags (i.e. from `--enable-compile-warnings`)

wbkboyer:
Contributes to process of addressing #147 

## Updated
* `devEnvSetupAndDefaults/.vscode/tasks.json` - Updated to include explicit list of compiler flags used with `clang` and `gcc` when one uses the `autotools` build pipeline to make the distribution with `--enable-compile-warnings`
    * __**N.B.**__ The list of compiler flags supplied to `clang` is a proper subset of those supplied to `gcc`, and both of these lists include (the same) flags not found in `m4/ax_compiler_flags_cflags.m4`. Additionally, although both the `clang` and `gcc` compiler flag lists include `-Wnull-dereference`, it appears there's an [issue with `llvm`](https://github.com/llvm/llvm-project/issues/56793) that causes some of the null dereference warnings to be missed. See [this Google sheet](https://docs.google.com/spreadsheets/d/1aejKPcN84pxDwzjQ1u0qWNRL2c9Zr7kZzD8Q5y4QCbs/edit?usp=sharing) for the full comparison.


john-boyer-phd (approved):
Thanks, this is what I need to let me incrementally work on eliminating the more pedantic warnings.

PR #164, opened 2025-07-23
Issue #160 - restore `beginG6(Read|Write)Iteration(From|To)G6(String|FilePath)` helper functions, and update `planarityApp` layer to leverage these functions

### LINKED ISSUE #160: Restore `G6ReadIterator` and `G6WriteIterator` endpoints that accept `char *` for `.g6` string or as path to `.g6` file
## Problem background

To satisfy #57 , #140 introduced the pattern that `G6ReadIterator` And `G6WriteIterator` functions should accept a `strOrFileP` to manage input/output, allowing the caller to allocate and initialize the `strOrFileP` before beginning iteration using `beginG6ReadIterationFromG6StrOrFile()` when reading or `beginG6WriteIterationToG6StrOrFile()` when writing. In doing so, the following functions were removed:
* `beginG6ReadIterationFromG6String()`
* `beginG6ReadIterationFromG6FilePointer()`
* `beginG6ReadIterationFromG6FilePath()`
* `beginG6WriteIterationToG6FilePath()`
* `beginG6WriteIterationToG6FilePointer()`
* `beginG6WriteIterationToG6String()`

Because these functions were removed, when exposing the `G6ReadIterator` and `G6WriteIterator` to Python applications via a Cython wrapper (see #137), one must add declarations for `strOrFileP` to the `.pxd` definition file and then allocate and initialize the `strOrFileP` before calling `beginG6ReadIterationFromG6StrOrFile()`, which adds unnecessary complexity to the wrapper implementation.

## Proposed solution

Restore the following functions:

* `beginG6ReadIterationFromG6String()`
* `beginG6ReadIterationFromG6FilePath()`
* `beginG6WriteIterationToG6String()`
* `beginG6WriteIterationToG6FilePath()`

These changes would require the minor version number of the package to be increased once a release is cut.

This work should be done after that of #161

wbkboyer:
Follow up to #57 and #140

Resolves #160

## Updated

Bumping version numbers for planarity app and libraries; since this work will add interfaces, the following changes were made:

* `configure.ac` - `LT_CURRENT` and `LT_AGE` were incremented, and the `AC_INIT` was updated to match values in `c/graphLib.h`
* `c/graphLib.h` -  `GP_PROJECTVERSION_MAINT` was incremented (new interface), and the `GP_LIBPLANARITYVERSION_CURRENT` and `GP_LIBPLANARITYVERSION_AGE` were updated to match `configure.ac`

* Restoring `beginG6(Read|Write)Iteration(From|To)G6(String|FilePath)` functions in `g6-(read|write)-iterator.c`\`.h`, which create the appropriate `strOrFileP` and call `beginG6(Read|Write)Iteration(From|To)G6StrOrFile`.
    * The only notable difference is that `beginG6WriteIterationToString` now only has a single parameter, the `G6WriteIteratorP`; this is because we never want to pass a pointer-pointer to memory to use for the strOrFileP's internal theStr when `ioMode` is `WRITETEXT` - it should be allocated and owned by the `strOrFileP` entirely until you `sf_takeTheStr()`.
* `c/planarityApp/planarityTestAllGraphs.c` - updated `testAllGraphs()` so that it calls `beginG6ReadIterationFromG6FilePath()`
    * ensured that the tables produced don't change:
        ```
        python3.12 planarity_testAllGraphs_orchestrator.py -p ../../Release/planarity.exe -n 5

        python3.12 test_table_generator.py -n 5 -c (p|d|o|2|3|4)
        ```
* `c/planarityApp/planarityRandomGraphs.c` - updated `RandomGraphs()` so that we either `beginG6WriteIterationToG6FilePath()` for `outfileName` (if we get to `RandomGraphs()` via command-line, e.g. `planarity -r -p 100 1000 n100.k1000.g6`) or `theFileName` (if we get to `RandomGraphs()` via the menu-driven system after `Reconfigure()` to output all generated random graphs to `.g6`)

### DIFF
--- a/c/graphLib/io/g6-write-iterator.c
+++ b/c/graphLib/io/g6-write-iterator.c
@@ -113,16 +113,16 @@ int getGraphBuff(G6WriteIteratorP pG6WriteIterator, char **ppCurrGraphBuff)
 
 int beginG6WriteIterationToG6String(G6WriteIteratorP pG6WriteIterator)
 {
-    strOrFileP outputContainer = sf_New(NULL, NULL, WRITETEXT);
-
-    return beginG6ReadIterationFromG6StrOrFile(pG6WriteIterator, outputContainer);
+    return beginG6WriteIterationToG6StrOrFile(
+        pG6WriteIterator,
+        sf_New(NULL, NULL, WRITETEXT));
 }
 
 int beginG6WriteIterationToG6FilePath(G6WriteIteratorP pG6WriteIterator, char *outputFilename)
 {
-    strOrFileP outputContainer = sf_New(NULL, outputFilename, WRITETEXT);
-
-    return beginG6ReadIterationFromG6StrOrFile(pG6WriteIterator, outputContainer);
+    return beginG6WriteIterationToG6StrOrFile(
+        pG6WriteIterator,
+        sf_New(NULL, outputFilename, WRITETEXT));
 }
 
 int beginG6WriteIterationToG6StrOrFile(G6WriteIteratorP pG6WriteIterator, strOrFileP outputContainer)
@@ -508,7 +508,7 @@ int _WriteGraphToG6String(graphP pGraph, char **g6OutputStr)
 
     // N.B. If g6OutputStr is a pointer to a pointer to a block of memory that
     // has been allocated, i.e. if g6OutputStr != NULL && (*g6OutputStr) != NULL
-    // then an error will be emitted.
+    // then an error will be emitted by _WriteGraphToG6StrOrFile().
     // N.B. Once the graph is successfully written, the string is taken from
     // the G6WriteIterator's outputContainer and assigned to (*g6OutputStr)
     // before ending G6 write iteration
--- a/c/graphLib/io/g6-write-iterator.c
+++ b/c/graphLib/io/g6-write-iterator.c
@@ -111,6 +111,20 @@ int getGraphBuff(G6WriteIteratorP pG6WriteIterator, char **ppCurrGraphBuff)
     return OK;
 }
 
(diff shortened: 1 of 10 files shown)


c/graphLib/io/g6-write-iterator.c
```diff
@@ -492,6 +506,12 @@ int _WriteGraphToG6String(graphP pGraph, char **g6OutputStr)
         return NOTOK;
     }
 
+    // N.B. If g6OutputStr is a pointer to a pointer to a block of memory that
+    // has been allocated, i.e. if g6OutputStr != NULL && (*g6OutputStr) != NULL
+    // then an error will be emitted.
```
john-boyer-phd, line 511: An error will be emitted by what?

c/planarityApp/planarityRandomGraphs.c
```diff
@@ -67,6 +69,12 @@ int RandomGraphs(char command, int NumGraphs, int SizeOfGraphs, char *outfileNam
         }
     }
 
+    // FIXME: I don't want to refactor this to use
+    // beginG6WriteIterationToG6FilePath(), because there's two possible ways to
+    // get an output filename: either you called planarity command line and
+    // passed the optional outfileName, or you are using the planarity menu
+    // system after you Reconfigure() to output to .g6 (in which case,
+    // theFileName is constructed).
```
john-boyer-phd, line 77: It seems that beginG6WriteIterationToG6FilePath() would work in either case, since sf_New() does (line 82).

john-boyer-phd (changes requested):
Coming along nicely. Seems like RandomGraphs can be refactored too.

john-boyer-phd (approved):
Nice changes to the helper functions and especially to RandomGraphs().

PR #166, opened 2025-08-01
Issue #165 - Add sentinel variable to and macro "getter" to signal caller should not attempt to continue to read using the `G6ReadIterator`

### LINKED ISSUE #165: Create helper to indicate whether `.g6` read iteration has terminated
When using the `G6ReadIterator` to read graphs from a `.g6` input container, one must first allocate the `G6ReadIterator` and begin iteration (part of this process ensures that the input is valid) before making repeated calls to `readGraphUsingG6ReadIterator()` to consume the lines of the input. Therefore, if this flow has been followed, `sf_fgets()` returns `NULL` when there's no more content to return from the input. In this case, the ownership of the `graphP` is returned to the caller by setting the `pG6ReadIterator->currGraph` pointer to `NULL`. The only `planarityApp` workflow that currently uses the `G6ReadIterator` is in `planarityTestAllGraphs.c`, which contains a `while(true)` loop that repeatedly calls `readGraphUsingG6ReadIterator()` and only breaks when the `pG6ReadIterator->currGraph == NULL`. 

While undertaking #137 , I discovered that we need an explicit way to interrogate the C layer to determine that read iteration should terminate; although I could implement this test only at the Cython layer, it is desirous to expose a function (or macro) that explicitly performs this test which would then be used both in `planarityTestAllGraphs.c` as well as within Cython.

wbkboyer:
Resolves #165 

## Updated
* `c/graphLib/io/g6-(read|write)-iterator.c`/`.h` - Strengthened validation and error messaging of `_isG6ReadIteratorAllocated()` and `_isG6WriteIteratorAllocated()`.
* `c/graphLib/io/g6-read-iterator.c`/`.h` - Added macro `G6ReadIteratorContentsExhausted()` which will indicate that one should no longer try to read from the `G6ReadIterator` (sentinel variable `contentsExhausted` added to `G6ReadIterator` `struct` is set in `readGraphUsingG6ReadIterator()` when `sf_fgets()` returns `NULL`)
* `c/graphLib/io/g6-write-iterator.c` - Removed unnecessary call to `_isG6WriteIteratorAllocated()` in `writeGraphUsingG6WriteIterator()` (since `_encodeAdjMatAsG6()` immediately calls it), and removed some spaces in `_WriteGraphToG6StrOrFile()` to bring the exitCode tests up beneath the preceding calls.
* `c/planarityApp/planarityTestAllGraphs.c` - Updated `testAllGraphs()` so that it references `theGraph` rather than accessing `pG6ReadIterator->currGraph`, since we already have this external pointer. Also added `FIXME` statements regarding access of `struct` members, that we should have "getter" macros.

### DIFF
--- a/c/planarityApp/planarityTestAllGraphs.c
+++ b/c/planarityApp/planarityTestAllGraphs.c
@@ -143,13 +143,14 @@ int testAllGraphs(graphP theGraph, char command, char *infileName, testAllStatsP
         return Result;
     }
 
-    int graphOrder = pG6ReadIterator->graphOrder;
+    // FIXME: Need to implement "getter" macro to access attributes of graphP
+    int graphOrder = theGraph->N;
     // We have to set the maximum arc capacity (i.e. (N * (N - 1))) because some of the test files
     // can contain complete graphs, and the graph drawing, K_{3, 3} search, and K_4 search extensions
     // don't support expanding the arc capacity after being attached.
     if (strchr("d34", command) != NULL)
     {
-        Result = gp_EnsureArcCapacity(pG6ReadIterator->currGraph, (graphOrder * (graphOrder - 1)));
+        Result = gp_EnsureArcCapacity(theGraph, (graphOrder * (graphOrder - 1)));
         if (Result != OK)
         {
             ErrorMessage("Unable to maximize arc capacity of G6ReadIterator's graph struct.\n");
@@ -159,7 +160,7 @@ int testAllGraphs(graphP theGraph, char command, char *infileName, testAllStatsP
         }
     }
 
-    AttachAlgorithm(pG6ReadIterator->currGraph, command);
+    AttachAlgorithm(theGraph, command);
 
     copyOfOrigGraph = gp_New();
     if (copyOfOrigGraph == NULL)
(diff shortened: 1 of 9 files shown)


c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -143,13 +143,14 @@ int testAllGraphs(graphP theGraph, char command, char *infileName, testAllStatsP
         return Result;
     }
 
-    int graphOrder = pG6ReadIterator->graphOrder;
+    // FIXME: Need to implement "getter" macro to access attributes of graphP
```
john-boyer-phd, line 146: This is true, but best to remove this comment and file an issue instead because it needs to be done uniformly across the codebase.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -179,40 +180,28 @@ int testAllGraphs(graphP theGraph, char command, char *infileName, testAllStatsP
...
-    }
-
     while (true)
     {
         Result = readGraphUsingG6ReadIterator(pG6ReadIterator);
 
         if (Result != OK)
         {
             messageFormat = "Unable to read graph on line %d from .g6 read iterator.\n";
+            // FIXME: Need to implement "getter" macro to access attributes of G6ReadIterator
```
john-boyer-phd, line 190: It's better to remove this comment and instead add the getter to this PR because this issue/PR was inspired by the desire to not expose data members of the read iterator to the new test all graphs implementation you're making in Python, so best to make sure it's also stamped out here too.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -222,6 +211,7 @@ int testAllGraphs(graphP theGraph, char command, char *infileName, testAllStatsP
         else
         {
             messageFormat = "Error applying algorithm '%c' to graph on line %d.\n";
+            // FIXME: Need to implement "getter" macro to access attributes of G6ReadIterator
```
john-boyer-phd, line 214: This can use the same new getter.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -231,6 +221,7 @@ int testAllGraphs(graphP theGraph, char command, char *infileName, testAllStatsP
         gp_ReinitializeGraph(copyOfOrigGraph);
     }
 
+    // FIXME: Need to implement "getter" macro to access attributes of G6ReadIterator
```
john-boyer-phd, line 224: And this too.

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Created #168 🎉 But for the meanwhile, added `gp_getN()` and use it throughout the `G6(Read|Write)Iterator` functions

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Created #167 because this is out of scope for #165 (which is blocking the Planarity Cython wrapper 💪🏼 )

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: See #167

PR #186, opened 2025-10-25
Issue #30 - remediate command parsing

### LINKED ISSUE #30: EPIC: Remediate character-based menu and command-line control
Currently, the user specifies the algorithm to run using a single character in the menu-driven mode or a single character within the `C` command-line parameter. This must be changed to support passing around a string for the algorithm specifier. This will enable future algorithms to be added which may require more than one character to specify, such as enhanced versions of some of the current algorithm offerings.

wbkboyer:
Resolves #30 

## Updated

* `Makefile.am` - moved `c/planarityApp/planarity.h` from `nobase_pkginclude_HEADERS` into `planarity_SOURCES`, since it is a header file associated with the `planarity` wrapper executable (i.e. the `planarityApp` layer); this aligns with the [`automake` documentation - 9.2 Header files](https://www.gnu.org/software/automake/manual/html_node/Headers.html).
    * _**N.B.**_ This was a vestigial remnant from the original commit ([`85e1918`](https://github.com/graph-algorithms/edge-addition-planarity-suite/commit/85e1918a84b1bce4ea8a8378440c9a18997959b4)), where `planarity.h` was included in `headers_HEADERS`; this variable name was changed to adhere to the standard of `pkginclude_HEADERS` in [`3e7159b`](https://github.com/graph-algorithms/edge-addition-planarity-suite/commit/3e7159bcf265033191b2c100f59adf2e7dd51299).
    * I created a test application that proves that after installing `planarity` from the distribution tarball, it's not possible to call functions from the `planarityApp` layer after including the `planarity.h` header (i.e. demoing linker error)

        <details>

        `planaritytest.c`:

        ```
        #include <stdlib.h>
        #include <stdio.h>

        #include "planaritytest.h"

        int main(int argc, char *argv[])
        {
	        int exitCode = 0;
	        char Choice = 0, *infileName = NULL, *outfileName = NULL, *outfile2Name = NULL;

	        graphP theGraph = gp_New();
	        printf("theGraph address is %x\n", theGraph);

	        if (argc < 4)
		        return -1;
	
	        if (strlen(argv[1]) == 0)
	    	        return -1;

	        Choice = argv[1][0];

	        infileName = argv[2];
	        outfileName = argv[3];

	        if (argc == 5)
		        outfile2Name = argv[4];

	        printf("Attempting to call SpecificGraph() from external application:\n");

	        exitCode = SpecificGraph(Choice, infileName, outfileName, outfile2Name, NULL, NULL, NULL);
	        printf("Exit code of SpecificGraph is %d\n", exitCode);


	        return exitCode;
        }

        ```

        `planaritytest.h`:

        ```
        #ifndef PLANARITYTEST_H
        #define PLANARITYTEST_H

        #ifdef __cplusplus
        extern "C"
        {
        #endif

        #include <planarity/graphLib.h>

        #include <planarity/c/planarityApp/planarity.h>

        #ifdef __cplusplus
        }
        #endif

        #endif

        ```

        Using the following command to compile the test program:

        ```
        $ gcc -I/usr/local/include/planarity planaritytest.c -L/usr/local/lib -lplanarity -o planaritytest
        ```

        Linking failed:

        ```
        planarity planaritytest.c -L/usr/local/lib -lplanarity -o planaritytest
        /usr/bin/ld: /tmp/cclpIHvW.o: in function `main':
        planaritytest.c:(.text+0xdc): undefined reference to `SpecificGraph'
        collect2: error: ld returned 1 exit status
        ```

        </details>

* `configure.ac` and `c/graphLib/graphLib.h` - bumped `AC_INIT` maintenance value to match `GP_PROJECTVERSION_MAINT` so that the version is 4.0.2.0.
    * _**N.B.**_ I confirmed that the `planarity` wrapper executable as of the `HEAD` of `master` is able to call the `graphLib` functions whose function signatures changed for const-correctness
        <details>

        Confirmed that const-correctness changes to function signatures within `graphLib` (i.e. `sf_New()` and `beginG6ReadIterationFromG6FilePath()` now must take `char const * const` for `infileName` to accommodate change to `TestAllGraphs()`) do not cause compatibility issues:

        1. Made the distribution tarball from HEAD of master branch (after adding `printf()` in `planarityApp` and `graphLib` layers indicating that they corresponded to functions built from `master`)
        2. Made the distribution tarball from `HEAD` of Issue 30 branch (after adding `printf()` in `planarityApp` and `graphLib` layers indicating that they corresponded to functions built from Issue30 branch)
        3. Untarred the builds into two separate directories, then ran through process to `./configure && make` (so that we have the wrapper executable + libs for both the `HEAD` of `master` + `HEAD` of Issue 30 branch in their respective directories)
        4. Made backup copies of `.libs/planarity` (i.e. `.libs/planarity-master` and `.libs/planarity-Issue30` in their respective directories)
        5. Copied `.libs/planarity` from the code compiled from the `HEAD` of `master` into the `.libs` subdir of the code compiled from the `HEAD` of Issue 30
        6. Confirmed I was able to run `-test` and see “planarity wrapper from master” and “graphLib from Issue30 branch”. I then also ran `-t` (`TestAllGraphs()`) as well as `-r` (`RandomGraphs()`) successfully, meaning the const-correctness changes to function signatures of `sf_New()` and `beginG6ReadIterationFromG6FilePath()` didn’t result in incompatibilities.

        </details>
* `c/graphLib/io/g6-read-iterator.c`/`.h` - updated function signature of `beginG6ReadIterationFromG6FilePath()` to accept `char const *const infileName`, since neither the pointer nor the contents of the `infileName` string shall be modified.
* `c/graphLib/io/strOrFile.c`/`.h` - updated function signature of `sf_New()` to accept `char const *const theStr` (if we pass in a non-`NULL` string, that simply gets appended to the internal `strBuff`; we don't manipulate `theStr` passed in!) and `char const *const fileName` (want to signal to callers that we do not manipulate the pointer nor contents of the `fileName`). Also updated the docstring to clarify the use of the `strOrFile` container.
* `c/planarityApp/planarity.h` - 
    * moved `FILENAMEMAXLENGTH`, `ALGORITHMNAMEMAXLENGTH`, `SUFFIXMAXLENGTH` from `c/planarityApp/planarityUtils.c` so that they might be used throughout the `planarityApp` layer (since I needed `FILENAMEMAXLENGTH` in particular for `c/planarityApp/planarityMenu.c` to constrain how long the user-input input and output filenames should be)
    * Added `COMMANDSTRINGMAXLENGTH` to be used when constructing the format string for when we wish to read a command string from `stdin` via the menu-driven system.
    * Declare new functions `GetLineFromStdin()` and `GetCommandAndOptionalModifier()`, which are both defined in `c/planarityApp/planarityUtils.c`
    * Updated `planarityApp` endpoint function signatures for `SpecificGraph()`, `RandomGraph()`, and `TestAllGraphs()` to now accept `char const *const commandString`. Updated function signatures for `TransformGraph()` and `TestAllGraphs()` for const-correctness of how `commandString` is handled, as well as `infileName`.
    * Updated function signatures for `GetEmbedFlags()` and `AttachAlgorithm()` so that both now return `OK`/`NOTOK` to allow for better error handling; this means that `GetEmbedFlags()` now returns the `embedFlags` via the `int *` pointer rather than as the direct return value.
    * Updated function signature for `Reconfigure()` so that the function now returns an `int` to allow for better error handling via the menu (i.e. only return `NOTOK` on unrecoverable errors)
* `c/planarityApp/planarityCommandLine.c` - 
    * For Debug builds (i.e. compiled with `(/|-)DDEBUG` flag), `commandLine()` now waits for user input using `GetLineFromStdin()` (contents of `lineBuff` aren't used; only alternative would be to change how `GetLineFromStdin()` would respond to `NULL` for `lineBuff`, and this seemed less gross)
    * Updated how we parse `argv` for `callRandomGraphs()` and `callSpecificGraph()` so that we pass the entire `char *` from `argv` corresponding to the `commandString`; for `callRandomMaxPlanarGraph()` and `callRandomNonplanarGraph()`, we now pass the command string `"p"` to `RandomGraph()`. Changed how we detect `-q` flag to match how it is done in `commandLine()` (i.e. conditioning on `strncmp(argv[2], "-q", 2) == 0` to determine the `offset`). Added checks for when too many arguments are provided on the command-line. 
    * `runSpecificGraphTest()` now uses `GetCommandAndOptionalModifier()` on the `commandString` to extract the `command` to pass to `ConstructPrimaryOutputFilename()` as well as for post-processing to check the secondary output file for Draw Planar. Passes the `commandString` down into `SpecificGraph()`.
* `c/planarityApp/planarityMenu.c` -
    * Introduced pattern of using stack-allocated `lineBuff` to get `MAXLINE` `char`s of input from the user via `GetLineFromStdin()`, then parse the `lineBuff` using `sscanf()` (erroring on inappropriate input).
    * `TestAllGraphsMenu()` now uses `sscanf()` on `lineBuff` read using `GetLineFromStdin()` to extract command-string, which is directly sent to `TestAllGraphs()`
* `c/planarityApp/planarityRandomGraphs.c` -
    * Updated `RandomGraphs()` and `RandomGraph()` to both accept `char const *const commandString`, and to leverage `GetCommandAndOptionalModifier()` and `GetEmbedFlags()`. Also updated to incorporate error handling from `GetNumberIfZero()` (i.e. invalid user input or read error means we return 0)  
    * In `RandomGraphs()`, if `gp_CopyGraph()` fails, we now free and exit (since issues allocating or manipulating memory indicate a serious error has occurred and that we should not proceed with trying to embed the graph).
    * In `RandomGraph()`, when prompting the user whether they want to `saveEdgeListFormat`, now use a loop to `scanf()` of `lineBuff` from `GetLineFromStdin()` and handle bad input.
    * Updated `GetNumberIfZero()` function signature to return `int` for better error handling, and now use a loop around the call to `sscanf()` the `lineBuff` read using `GetLineFromStdin()` to request an integer from the user to re-prompt on bad input (rather than just coercing to the mean of `min` and `max`).
        * __**N.B.**__ This might be a good candidate for promotion to `planarityUtils.c`, even though it's currently only used in `planarityRandomGraphs.c`
    * `MakeGraph()` - Now calls `AttachAlgorithm()` rather than duplicating the switch statement; improved error handling and messaging 
* `c/planarityApp/planaritySpecificGraph.c` - updated `SpecificGraph()` to accept `char const *const commandString`, and use `GetCommandAndOptionalModifier()` to extract the `command` and optional `modifier` to pass to `GetEmbedFlags()`, and the `command` to pass to `AttachAlgorithm()`. Now also handle errors indicated by the return code of these functions. Also handle error state from `ConstructInputFilename()`
    * __**N.B.**__ `char const *infileName` was not changed to `char const * const infileName` because we *always* assign the pointer to the global `theFileName` string to it from `ConstructInputFilename()` (either construct new filename, or `strcpy()` the `infileName` into `theFileName`).
* `c/planarityApp/planarityTestAllGraphs.c` - updated `testAllGraphs()` and `outputTestAllGraphsResults()` so that they are sensitive to whether an optional `modifier` character was supplied alongside the `command` (both extracted from the `commandString` in `TestAllGraphs()`). Updated `testAllGraphs()` so that it uses the new `GetEmbedFlags()` and `AttachAlgorithm()` and handles errors that arise from invalid `commandString` being supplied to `TestAllGraphs()`
* `c/planarityApp/planarityTransformGraph.c` - updated `TransformGraph()` function signature to signal that neither the contents of nor the pointer to the `commandString` nor `infileName` will be modified in the scope of the function
* `c/planarityApp/planarityUtils.c` -
    * Added `GetLineFromStdin()`, which uses `fgets()` to populate the `lineBuff` from `stdin`, and replaces the first `\n` or `\r` character with `\0` to terminate the string (so those chars are consumed from `stdin`, but then don't affect subsequent parsing the `lineBuff`)
    * Added `GetCommandAndOptionalModifier()`, which takes the `commandString` and throws away the `-` if passed in, extracts one character for the `command`, and optionally extracts the `modifier` character if provided.
        * __**N.B.**__ This *may* have to change in the future if we want to specify more than one character for the command itself (e.g. projective planarity as `pp`, or $K_5$ minor search with embedder having command string `5me`, i.e.,command `5m` with modifier `e`)
    * Updated `Reconfigure()` to process user input by fetching from `stdin` using `GetLineFromStdin()` and calling `sscanf()` to ensure the user input what was expected. If unrecoverable error occurs (such as failing to read from `stdin`), then we bail out; otherwise, we re-prompt the user until they provide valid input.
    * Updated `GetEmbedFlags()` so that it accepts the `command` and optional `modifier` to set the appropriate flags in the `embedFlags` (must dereference `int *embedFlagsP`). Now, the `int` returned signals success or if an error was encountered, which must be handled by the caller.
        * __**N.B.**__ Only when the $K_{3,3}$ certifier branch is merged will we actually allow a modifier character to be supplied in the `commandString`
    * Updated `AttachAlgorithm()` so that it now returns an `int` to signal success or if an error was encountered, which must be handled by the caller. Also modified the `switch` statement so that planarity and outerplanarity are explicitly included, in case in the future we decide to break those out into extensions that aren't attached by default.
    * Updated `ConstructInputFilename()` so that it uses the format-format string pattern to ensure we are reading a sensible number of characters from the user input when we `sscanf()` the `lineBuff` populated by `GetLineFromStdin()`. Also now loop to re-prompt user if input is invalid, and only return `NOTOK` on unrecoverable error.

### "While I was in the neighbourhood" changes

* `c/graphLib/io/graphIO.c` - removed comment about how we want to handle when `sf_fgets()` returns `NULL` and created #187 : it might be beneficial for `sf_fgets()` to set `errno` when a read error is encountered rather than just returning `NULL`, and for functions using the `strOrFile` container to distinguish between "no more content to read" and "unable to read due to error". Also updated functions so that I `memset()` stack-allocated arrays with appropriate default values, and that the return code variable is is initialized with value `OK`.
* `c/planarityApp/planarityMenu.c` - 
    * Updated `TransformGraphMenu()` and `TestAllGraphsMenu()` to use the format-format string pattern to restrict how many characters are allowed to be read for the user-supplied `infileName` and `outfileName`.
    * Updated function signatures of `Reconfigure()`, `TransformGraphMenu()`, and `TestAllGraphsMenu()` so that they return `int` for better error handling
        * __**N.B.**__ There still is a bit of tension here for these two sub-menus, as I'm not breaking out of the main `menu()` loop if we encounter `NOTOK` (which might arise from a serious error like `GetLineFromStdin()` failing, or might occur when a user simply mistypes a file input name). Might require further examination to make error handling more fine-grained and robust.
    * Now makes sure to translate `OK`/`NONEMBEDDABLE`/`NOTOK` to the expected process exit codes to return to `planarity.c`'s `main()`
    * Changed user input retry mechanism so that I'm using `while (1) {...}` loop, only breaking out if an unrecoverable error has occurred (in which case `Result` is set to `NOTOK`, and some subsequent steps are skipped before cleanup and returning to caller) or if the input was determined to be valid.
    * I use the `memset()` pattern on stack-allocated arrays to initialize (e.g. stack-allocated strings are set to only contain `\0`)
* Moved `WriteAlgorithmResults()` from `c/planarityApp/planaritySpecificGraph.c` to `c/planarityApp/planarityUtils.c` since it's used in both `planarityRandomGraphs.c` and `planaritySpecificGraph.c`
* Minor source code formatting changes to `ConstructPrimaryOutputFilename()` and `ConstructTransformationExpectedResultFilename()` in `c/planarityApp/planarityUtils.c`
* In `c/planarityApp/planarityUtils.c`, updated `FlushConsole()` so that it returns when one tries to `fflush(stdin)` (Since according to [the standard](https://www.iso-9899.info/wiki/Implementation), this [constitutes undefined behaviour](https://faq.cprogramming.com/cgi-bin/smartfaq.cgi?answer=1052863818&id=1043284351)). Note that comparing `FILE *` using `==` against the macro for `stdin` *should* suffice (i.e. don't think we have to dig so deep as to look at file descriptors, but I *could* be wrong).
* `c/planarityApp/planarityCommandLine.c` - 
    * I've tried to make sure that all functions called from `commandLine()` are returning `OK`/`NONEMBEDDABLE`/`NOTOK`, since those get translated to valid process exit codes at the end of that function. For example, `runSpecificGraphTests()` was returning `-1` in the case of an error state, which (currently) corresponds to `NONEMBEDDABLE`, and would therefore cause the process to return `1` rather than `-1`. I also updated and `runGraphTransformationTest()` so that it returns `OK` or `NOTOK` (error state), and `runSpecificGraphTest()` so that it returns `OK` when the result was either `OK` or `NONEMBEDDABLE`, and `NOTOK` when an error occurred. Finally, I updated all of the `call` functions, i.e. `callRandomGraphs()`, `callSpecificGraph()`, `callRandomMaxPlanarGraph()`, `callRandomNonplanarGraph()`, `callTransformGraph()`, and `callTestAllGraphs()`, so that they return `NOTOK` on error (e.g. incorrect number of arguments)
* `c/planarityApp/planarityTestAllGraphs.c` - Updated allocation of `resultsStr` in `outputTestAllGraphsResults()` so that it's clear about what we're trying to accommodate

### DIFF
--- a/c/graphLib/io/strOrFile.c
+++ b/c/graphLib/io/strOrFile.c
@@ -31,7 +31,7 @@ See the LICENSE.TXT file for licensing information.
  string-or-file object and can be obtained after all output has been written
  using sf_takeTheStr(). If the fileName is non-NULL and nonempty (which may
  correspond to stdout or stderr), then the output will go to the output stream
- and an output string will not be constructured.
+ and an output string will not be constructed.
 
  Returns the allocated string-or-file container, or NULL on error.
  ********************************************************************/
--- a/c/planarityApp/planarityCommandLine.c
+++ b/c/planarityApp/planarityCommandLine.c
@@ -218,56 +218,56 @@ int runSpecificGraphTests(char const *samplesDir)
 
     if (runSpecificGraphTest("-p", "maxPlanar5.txt", TRUE) != OK)
     {
-        Message("Planarity test on maxPlanar5.txt failed.\n");
+        ErrorMessage("Planarity test on maxPlanar5.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-d", "maxPlanar5.txt", FALSE) != OK)
     {
-        Message("Graph drawing test maxPlanar5.txt failed.\n");
+        ErrorMessage("Graph drawing test maxPlanar5.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-d", "drawExample.txt", TRUE) != OK)
     {
-        Message("Graph drawing on drawExample.txt failed.\n");
+        ErrorMessage("Graph drawing on drawExample.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-p", "Petersen.txt", FALSE) != OK)
     {
-        Message("Planarity test on Petersen.txt failed.\n");
+        ErrorMessage("Planarity test on Petersen.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-o", "Petersen.txt", TRUE) != OK)
     {
-        Message("Outerplanarity test on Petersen.txt failed.\n");
+        ErrorMessage("Outerplanarity test on Petersen.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-2", "Petersen.txt", FALSE) != OK)
     {
-        Message("K_{2,3} search on Petersen.txt failed.\n");
+        ErrorMessage("K_{2,3} search on Petersen.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-3", "Petersen.txt", TRUE) != OK)
     {
-        Message("K_{3,3} search on Petersen.txt failed.\n");
+        ErrorMessage("K_{3,3} search on Petersen.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-4", "Petersen.txt", FALSE) != OK)
     {
-        Message("K_4 search on Petersen.txt failed.\n");
+        ErrorMessage("K_4 search on Petersen.txt failed.\n");
 
         retVal = NOTOK;
     }
@@ -277,56 +277,56 @@ int runSpecificGraphTests(char const *samplesDir)
 
     if (runSpecificGraphTest("-p", "maxPlanar5.0-based.txt", FALSE) != OK)
     {
-        Message("Planarity test on maxPlanar5.0-based.txt failed.\n");
+        ErrorMessage("Planarity test on maxPlanar5.0-based.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-d", "maxPlanar5.0-based.txt", TRUE) != OK)
     {
-        Message("Graph drawing test maxPlanar5.0-based.txt failed.\n");
+        ErrorMessage("Graph drawing test maxPlanar5.0-based.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-d", "drawExample.0-based.txt", FALSE) != OK)
     {
-        Message("Graph drawing on drawExample.0-based.txt failed.\n");
+        ErrorMessage("Graph drawing on drawExample.0-based.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-p", "Petersen.0-based.txt", TRUE) != OK)
     {
-        Message("Planarity test on Petersen.0-based.txt failed.\n");
+        ErrorMessage("Planarity test on Petersen.0-based.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-o", "Petersen.0-based.txt", FALSE) != OK)
     {
-        Message("Outerplanarity test on Petersen.0-based.txt failed.\n");
+        ErrorMessage("Outerplanarity test on Petersen.0-based.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-2", "Petersen.0-based.txt", TRUE) != OK)
     {
-        Message("K_{2,3} search on Petersen.0-based.txt failed.\n");
+        ErrorMessage("K_{2,3} search on Petersen.0-based.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-3", "Petersen.0-based.txt", FALSE) != OK)
     {
-        Message("K_{3,3} search on Petersen.0-based.txt failed.\n");
+        ErrorMessage("K_{3,3} search on Petersen.0-based.txt failed.\n");
 
         retVal = NOTOK;
     }
 
     if (runSpecificGraphTest("-4", "Petersen.0-based.txt", TRUE) != OK)
     {
-        Message("K_4 search on Petersen.0-based.txt failed.\n");
+        ErrorMessage("K_4 search on Petersen.0-based.txt failed.\n");
 
         retVal = NOTOK;
     }
@@ -339,7 +339,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading file contents into string
     if (runGraphTransformationTest("-a", "nauty_example.g6", TRUE) != OK)
     {
-        Message("Transforming nauty_example.g6 file contents as string to adjacency list failed.\n");
+        ErrorMessage("Transforming nauty_example.g6 file contents as string to adjacency list failed.\n");
 
         retVal = NOTOK;
     }
@@ -347,7 +347,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading from file
     if (runGraphTransformationTest("-a", "nauty_example.g6", FALSE) != OK)
     {
-        Message("Transforming nauty_example.g6 using file pointer to adjacency list failed.\n");
+        ErrorMessage("Transforming nauty_example.g6 using file pointer to adjacency list failed.\n");
 
         retVal = NOTOK;
     }
@@ -355,7 +355,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading first graph from file into string
     if (runGraphTransformationTest("-a", "N5-all.g6", TRUE) != OK)
     {
-        Message("Transforming first graph in N5-all.g6 (read as string) to adjacency list failed.\n");
+        ErrorMessage("Transforming first graph in N5-all.g6 (read as string) to adjacency list failed.\n");
 
         retVal = NOTOK;
     }
@@ -363,7 +363,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading first graph from file pointer
     if (runGraphTransformationTest("-a", "N5-all.g6", FALSE) != OK)
     {
-        Message("Transforming first graph in N5-all.g6 (read from file pointer) to adjacency list failed.\n");
+        ErrorMessage("Transforming first graph in N5-all.g6 (read from file pointer) to adjacency list failed.\n");
 
         retVal = NOTOK;
     }
@@ -371,7 +371,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading file contents corresponding to dense graph into string
     if (runGraphTransformationTest("-a", "K10.g6", TRUE) != OK)
     {
-        Message("Transforming K10.g6 file contents as string to adjacency list failed.\n");
+        ErrorMessage("Transforming K10.g6 file contents as string to adjacency list failed.\n");
 
         retVal = NOTOK;
     }
@@ -379,7 +379,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading dense graph from file
     if (runGraphTransformationTest("-a", "K10.g6", FALSE) != OK)
     {
-        Message("Transforming K10.g6 using file pointer to adjacency list failed.\n");
+        ErrorMessage("Transforming K10.g6 using file pointer to adjacency list failed.\n");
 
         retVal = NOTOK;
     }
@@ -389,7 +389,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading file contents into string
     if (runGraphTransformationTest("-m", "nauty_example.g6", TRUE) != OK)
     {
-        Message("Transforming nauty_example.g6 file contents as string to adjacency matrix failed.\n");
+        ErrorMessage("Transforming nauty_example.g6 file contents as string to adjacency matrix failed.\n");
 
         retVal = NOTOK;
     }
@@ -397,7 +397,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading from file
     if (runGraphTransformationTest("-m", "nauty_example.g6", FALSE) != OK)
     {
-        Message("Transforming nauty_example.g6 using file pointer to adjacency matrix failed.\n");
+        ErrorMessage("Transforming nauty_example.g6 using file pointer to adjacency matrix failed.\n");
 
         retVal = NOTOK;
     }
@@ -405,7 +405,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading first graph from file into string
     if (runGraphTransformationTest("-m", "N5-all.g6", TRUE) != OK)
     {
-        Message("Transforming first graph in N5-all.g6 (read as string) to adjacency matrix failed.\n");
+        ErrorMessage("Transforming first graph in N5-all.g6 (read as string) to adjacency matrix failed.\n");
 
         retVal = NOTOK;
     }
@@ -413,7 +413,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading first graph from file pointer
     if (runGraphTransformationTest("-m", "N5-all.g6", FALSE) != OK)
     {
-        Message("Transforming first graph in N5-all.g6 (read from file pointer) to adjacency matrix failed.\n");
+        ErrorMessage("Transforming first graph in N5-all.g6 (read from file pointer) to adjacency matrix failed.\n");
 
         retVal = NOTOK;
     }
@@ -421,7 +421,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading file contents corresponding to dense graph into string
     if (runGraphTransformationTest("-m", "K10.g6", TRUE) != OK)
     {
-        Message("Transforming K10.g6 file contents as string to adjacency matrix failed.\n");
+        ErrorMessage("Transforming K10.g6 file contents as string to adjacency matrix failed.\n");
 
         retVal = NOTOK;
     }
@@ -429,7 +429,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading dense graph from file
     if (runGraphTransformationTest("-m", "K10.g6", FALSE) != OK)
     {
-        Message("Transforming K10.g6 using file pointer to adjacency matrix failed.\n");
+        ErrorMessage("Transforming K10.g6 using file pointer to adjacency matrix failed.\n");
 
         retVal = NOTOK;
     }
@@ -439,7 +439,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading from file
     if (runGraphTransformationTest("-g", "nauty_example.g6.0-based.AdjList.out.txt", TRUE) != OK)
     {
-        Message("Transforming nauty_example.g6.0-based.AdjList.out.txt using file pointer to .g6 failed.\n");
+        ErrorMessage("Transforming nauty_example.g6.0-based.AdjList.out.txt using file pointer to .g6 failed.\n");
 
         retVal = NOTOK;
     }
@@ -447,7 +447,7 @@ int runSpecificGraphTests(char const *samplesDir)
     // runGraphTransformationTest by reading from file
     if (runGraphTransformationTest("-g", "K10.g6.0-based.AdjList.out.txt", TRUE) != OK)
     {
-        Message("Transforming K10.g6.0-based.AdjList.out.txt using file pointer to .g6 failed.\n");
+        ErrorMessage("Transforming K10.g6.0-based.AdjList.out.txt using file pointer to .g6 failed.\n");
 
         retVal = NOTOK;
     }
@@ -474,7 +474,7 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int 
 
     if (GetCommandAndOptionalModifier(commandString, &command, &modifier) != OK)
     {
-        Message("Unable to extract command (and optionally modifier) from command string.\n");
+        ErrorMessage("Unable to extract command (and optionally modifier) from command string.\n");
 
         return NOTOK;
     }
@@ -490,7 +490,7 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int 
         inputString = ReadTextFileIntoString(infileName);
         if (inputString == NULL)
         {
-            Message("Failed to read input file into string.\n");
+            ErrorMessage("Failed to read input file into string.\n");
 
             Result = NOTOK;
         }
@@ -507,7 +507,8 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int 
 
     if (Result != OK && Result != NONEMBEDDABLE)
     {
-        Message("Test failed (graph processor returned failure result).\n");
+        ErrorMessage("Test failed (graph processor returned failure result).\n");
+        Result = NOTOK;
     }
     else
     {
@@ -516,7 +517,7 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int 
             Message("Test succeeded (result equal to exemplar).\n");
         else
         {
-            Message("Test failed (result not equal to exemplar).\n");
+            ErrorMessage("Test failed (result not equal to exemplar).\n");
 
             Result = NOTOK;
         }
@@ -529,7 +530,7 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int 
 
         if (expectedSecondaryResultFileName == NULL)
         {
-            Message("Unable to allocate memory for expected secondary output filename.\n");
+            ErrorMessage("Unable to allocate memory for expected secondary output filename.\n");
 
             Result = NOTOK;
         }
@@ -541,25 +542,38 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int 
                 Message("Test succeeded (secondary result equal to exemplar).\n");
             else
             {
-                Message("Test failed (secondary result not equal to exemplar).\n");
+                ErrorMessage("Test failed (secondary result not equal to exemplar).\n");
 
                 Result = NOTOK;
             }
 
-            free(expectedSecondaryResultFileName);
+            if (expectedSecondaryResultFileName != NULL)
+            {
+                free(expectedSecondaryResultFileName);
+                expectedSecondaryResultFileName = NULL;
+            }
         }
     }
 
     Message("\n");
 
     if (inputString != NULL)
+    {
         free(inputString);
+        inputString = NULL;
+    }
 
     if (actualOutput != NULL)
+    {
         free(actualOutput);
+        actualOutput = NULL;
+    }
 
     if (actualOutput2 != NULL)
+    {
         free(actualOutput2);
+        actualOutput2 = NULL;
+    }
 
     // NOTE: Test run successfully if OK or NONEMBEDDABLE result; Result is only
     // NOTOK when an error occurs during one of the subordinate function calls,
@@ -578,7 +592,7 @@ int runGraphTransformationTest(char const *command, char const *infileName, int 
     // input graph; it will only support "-(gam)"
     if (command == NULL || strlen(command) < 2)
     {
-        Message("runGraphTransformationTest only supports -(gam).\n");
+        ErrorMessage("runGraphTransformationTest only supports -(gam).\n");
 
         return NOTOK;
     }
@@ -592,7 +606,7 @@ int runGraphTransformationTest(char const *command, char const *infileName, int 
         inputString = ReadTextFileIntoString(infileName);
         if (inputString == NULL)
         {
-            Message("Failed to read input file into string.\n");
+            ErrorMessage("Failed to read input file into string.\n");
 
             Result = NOTOK;
         }
@@ -611,7 +625,9 @@ int runGraphTransformationTest(char const *command, char const *infileName, int 
 
         if (Result != OK || actualOutput == NULL)
         {
-            Message("Failed to perform transformation.\n");
+            ErrorMessage("Failed to perform transformation.\n");
+
+            Result = NOTOK;
         }
         else
         {
@@ -626,7 +642,7 @@ int runGraphTransformationTest(char const *command, char const *infileName, int 
 
             if (Result != OK || expectedOutfileName == NULL)
             {
-                Message("Unable to construct output filename for expected transformation output.\n");
+                ErrorMessage("Unable to construct output filename for expected transformation output.\n");
 
                 Result = NOTOK;
             }
@@ -654,7 +670,7 @@ int runGraphTransformationTest(char const *command, char const *infileName, int 
 #pragma GCC diagnostic ignored "-Wformat-nonliteral"
                     sprintf(messageContents, messageFormat, command, charsAvailForFilename, infileName);
 #pragma GCC diagnostic pop
-                    Message(messageContents);
+                    ErrorMessage(messageContents);
 
                     Result = NOTOK;
                 }
@@ -666,7 +682,10 @@ int runGraphTransformationTest(char const *command, char const *infileName, int 
                 }
 
                 if (actualOutput != NULL)
+                {
                     free(actualOutput);
+                    actualOutput = NULL;
+                }
             }
         }
     }
@@ -674,7 +693,10 @@ int runGraphTransformationTest(char const *command, char const *infileName, int 
     Message("\n");
 
     if (inputString != NULL)
+    {
         free(inputString);
+        inputString = NULL;
+    }
 
     return Result;
 }
@@ -777,7 +799,7 @@ int callRandomMaxPlanarGraph(int argc, char *argv[])
     if (argc == 5 + offset)
         outfile2Name = argv[4 + offset];
 
-    return RandomGraph("p", 0, numVertices, outfileName, outfile2Name);
+    return RandomGraph("-p", 0, numVertices, outfileName, outfile2Name);
 }
 
 /****************************************************************************
@@ -810,7 +832,7 @@ int callRandomNonplanarGraph(int argc, char *argv[])
     if (argc == 5 + offset)
         outfile2Name = argv[4 + offset];
 
-    return RandomGraph("p", 1, numVertices, outfileName, outfile2Name);
+    return RandomGraph("-p", 1, numVertices, outfileName, outfile2Name);
 }
 
 /****************************************************************************
--- a/c/planarityApp/planarityMenu.c
+++ b/c/planarityApp/planarityMenu.c
@@ -23,7 +23,7 @@ int menu(void)
     char *choiceStringFormat = NULL;
     char choiceString[COMMANDSTRINGMAXLENGTH + 1];
 
-    char *commandString = NULL, *secondOutfile = NULL;
+    char *secondOutfile = NULL;
 
     char command = '\0';
 
@@ -117,10 +117,10 @@ int menu(void)
                 break;
             else
             {
-                commandString = choiceString;
+                char *commandString = choiceString;
                 if (GetCommandAndOptionalModifier(commandString, &command, NULL) != OK)
                 {
-                    Message("Unable to extract command from choice, please retry.\n");
+                    ErrorMessage("Unable to extract command from choice, please retry.\n");
                     commandString = NULL;
 
                     continue;
@@ -130,7 +130,7 @@ int menu(void)
                     secondOutfile = (char *)"";
 
                 if (!strchr(GetAlgorithmChoices(), command))
-                    Message("Invalid algorithm command choice, please retry.\n");
+                    ErrorMessage("Invalid algorithm command choice, please retry.\n");
 
                 else
                 {
@@ -146,7 +146,7 @@ int menu(void)
                         Result = RandomGraph(commandString, 0, 0, NULL, NULL);
                         break;
                     case 'n':
-                        Result = RandomGraph(choiceString, 1, 0, NULL, NULL);
+                        Result = RandomGraph(commandString, 1, 0, NULL, NULL);
                         break;
                     default:
                         break;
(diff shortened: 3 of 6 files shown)


wbkboyer:
Ran through Address Sanitizer tests for command-line access of `planarityApp` endpoints on Debian 13 (unstable Sid) and didn't encounter any leaks. However, when I went to check whether there were any wee beasties lurking behind the menu-driven access, I discovered an issue when I tried to call `SpecificGraph()` (Expand details for logs)

<details>

```
==4657==ERROR: AddressSanitizer: global-buffer-overflow on address 0x55e1e2a50d03 at pc 0x7f3c1cca2cbc bp 0x7ffec8bdc9a0 sp 0x7ffec8bdc160
WRITE of size 1025 at 0x55e1e2a50d03 thread T0
    #0 0x7f3c1cca2cbb in scanf_common ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors_format.inc:342
    #1 0x7f3c1cce33b8 in __isoc23_vsscanf ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:1558
    #2 0x7f3c1cce3b7c in __isoc23_sscanf ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:1590
    #3 0x55e1e29bfcfc in ConstructInputFilename /home/wbkboyer/git/edge-addition-planarity-suite/c/planarityApp/planarityUtils.c:737
    #4 0x55e1e29bacca in SpecificGraph /home/wbkboyer/git/edge-addition-planarity-suite/c/planarityApp/planaritySpecificGraph.c:71
    #5 0x55e1e29b5abe in menu /home/wbkboyer/git/edge-addition-planarity-suite/c/planarityApp/planarityMenu.c:78
    #6 0x55e1e29b2539 in main /home/wbkboyer/git/edge-addition-planarity-suite/c/planarityApp/planarity.c:18
    #7 0x7f3c1ca33ca7  (/lib/x86_64-linux-gnu/libc.so.6+0x29ca7) (BuildId: def5460e3cee00bfee25b429c97bcc4853e5b3a8)
    #8 0x7f3c1ca33d64 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29d64) (BuildId: def5460e3cee00bfee25b429c97bcc4853e5b3a8)
    #9 0x55e1e29b2450 in _start (/home/wbkboyer/git/edge-addition-planarity-suite/Debug/planarity+0xc450) (BuildId: 12f97b666fdbabe5b59b9f62facf6cf10cd382fe)

0x55e1e2a50d03 is located 0 bytes after global variable 'theFileName' defined in '/home/wbkboyer/git/edge-addition-planarity-suite/c/planarityApp/planarityUtils.c:694:6' (0x55e1e2a50c40) of size 195
SUMMARY: AddressSanitizer: global-buffer-overflow /home/wbkboyer/git/edge-addition-planarity-suite/c/planarityApp/planarityUtils.c:737 in ConstructInputFilename
Shadow bytes around the buggy address:
  0x55e1e2a50a80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x55e1e2a50b00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x55e1e2a50b80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x55e1e2a50c00: 01 f9 f9 f9 f9 f9 f9 f9 00 00 00 00 00 00 00 00
  0x55e1e2a50c80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x55e1e2a50d00:[03]f9 f9 f9 f9 f9 f9 f9 00 00 00 00 00 00 00 00
  0x55e1e2a50d80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x55e1e2a50e00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x55e1e2a50e80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x55e1e2a50f00: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x55e1e2a50f80: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==4657==ABORTING

```

Reproduced on macOS with `clang` via VSCode Debug build configuration:

```
 % ASAN_OPTIONS=detect_leaks=1:symbolize=1 ASAN_SYMBOLIZER_PATH=/opt/homebrew/opt/llvm/bin/llvm-symbolizer ./Debug/planarity 
planarity(30586,0x1ef55a140) malloc: nano zone abandoned due to inability to reserve vm space.

===================================================
The Edge Addition Planarity Suite version 4.0.2.0
based on libPlanarity graph library version 3:0:1
Copyright (c) 1997-2025 by John M. Boyer
All rights reserved.
See the LICENSE.TXT file for licensing information.
Contact info: jboyer at acm.org
===================================================
P. Planar embedding and Kuratowski subgraph isolation
D. Planar graph drawing by visibility representation
O. Outerplanar embedding and obstruction isolation
2. Search for subgraph homeomorphic to K_{2,3}
3. Search for subgraph homeomorphic to K_{3,3}
4. Search for subgraph homeomorphic to K_4
X. Transform single graph in supported file to .g6, adjacency list, or adjacency matrix
T. Perform an algorithm test on all graphs in .g6 input file
H. Help message for command line version
R. Reconfigure options
Q. Quit

Enter Choice: r

Do you want to 
  Randomly generate graphs (r),
  Specify a graph (s),
  Randomly generate a maximal planar graph (m), or
  Randomly generate a non-planar graph (n)?
	s

===================================================
The Edge Addition Planarity Suite version 4.0.2.0
based on libPlanarity graph library version 3:0:1
Copyright (c) 1997-2025 by John M. Boyer
All rights reserved.
See the LICENSE.TXT file for licensing information.
Contact info: jboyer at acm.org
===================================================
P. Planar embedding and Kuratowski subgraph isolation
D. Planar graph drawing by visibility representation
O. Outerplanar embedding and obstruction isolation
2. Search for subgraph homeomorphic to K_{2,3}
3. Search for subgraph homeomorphic to K_{3,3}
4. Search for subgraph homeomorphic to K_4
X. Transform single graph in supported file to .g6, adjacency list, or adjacency matrix
T. Perform an algorithm test on all graphs in .g6 input file
H. Help message for command line version
R. Reconfigure options
Q. Quit

Enter Choice: p
Enter graph file name: c/samples/K33.AdjList.txt
=================================================================
==30586==ERROR: AddressSanitizer: global-buffer-overflow on address 0x0001004065a3 at pc 0x0001009adad4 bp 0x00016fc65e10 sp 0x00016fc65590
WRITE of size 1025 at 0x0001004065a3 thread T0
    #0 0x0001009adad0 in scanf_common(void*, int, bool, char const*, char*) (/opt/homebrew/Cellar/llvm/21.1.2/lib/clang/21/lib/darwin/libclang_rt.asan_osx_dynamic.dylib:arm64+0x21ad0)
    #1 0x0001009ae104 in sscanf (/opt/homebrew/Cellar/llvm/21.1.2/lib/clang/21/lib/darwin/libclang_rt.asan_osx_dynamic.dylib:arm64+0x22104)
    #2 0x0001001ae45c in ConstructInputFilename /Users/wbkboyer/git/edge-addition-planarity-suite-fork/c/planarityApp/planarityUtils.c:737:17
    #3 0x0001001a6368 in SpecificGraph /Users/wbkboyer/git/edge-addition-planarity-suite-fork/c/planarityApp/planaritySpecificGraph.c:71:30
    #4 0x00010019f96c in menu /Users/wbkboyer/git/edge-addition-planarity-suite-fork/c/planarityApp/planarityMenu.c:78:21
    #5 0x000100198894 in main /Users/wbkboyer/git/edge-addition-planarity-suite-fork/c/planarityApp/planarity.c:18:18
    #6 0x00018125eb94  (<unknown module>)

0x0001004065a3 is located 0 bytes after global variable 'theFileName' defined in '/Users/wbkboyer/git/edge-addition-planarity-suite-fork/c/planarityApp/planarityUtils.c:694' (0x0001004064e0) of size 195
SUMMARY: AddressSanitizer: global-buffer-overflow /Users/wbkboyer/git/edge-addition-planarity-suite-fork/c/planarityApp/planarityUtils.c:737:17 in ConstructInputFilename
Shadow bytes around the buggy address:
  0x000100406300: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x000100406380: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x000100406400: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x000100406480: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x000100406500: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x000100406580: 00 00 00 00[03]f9 f9 f9 f9 f9 f9 f9 00 00 00 00
  0x000100406600: 04 f9 f9 f9 00 00 00 00 04 f9 f9 f9 00 00 00 00
  0x000100406680: 04 f9 f9 f9 00 00 00 00 04 f9 f9 f9 00 00 00 00
  0x000100406700: 04 f9 f9 f9 00 00 00 00 00 00 00 00 00 00 00 00
  0x000100406780: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x000100406800: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==30586==ABORTING
zsh: abort      ASAN_OPTIONS=detect_leaks=1:symbolize=1 ASAN_SYMBOLIZER_PATH= 
```

</details>

It appears as if there's something questionable about my changes to `ConstructInputFilename()`. _Further investigation coming soon_

c/planarityApp/planarityCommandLine.c
```diff
@@ -416,18 +423,23 @@ int runSpecificGraphTests(char const *samplesDir)
...
-    char algorithmCode = command[1];
+    char *inputString = NULL, *actualOutput = NULL, *actualOutput2 = NULL;
+    char const *expectedPrimaryResultFileName = "";
+
+    char command = '\0', modifier = '\0';
+
+    if (GetCommandAndOptionalModifier(commandString, &command, &modifier) != OK)
+    {
+        Message("Unable to extract command (and optionally modifier) from command string.\n");
+        return -1;
```
john-boyer-phd: Thank you for noticing and pointing out in our live code review that you have returned -1 for consistency with the existing code (always good) but that the return should actually be NOTOK for all of these -1 occurrences. A long time ago, I made some changes to how the return codes worked, but inadvertently didn't change these returns correctly. So, since more returns have to be added, and they should be consistent, it would be best to fix all the -1 returns that should be return NOTOK instead. The current return -1 will incorrectly cause the process to return an indication of successfully processing a NONEMBEDDABLE graph, but we want the process to return a failure code indicating that the command line was incorrect.

c/planarityApp/planarityUtils.c
```diff
@@ -37,29 +44,88 @@ void Reconfigure(void)
...
+                ErrorMessage(reconfigureErrorMessage);
+                return;
+            }
         }
 
         Prompt("Do you want adj. matrix of embeddable graphs in directory 'embedded' (last 10 max))? (y/n) ");
-        scanf(" %c", &EmbeddableOut);
+        if (GetLineFromStdin(lineBuff, MAXLINE) != OK ||
+            strlen(lineBuff) != 1 ||
+            sscanf(lineBuff, " %c", &EmbeddableOut) != 1)
```
john-boyer-phd: May as well also error check with strchr on tolower of input char, to ensure yn response.

c/planarityApp/planarityUtils.c
```diff
@@ -37,29 +44,88 @@ void Reconfigure(void)
...
+        {
+            ErrorMessage(reconfigureErrorMessage);
+            return;
+        }
 
         Prompt("Do you want adj. matrix of obstructed graphs in directory 'obstructed' (last 10 max)? (y/n) ");
-        scanf(" %c", &ObstructedOut);
+        if (GetLineFromStdin(lineBuff, MAXLINE) != OK ||
+            ferror(stdin) || strlen(lineBuff) != 1 ||
+            sscanf(lineBuff, " %c", &ObstructedOut) != 1)
```
john-boyer-phd: May as well also error check with strchr on tolower of input char, to ensure yn response.

c/planarityApp/planarityUtils.c
```diff
@@ -37,29 +44,88 @@ void Reconfigure(void)
...
+        {
+            ErrorMessage(reconfigureErrorMessage);
+            return;
+        }
 
         Prompt("Do you want adjacency list format of embeddings in directory 'adjlist' (last 10 max)? (y/n) ");
-        scanf(" %c", &AdjListsForEmbeddingsOut);
+        if (GetLineFromStdin(lineBuff, MAXLINE) != OK ||
+            strlen(lineBuff) != 1 ||
+            sscanf(lineBuff, " %c", &AdjListsForEmbeddingsOut) != 1)
```
john-boyer-phd: May as well also error check with strchr on tolower of input char, to ensure yn response.

c/planarityApp/planarityUtils.c
```diff
@@ -487,27 +641,48 @@ char const *GetBaseName(int baseFlag)
...
+ * used to signal that the behaviour of the main extension should be modified.
+ *
+ * Returns OK if graph algorithm extension corresponding to the command char
+ * was successfully attached. Returns NOTOK if theGraph is not properly
+ * initialized, if attaching the graph algorithm extension failed, or if an
+ * invalid command char was supplied.
  ****************************************************************************/
 
-void AttachAlgorithm(graphP theGraph, char command)
+int AttachAlgorithm(graphP theGraph, char command)
```
john-boyer-phd: Much improved.

c/planarityApp/planarityUtils.c
```diff
@@ -532,13 +703,43 @@ char theFileName[FILENAMEMAXLENGTH + 1 + ALGORITHMNAMEMAXLENGTH + 1 + SUFFIXMAXL
 
 char *ConstructInputFilename(char const *infileName)
 {
+    int numCharsToReprFILENAMEMAXLENGTH = 0;
+    char const *fileNameFormatFormat = " %%%d[^\r\n]";
+    char *fileNameFormat = NULL;
```
john-boyer-phd: Do you really need malloc/free in here, i.e., can you just do 
char fileNameFormat[strlen(fileNameFormatFormat) + numCharsToReprFILENAMEMAXLENGTH + 1];

c/planarityApp/planarityUtils.c
```diff
@@ -681,3 +893,57 @@ int ConstructTransformationExpectedResultFilename(char const *infileName, char *
 
     return Result;
 }
+
+/****************************************************************************
+ * WriteAlgorithmResults()
+ ****************************************************************************/
+
+void WriteAlgorithmResults(graphP theGraph, int Result, char command, platform_time start, platform_time end, char const *infileName)
```
john-boyer-phd: Great you moved this function into this file.

c/graphLib/io/graphIO.c
```diff
@@ -539,7 +539,10 @@ int _ReadGraph(graphP theGraph, strOrFileP inputContainer)
                     RetVal = NOTOK;
                 else
                 {
-                    // FIXME: how do I distinguish between "there's no more content on input stream" and "I've hit an error state"
+                    // TODO: how do I distinguish between "there's no more
```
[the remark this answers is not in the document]
wbkboyer: Right on, I created #187 to capture this!

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -291,7 +335,7 @@ int outputTestAllGraphsResults(char command, testAllStatsP stats, char *infileNa
 
     resultsStr = (char *)malloc(
         (
-            3 + numCharsToReprNumGraphsRead +
+            4 + numCharsToReprNumGraphsRead +
```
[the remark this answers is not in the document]
wbkboyer: This is a great question! I rolled the 1 character for `-` and the 1 character for the subsequent space into the calculation for the number of characters required to express the command that was run; for example, `-3 ` requires 3 characters, and `-3e ` requires 4 characters, so I just said "allocate 4 upfront"!

Should I break this malloc out further with comments for what each part of the allocation calculation corresponds to?

c/planarityApp/planarityUtils.c
```diff
@@ -19,14 +19,21 @@ char Mode = 'r',
...
     Prompt("\nDo you want to \n"
            "  Randomly generate graphs (r),\n"
            "  Specify a graph (s),\n"
            "  Randomly generate a maximal planar graph (m), or\n"
-           "  Randomly generate a non-planar graph (n)? ");
-    scanf(" %c", &Mode);
+           "  Randomly generate a non-planar graph (n)?\n\t");
+    if (GetLineFromStdin(lineBuff, MAXLINE) != OK ||
+        strlen(lineBuff) != 1 ||
+        sscanf(lineBuff, " %c", &Mode) != 1)
```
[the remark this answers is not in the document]
wbkboyer: This is a great suggestion; I've gone ahead and done the same check for the `y`/`n`, `a`/`g` choice checks as well (edit: oops, you'd already suggested this below 😆 )! I have a couple follow-up questions:
1. I've created a macro `#define YESNOCHOICECHARS "yYnN"` in `planarity.h`; should I add a macro `#define RANDOMGRAPHSOUTPUTCHOICES "aAgG"`?
2. Should `GetSupportedOutputFormats()` and `GetAlgorithmChoices()` be converted to macros evaluating to string literals too, i.e., 
    ```
    #define TRANSFORMGRAPHOUTPUTCHOICES "gam"
    #define GRAPHALGORITHMCHOICES "pdo234"
    ```

c/graphLib/io/strOrFile.c
```diff
@@ -16,13 +16,22 @@ See the LICENSE.TXT file for licensing information.
...
  sf_New()
 
- Accepts an input string (do not pass pointer to output string; if you wish to
- create a strOrFile OUTPUT_CONTAINER that will be used to write to string,
- create it using sf_New(NULL, NULL, WRITETEXT)) XOR the name of a file filename.
-
- Note that when used as an input container, we're appending theStr to the
- internal strBuf; the original pointer is not manipulated, hence the const
- qualifiers.
+ The string-or-file object supports two IO modes: reading input (ioMode ==
```
john-boyer-phd: A model of commenting excellence! This is exactly what an API caller needs to know to be successful with this object.

c/planarityApp/planarityCommandLine.c
```diff
@@ -117,6 +120,10 @@ int legacyCommandLine(int argc, char *argv[])
             sprintf(messageContents, messageFormat, charsAvailForFilename, argv[1]);
 #pragma GCC diagnostic pop
             ErrorMessage(messageContents);
+
+            if (theGraph != NULL)
+                gp_Free(&theGraph);
```
john-boyer-phd: Good catch!

c/planarityApp/planarityCommandLine.c
```diff
@@ -462,77 +505,81 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int
...
 
-    // Change from internal OK/NONEMBEDDABLE/NOTOK result to a command-line style 0/-1 result
-    if (Result == OK || Result == NONEMBEDDABLE)
-        Result = 0;
-    else
+    if (Result != OK && Result != NONEMBEDDABLE)
     {
-        ErrorMessage("Test failed (graph processor returned failure result).\n");
-        Result = -1;
+        Message("Test failed (graph processor returned failure result).\n");
```
john-boyer-phd: I believe Result = NOTOK; should be added here, because it is what used to happen (except with Result = -1).

c/planarityApp/planarityCommandLine.c
```diff
@@ -462,77 +505,81 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int
...
 
+    char *inputString = NULL;
     char transformationCode = '\0';
+
     // runGraphTransformationTest will not test performing an algorithm on a given
     // input graph; it will only support "-(gam)"
     if (command == NULL || strlen(command) < 2)
     {
-        ErrorMessage("runGraphTransformationTest only supports -(gam).\n");
+        Message("runGraphTransformationTest only supports -(gam).\n");
```
john-boyer-phd: Same if returning NOTOK as assigning Result to NOTOK. Both should be preceded by ErrorMessage() rather than Message().

c/planarityApp/planarityCommandLine.c
```diff
@@ -578,43 +626,48 @@ int runGraphTransformationTest(char const *command, char const *infileName, int
...
+                if (expectedOutfileName != NULL)
+                {
+                    free(expectedOutfileName);
+                    expectedOutfileName = NULL;
+                }
 
-            if (actualOutput != NULL)
-                free(actualOutput);
+                if (actualOutput != NULL)
+                    free(actualOutput);
```
john-boyer-phd: Probably best, for maximum safety, to add a braced code block and assign actualOutput to NULL after the free() invocation.

c/planarityApp/planarityCommandLine.c
```diff
@@ -704,20 +758,22 @@ int callRandomMaxPlanarGraph(int argc, char *argv[])
...
         offset = 1;
     }
 
     if (argc > (5 + offset))
-        return -1;
+        return NOTOK;
 
     numVertices = atoi(argv[2 + offset]);
     outfileName = argv[3 + offset];
+
```
john-boyer-phd: Since we're in truly minor tweak land, might be best to use "-p" rather than just "p" below, bcz the mindset in this function is command-line args.

c/planarityApp/planarityMenu.c
```diff
@@ -9,288 +9,443 @@ See the LICENSE.TXT file for licensing information.
...
+                    continue;
+                }
+
+                if (command == 'p' || command == 'd' || command == 'o')
+                    secondOutfile = (char *)"";
 
-            Message("\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n");
-            FlushConsole(stdout);
+                if (!strchr(GetAlgorithmChoices(), command))
+                    Message("Invalid algorithm command choice, please retry.\n");
```
john-boyer-phd: ErrorMessage() seems appropriate.

c/planarityApp/planarityMenu.c
```diff
@@ -9,288 +9,443 @@ See the LICENSE.TXT file for licensing information.
...
-    } while (command != 'q');
+    }
 
     // Certain debuggers don't terminate correctly with pending output content
     FlushConsole(stdout);
     FlushConsole(stderr);
 
-    return 0;
+    if (choiceStringFormat != NULL)
+        free(choiceStringFormat);
```
john-boyer-phd: As a habit for max safety, please add a NULL assignment after every free() :-)

c/planarityApp/planarityMenu.c
```diff
@@ -9,288 +9,443 @@ See the LICENSE.TXT file for licensing information.
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+    sprintf(commandStringFormat, commandStringFormatFormat, COMMANDSTRINGMAXLENGTH);
+#pragma GCC diagnostic pop
 
     if (GetNumCharsToReprInt(FILENAMEMAXLENGTH, &numCharsToReprFILENAMEMAXLENGTH) != OK)
     {
         ErrorMessage("Unable to determine number of characters required to represent FILENAMEMAXLENGTH.\n");
-        return;
+        return NOTOK;
```
john-boyer-phd: Also free and NULL commandStringFormat in here, since returning.

c/planarityApp/planarityMenu.c
```diff
@@ -9,288 +9,443 @@ See the LICENSE.TXT file for licensing information.
...
+    }
 
-    Result = TestAllGraphs(commandString, infileName, outfileName, NULL);
-    if (Result != OK)
-        ErrorMessage("Algorithm test on all graphs in .g6 input file failed.\n");
+    if (Result == OK)
+        Result = TestAllGraphs(commandString, infileName, outfileName, NULL);
+
+    if (commandStringFormat != NULL)
+        free(commandStringFormat);
```
john-boyer-phd: NULL after free here and immediately below.

c/planarityApp/planaritySpecificGraph.c
```diff
@@ -66,26 +66,37 @@ int SpecificGraph(
...
             {
                 infileName = ConstructInputFilename(infileName);
-                if (infileName != NULL && strncmp(infileName, "stdin", strlen("stdin")) == 0)
+                if (infileName == NULL || strlen(infileName) == 0)
+                {
+                    ErrorMessage("Unrecoverable error while trying to ConstructInputFilename().\n");
+                    Result = NOTOK;
+                    break;
+                }
+                else if (strncmp(infileName, "stdin", strlen("stdin")) == 0)
```
john-boyer-phd: Prior bad act that may need to be fixed here while all the other cleansing is happening.
Since stdin is not accepted here, it would be better to emit a strong _error_ message then assign NOTOK to Result and break in this case, too. The issue is that I think SpecificGraph() can also be called from the command-line, so a menu-specific error seems wrong. So, if someone gives us "stdin" as the infilename, whether from menu mode or command line, just error and don't give them a chance to try again. This is especially important because command-line mode might be expected to be in quiet mode and not switching to a blocking input mode.

c/planarityApp/planarityUtils.c
```diff
@@ -17,82 +17,164 @@ char Mode = 'r',
...
+        if (GetLineFromStdin(lineBuff, MAXLINE) != OK)
+        {
+            ErrorMessage("Unable to fetch reconfigure choice from stdin.\n");
+            Result = NOTOK;
+            break;
+        }
+
+        if (strlen(lineBuff) != 1 ||
+            sscanf(lineBuff, " %c", &Mode) != 1 ||
+            !strchr("rsmn", tolower(Mode)))
```
john-boyer-phd: Since we're going super cleansed anyway, it's probably best to define a macro containing the legal mode characters, and use capital and small since you're not calling tolower() anyplace else. 

OR, my personal fave is to use small letters only and call tolower() here, but then you'd have to search everyplace else and add tolower() calls, then change all the other strings to small letters only. If you're OK with that, then, well, it's my fave so obviously OK with me too :-)

john-boyer-phd (changes requested):
Looking super-good, just a few more tweakies.

c/planarityApp/planarityCommandLine.c
```diff
@@ -462,77 +505,81 @@ int runSpecificGraphTest(char const *commandString, char const *infileName, int
...
+    else
     {
+        // Test that the primary actual output matches the primary expected output
         if (TextFileMatchesString(expectedPrimaryResultFileName, actualOutput) == TRUE)
             Message("Test succeeded (result equal to exemplar).\n");
         else
         {
-            ErrorMessage("Test failed (result not equal to exemplar).\n");
-            Result = -1;
+            Message("Test failed (result not equal to exemplar).\n");
```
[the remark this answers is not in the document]
wbkboyer: This is a *great* call-out! My original reason for changing all `ErrorMessage()` to `Message()` throughout the tests and menu-driven experience was that I thought if the user invokes with `-q`, then `ErrorMessage()` will not be emitted... However, I just tested this hypothetical situation, and *of course* in both of those cases, including `-q` is invalid! WHOOPS. I'll change all these back to `ErrorMessage()`!!

c/planarityApp/planarityCommandLine.c
```diff
@@ -623,7 +676,7 @@ int runGraphTransformationTest(char const *command, char const *infileName, int
     if (inputString != NULL)
         free(inputString);
 
-    return (Result == OK) ? 0 : -1;
+    return Result;
```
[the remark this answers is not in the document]
wbkboyer: I think that's a good call; even though it's highly unlikely that any code will be injected at some point in the future between the `free(inputString)` and the `return` statement, it's a good practice to `NULL`-out pointers we should no longer be able to use! :tada:

c/planarityApp/planaritySpecificGraph.c
wbkboyer: Turns out, the only way you could get to this error message is via the menu; the command line a) never gets here and b) is actually responsive to `stdin` (as long as you pipe the file in (i.e. `./planarity -s C stdin O < ./c/samples/K33.AdjList.txt`) or if you manually type/paste the input and then press Ctrl + D). As per meeting convo, we'll be leaving this menu-specific `ErrorMessage()` in!

PR #192, opened 2025-11-20
Enabling the certifying `K_{3,3}` search

wbkboyer:
As a follow-up to #186 , these updates ensure that the user may supply the command string `3e` to run certifying $K_{3,3}$ search.

## Updates

* `c/planarityApp/planarity.1`, `c/planarityApp/planarityHelp.c`, `c/planarityApp/planarityUtils.c` - Updated manpage and help messages so that they include `3e` command string option and description. Also, in `planarityUtils.c`'s `GetAlgorithmChoices()`, updated error handling.
* `c/graphLib/homeomorphSearch/graphK33Search_Extensions.c` - updated `_K33Search_EmbedPostprocess()` extension function so that when a graph is determined to be $K_{3,3}$-free (i.e. `edgeEmbeddingResult == OK`) using the $K_{3,3}$ search with embedder (i.e. when `EMBEDFLAGS_SEARCHWITHEMBEDDER` is set), we make sure to save the root of the embedding obstruction tree as well as the correspondence between original labelling and DFI ordering of vertices before performing the shared save-state steps (i.e. save the `embedFlags` as well as whether `FLAGS_ZEROBASEDIO` is set in `internalFlags` before blowing away the unnecessary leftovers in `theGraph` using `gp_ReinitializeGraph()`). Then, we must free the $K_{3,3}$ search `context`'s `associatedEONode` that was allocated during `_K33Search_ReinitializeGraph()` (i.e. $K_{3,3}$ search's extension that's called when we `gp_ReinitializeGraph()`) before reassigning the original root of the EO tree, and then must restore the DFI correspondence. This is all so that `_K33Search_CheckEmbeddingIntegrity()` (i.e. $K_{3,3}$ search override for `fpCheckEmbeddingIntegrity()`, called when we `gp_TestEmbedResultIntegrity()`) succeeds on the call to `_K33Search_ValidateEmbeddingObstructionTree()` (specifically the call `_K33Search_ValidateEmbeddingObstructionTreeEdgeSet()` requires the DFI label correspondence), and so that we keep the certificate of $K_{3,3}$-freeness, the EO tree, around for further processing or writing.
* `TestSupport/planaritytesting/` -
    * `planarity_testAllGraphs_orchestrator.py`, `planaritytesting_utils.py`, and `planarity_testAllGraphs_output_parsing.py` updated to be responsive to user specifying command string `3e`. Also, now users can specify which graph algorithm extensions they wish to `TestAllGraphs()` using the `-c` flag (rather than manually editing the tuple returned by `PLANARITY_ALGORITHM_SPECIFIERS()`, lol)

### While I was in the neighbourhood
* `c/planarityApp/planarityCommandLine.c` - when `-q` supplied to Debug build (i.e. built using `(/|-)DDEBUG` flag), executable no longer waits for user confirmation before terminating.

### DIFF
--- a/c/graphLib/homeomorphSearch/graphK33Search_embedder.c
+++ b/c/graphLib/homeomorphSearch/graphK33Search_embedder.c
@@ -251,8 +251,8 @@ int _K33Search_EONode_NewONode(graphP theGraph, K33Search_EONodeP *pNewONode)
                 gp_SetEdgeVirtual(theNewK5Graph, gp_GetTwinArc(theNewK5Graph, e));
             }
 
-    // Now we can contruct an embedding obstruction tree node to associate with theNewK5Graph
-    // and then tell it to be an O-node that owns theNewK5Graph.
+    // Now we can construct an embedding obstruction tree node to associate with
+    // theNewK5Graph and then tell it to be an O-node that owns theNewK5Graph.
     if ((theNewONode = _K33Search_EONode_New(K33SEARCH_EOTYPE_ONODE, theNewK5Graph, TRUE)) == NULL)
     {
         gp_Free(&theNewK5Graph);
@@ -669,8 +669,9 @@ int _K33Search_MarkBridgeSetToExtract(graphP theGraph, int R, int cutv1, int cut
         // Push a next vertex to visit
         sp_Push(theGraph->theStack, gp_GetNeighbor(theGraph, e));
 
-        // Go to the next edge in the rotation, but keep record of the edge just processed
-        // so we can deetect when we have finsished processing the lastStartingEdge
+        // Go to the next edge in the rotation, but keep record of the edge just
+        // processed so we can deetect when we have finished processing the
+        // lastStartingEdge
         ePrev = e;
         e = theGraph->E[e].link[linkDir];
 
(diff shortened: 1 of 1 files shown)


c/planarityApp/planarityUtils.c
```diff
@@ -630,11 +632,21 @@ int GetEmbedFlags(char command, char modifier, int *pEmbedFlags)
 
     if (modifier != '\0')
     {
-        if (command == '3' && modifier == 'e')
-            (*pEmbedFlags) |= EMBEDFLAGS_SEARCHWITHEMBEDDER;
+        if (command == '3')
```
john-boyer-phd: Nice change bcz better error message in -3* case

TestSupport/planaritytesting/planarity_testAllGraphs_orchestrator.py
```diff
@@ -30,16 +30,18 @@
...
 from pathlib import Path
+import re
+import subprocess
+import sys
 from typing import Optional
 
 
 from planaritytesting_utils import (
     PLANARITY_ALGORITHM_SPECIFIERS,
+    PLANARITY_ALLOWED_COMMAND_STRINGS,
```
john-boyer-phd: Good job baking 3e into the test all graphs orchestrator level.

john-boyer-phd (approved):
Excellent, great to have run-time activation of the embedder on the branch.

PR #193, opened 2025-12-03
Issue #191 - Ensure after `free()`ing memory, immediately set the pointer to `NULL`

### LINKED ISSUE #191: Ensure after `free()`ing memory, immediately set the pointer to `NULL`
During the course of #186, I ended up touching a lot of the error handling code in the `planarityApp` layer. One thing I tried to make consistent was how we free resources at the end of processing (either when we hit an error state or at the end of the function's scope), i.e., if a pointer is not `NULL`, then free the resource to which it points and immediately set the pointer to `NULL`:

```
if (thing != NULL)
{
    free(&thing);
    thing = NULL;
{
```

This guards against future additions to the code (e.g. changes to error-handling flow) where we might erroneously try to do something with memory that's already been `free()`d: since we often (and should in fact endeavour to always) check that a pointer is not `NULL` before trying to do something with it, setting the pointer to `NULL` after `free()` will mean we can prevent memory access errors.

wbkboyer:
Resolves #191 

Initial audit of instances where we `free()`, making sure to set to `NULL`.

As well, updated error checking and handling, and surfaced some issues with `gp_ReadFromString()` handling `inputStr` cleanup and overzealous cleanup when we hit an error state in `_encodeAdjMatAsG6()` (called by `writeGraphUsingG6WriteIterator()`)

## Updates
* `c/graphLib/graphUtils.c` - in `gp_DupGraph()`, initialize local `graphP result` to `NULL`.
* `c/graphLib/homeomorphSearch/`
    * `graphK23Search_Extensions.c` - updated `gp_AttachK23Search()` so that in the case where we fail to `gp_AddExtension()`, after `_K23Search_FreeContext(context)`, we set the local `context` pointer to `NULL`.
    * `graphK33Search_Extensions.c` - updated `gp_AttachK33Search()` and `_K33Search_DupContext()` so in error cases, we set the local `context`/`newContext` pointer to `NULL` after we call `_K33Search_FreeContext()`.
    * `graphK4Search_Extensions.c` - updated `gp_AttachK4Search()` and `_K4Search_DupContext()` so in error cases, we set the local `context`/`newContext` pointer to `NULL` after we call `_K4Search_FreeContext()`.
* `c/graphLib/io/`
    * `g6-read-iterator.c` - in `_beginG6ReadIteration()`, initialize `graphOrder` to `NIL` rather than `-1` (technically speaking, should probably be initialized to 0, but both are invalid values for the graph order; the `graphOrder` is updated after successfully reading graph order from `.g6` input file with `_getGraphOrder()`)
    * `g6-write-iterator.c` - `allocateG6WriteIterator()` now checks that the `pGraph` is valid *before* allocating memory for the `G6WriteIterator` so that we can return early and don't have to worry about `freeG6WriteIterator()` in error state. `_encodeAdjMatAsG6()` now does not free memory it does not own in an error state; the memory that local pointers `columnOffsets` and `g6Encoding` point to in the `pG6WriteIterator` should be freed by error handling up the callstack in the function that "owns" the `G6WriteIterator`
    * `graphIO.c` - Making sure local variables are initialized with specific invalid values that should be updated to valid values during processing. Updated docstring of `gp_ReadFromString()` to indicate that the *caller* owns the memory for `inputStr`, the contents of which are only copied into the internal `strBuf` of the `inputContainer`. Found two instances of where we failed to `NULL` out a local variable after `free()` (`_WriteAdjMatrix()` and `_WriteGraph()`). Also note that `sf_Free()` (like `gp_Free()`, `LCFree()`, `sp_Free()`, etc) will make sure the pointer-pointer (pointer to `strOrFileP`) and pointer (dereference to get `strOrFileP`) are not `NULL` before freeing memory associated with the `strOrFile` and finally setting the pointer to `NULL`; this means I was able to clean up `gp_Write()` and `gp_WriteToString()` slightly.
* `c/graphLib/planarityRelated/`
     * `graphDrawPlanar.c` - found a few instances where I needed to set the local pointer to `NULL` after freeing a resource. More importantly, in `gp_DrawPlanar_RenderToFile()`, changed where we determine the `Result` so that after writing the result of `_RenderToString()` to file, we can `NULL` the pointer `theRendition` after freeing it.
     * `graphDrawPlanar_Extensions.c` - `_DrawPlanar_WritePostprocess()` has been updated so that we disallow stacking `_WritePostprocess()` calls from multiple extensions
         * __**N.B.**__ This decision was made because I discovered a potential memory leak: if in the future we implement an override for `_WritePostprocess()` for some other extension, then attach both `DrawPlanar` and that extension, the original implementation would erroneously overwrite the pointer to any `extraData` that was generated by the other extension. After discussion with @john-boyer-phd , an issue needs to be written to refactor this whole approach 
* `c/planarityApp/`
    * `planarityCommandLine.c` - updated error-handling in `legacyCommandLine()`. Added initial values for local variables. Fixed comments in `runSpecificGraphTest()` and `runGraphTransformationTest()` about how those functions own `inputString` when `inputInMemFlag` is set, and therefore must clean up (`free()` and set local pointer to `NULL`)
    * `planarityMenu.c` - only added whitespace; verified there were no instances where we failed to `NULL` pointers to memory that had been `free()`d
    * `planaritySpecificGraph.c` - in `SpecificGraph()`, added a lot of whitespace for readability, added additional error-handling in the case that we fail to `gp_New()` (catch early before `gp_Embed()`), made sure to free both `theGraph` and `origGraph` when appropriate in error states, added `ErrorMessage()` for when we fail to write to primary or secondary output files.
    * `planarityRandomGraphs.c` - initializing local variables and ensuring we `gp_Free()` `theGraph` and `origGraph` in case of error. In `RandomGraphs()`, got rid of `writeErrorReported_(Embedded|Obstructed|Random|AdjList|Error)` since we now bail out of the loop as soon as we encounter a write error. Also changed the order of some of the checks in the loop so that we don't even try to `ReinitializeGraph()` if there's been a write error, and don't bother to write the statistics unless the `Result` was `OK`/`NONEMBEDDABLE`.

### DIFF
--- a/c/graphLib/io/g6-write-iterator.c
+++ b/c/graphLib/io/g6-write-iterator.c
@@ -12,11 +12,16 @@ See the LICENSE.TXT file for licensing information.
 
 int allocateG6WriteIterator(G6WriteIteratorP *ppG6WriteIterator, graphP pGraph)
 {
-    int exitCode = OK;
-
     if (ppG6WriteIterator != NULL && (*ppG6WriteIterator) != NULL)
     {
         ErrorMessage("G6WriteIterator is not NULL and therefore can't be allocated.\n");
+        return NOTOK;
+    }
+
+    if (pGraph == NULL || gp_getN(pGraph) <= 0)
+    {
+        ErrorMessage("Must allocate and initialize graph with an order greater than 0 to use the G6WriteIterator.\n");
+
         return NOTOK;
     }
 
@@ -33,20 +38,9 @@ int allocateG6WriteIterator(G6WriteIteratorP *ppG6WriteIterator, graphP pGraph)
     (*ppG6WriteIterator)->g6Output = NULL;
     (*ppG6WriteIterator)->currGraphBuff = NULL;
     (*ppG6WriteIterator)->columnOffsets = NULL;
+    (*ppG6WriteIterator)->currGraph = pGraph;
 
-    if (pGraph == NULL || gp_getN(pGraph) <= 0)
-    {
-        ErrorMessage("[ERROR] Must allocate and initialize graph with an order greater than 0 to use the G6WriteIterator.\n");
-
-        if (freeG6WriteIterator(ppG6WriteIterator) != OK)
-            ErrorMessage("Unable to free the G6WriteIterator.\n");
-
-        exitCode = NOTOK;
-    }
-    else
-        (*ppG6WriteIterator)->currGraph = pGraph;
-
-    return exitCode;
+    return OK;
 }
 
 bool _isG6WriteIteratorAllocated(G6WriteIteratorP pG6WriteIterator)
@@ -278,11 +272,11 @@ int _encodeAdjMatAsG6(G6WriteIteratorP pG6WriteIterator)
 
     if (graphOrder > 62)
     {
-        int i, intermediate;
-        g6Encoding[0] = 126;
         // bytes 1 through 3 will be populated with the 18-bit representation of the graph order
-        intermediate = -1;
-        for (i = 0; i < 3; i++)
+        int intermediate = -1;
+        g6Encoding[0] = 126;
+
+        for (int i = 0; i < 3; i++)
         {
             intermediate = graphOrder >> (6 * i);
             g6Encoding[3 - i] = intermediate & 63;
@@ -339,22 +333,6 @@ int _encodeAdjMatAsG6(G6WriteIteratorP pG6WriteIterator)
         if (exitCode != OK)
         {
             ErrorMessage("Unable to fetch next edge in graph.\n");
-
-            // FIXME: I don't think I should be doing this here, as these belong
-            // to the G6WriteIterator and should only be cleaned up when the
-            // caller decides what to do in an error state, i.e to free the
-            // iterator using freeG6WriteIterator()... RandomGraphs() doesn't
-            // stop when there's an error from writeGraphUsingG6WriteIterator(),
-            // so this means for every subsequent iteration up to K graphs
-            // generated, _isG6WriteIteratorAllocated() will return NOTOK (so at
-            // least we're not trying to access data members, but there might be
-            // other places *not* appropriately safeguarded)
-
-            // free(columnOffsets);
-            // columnOffsets = NULL;
-
-            // free(g6Encoding);
-            // g6Encoding = NULL;
 
             return exitCode;
         }
--- a/c/graphLib/io/graphIO.c
+++ b/c/graphLib/io/graphIO.c
@@ -55,8 +55,8 @@ char _GetVertexObstructionTypeChar(graphP theGraph, int v);
 int _ReadAdjMatrix(graphP theGraph, strOrFileP inputContainer)
 
 {
-    int N = -1;
-    int v, w, Flag;
+    int N = 0;
+    int v = NIL, w = NIL, Flag = NIL;
 
     if (sf_ValidateStrOrFile(inputContainer) != OK)
         return NOTOK;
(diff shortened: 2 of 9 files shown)


c/graphLib/io/g6-write-iterator.c
```diff
@@ -38,10 +38,10 @@ int allocateG6WriteIterator(G6WriteIteratorP *ppG6WriteIterator, graphP pGraph)
     {
         ErrorMessage("[ERROR] Must allocate and initialize graph with an order greater than 0 to use the G6WriteIterator.\n");
 
-        exitCode = freeG6WriteIterator(ppG6WriteIterator);
-
-        if (exitCode != OK)
+        if (freeG6WriteIterator(ppG6WriteIterator) != OK)
             ErrorMessage("Unable to free the G6WriteIterator.\n");
+
+        exitCode = NOTOK;
```
wbkboyer: The original formulation would have caused issues, because if `freeG6WriteIterator()` succeeds, we'd be erroneously returning `OK`; regardless of whether or not `freeG6WriteIterator()` succeeds, if `pGraph` is not properly initialized, we need `exitCode` to be `NOTOK`.

Now that I think about it, we should `return NOTOK;` here; previously, the `exitCode` being incorrectly set didn't matter because of the subsequent check of `_isG6WriteIteratorAllocated()`.

c/graphLib/io/g6-write-iterator.c
```diff
@@ -339,8 +339,23 @@ int _encodeAdjMatAsG6(G6WriteIteratorP pG6WriteIterator)
         if (exitCode != OK)
         {
             ErrorMessage("Unable to fetch next edge in graph.\n");
-            free(columnOffsets);
-            free(g6Encoding);
+
+            // FIXME: I don't think I should be doing this here, as these belong
```
john-boyer-phd: I agree, definitely don't free those because they were not malloc'd here. g6Encoding and columnOffsets are local pointers that point to but don't own blocks of memory that were allocated by and are owned by the iterator data structure. They can only be freed when free iterator is called.

c/graphLib/io/graphIO.c
```diff
@@ -450,9 +450,29 @@ int gp_ReadFromString(graphP theGraph, char *inputStr)
     strOrFileP inputContainer = sf_New(inputStr, NULL, READTEXT);
     if (inputContainer == NULL)
     {
+        // FIXME: Should this actually be moved *above* this check, seeing as we
```
john-boyer-phd: The ownership was changed prior to the 4.0 release (on Feb. 26) such that gp_ReadFromString() no longer owned the string, due to use of the spiffy new string-or-file object, which takes a copy instead. The comment indicating that gp_ReadFromString() was added in the same commit, but was evidently added to better explain how it was working before we decided to bite the bullet and do string-or-file. It's also unsafe for gp_ReadFromString() to free memory it didn't allocate, so we definitely want to keep it that way.

So, the function comment above should be changed to something like "The caller owns the memory of inputStr". Also, the free() call in the error case here should be deleted so that we just return NOTOK if the inputContainer was not successfully allocated.

There are two other places in the code where you did already fix the fact that the caller must free the inputStr that they allocated, so no memory leaks related to this in the 4.0 release, but a couple of comments still say the old thing about passing ownership. Those comments should be similarly amended. Both are in planarityCommandLine.c. The first is in runSpecificGraphTest() at Lines 504-505, and the second is in runGraphTransformationTest() at Line 624-625.

c/graphLib/io/graphIO.c
```diff
@@ -130,7 +130,7 @@ int _ReadAdjMatrix(graphP theGraph, strOrFileP inputContainer)
 int _ReadAdjList(graphP theGraph, strOrFileP inputContainer)
 {
     int N = -1;
-    int v, W, adjList, e, indexValue, ErrorCode;
+    int v = -1, W = -1, adjList = -1, e = -1, indexValue = -1, ErrorCode = OK;
```
john-boyer-phd: In this and all subsequent locations where -1 has been used as an initializer, it is usually if not always better to use NIL instead of -1.  This includes vertices with names like u, v, w, x, and y, and edges with names like e and eTwin. Looking down into the code, indexValue refers to a vertex, so NIL again.

c/planarityApp/planaritySpecificGraph.c
```diff
@@ -62,7 +63,11 @@ int SpecificGraph(
         if (infileName != NULL)
         {
             if ((infileName = ConstructInputFilename(infileName)) == NULL)
-                return NOTOK;
+            {
+                ErrorMessage("Unrecoverable error while trying to ConstructInputFilename().\n");
```
john-boyer-phd: Addition of error message good! How about "Error constructing input filename."

PR #206, opened 2026-04-24
Issue # 205 - Refactor `testAllGraphs()` flow for ensuring edge capacity and extension of graph to embed 

### LINKED ISSUE #205: Refactor `testAllGraphs()` flow for ensuring edge capacity and extension of graph to embed
## Background

In `testAllGraphs()`, we currently `gp_EnsureEdgeCapacity()` of the graph associated with the `G6ReadIterator` before we `ExtendGraph()` with the graph algorithm extension associated with the `command`, then in the loop we `g6_ReadGraph()` into `theGraph` before we `gp_CopyGraph()` `theGraph` into `copyOfOrigGraph` (the graph used for the embedding integrity check), and finally `gp_Embed()` `theGraph`.

Although this works, we should be using the graph associated with the `G6ReadIterator` as the graph against which we perform the embed integrity check, and perform the embedding on the copy of that graph.

In the current implementation, if we *didn't* ensure edge capacity of `theGraph` before the loop in which we `g6_ReadGraph()`, then the `gp_DynamicAddEdge()` call would fail for `d34` for `M > DEFAULT_EDGE_LIMIT * N` (since those extensions' overrides currently return `NOTOK`; this will be remediated by a follow-up issue to actually resize the structures associated with those extensions).

## Proposed solution

- [x] Rename `theGraph` to `origGraphRead`, which is what gets passed to the `G6ReadIterator` and against which we will perform integrity checks after embedding the *copy* of this graph
- [x] Rename `copyOfOrigGraph` to `graphForEmbedding`
- [ ] Outside the loop, ensure edge capacity of `graphForEmbedding` to the max edge-count (for ALL extensions, not just `d34`, as well as the `origGraphRead` (otherwise the `gp_CopyGraph()` will fail due to `graphForEmbedding->edgeCapacity > origGraphRead->edgeCapacity` on line 1032 of `graphUtils.c`)
- [ ] Outside the loop, extend `origGraphRead` with the graph algorithm extension associated with the `command` (since every loop iteration, the `gp_CopyGraph()` will call `gp_CopyExtensions()`, which will free the extensions on the `graphForEmbedding` and then allocate new memory to copy the extensions from the `origGraphRead` to the `graphForEmbedding`)

## Related issues and PRs

* #64 
* #67 
* #68

wbkboyer:
Resolves #205 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Added

* N/A

### Updated

* `c/planarityApp/planarityTestAllGraphs.c` - renamed `theGraph` to `origGraphRead` and `copyOfOrigGraph` to `graphForEmbedding`, then changed the flow so that now, the graph associated with the reader (`origGraphRead`) is that which is used for the embedding integrity check, and that we embed the copy of the graph (`graphForEmbedding`). We must `gp_EnsureEdgeCapacity()` of both the `origGraphRead` and `graphForEmbedding`, because if you try to `gp_EnsureEdgeCapacity()` of a graph after drawplanar, $K_{3,3}$ search, or $K_4$ search are attached, those extensions' overrides currently return `NOTOK`, and if the source graph (`origGraphRead`)'s edge capacity is less than the destination graph (`graphForEmbedding`), `gp_CopyGraph()` fails.

Sweeping changes to address the comment about getting rid of a lot of those pesky
```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wformat-nonliteral"
        ...
#pragma GCC diagnostic pop
```

* `c/graphLib/io/g6-api-utilities.c` - updated `ErrorMessage()` so that I pass a string literal and arguments to the updated `ErrorMessage()` function
* `c/graphLib/io/g6-read-iterator.c` - updated `ErrorMessage()` so that I pass a string literal and arguments to the updated `ErrorMessage()` function, and was able to get rid of `messageContents` and `charsAvailForStr` (typically use `FILENAME_MAX` from `stdio.h` when I need to limit the output for how long a filename should be)
* `c/graphLib/lowLevelUtils/apiutils.c`/`.h` - `Message()` and `ErrorMessage()` are now variadic functions, so that one may provide a format string and all arguments to populate that string
* `c/planarityApp/planarity.h` - removed `Prompt()`, since it was just a `Message()` with a `FlushConsole()` (which was already done in `Message()`
* `c/planarityApp/planarityCommandLine.c` - removed need for `messageFormat`, `messageContents`, `charsAvailForXXX` due to refactor with new variadic `Message()` and `ErrorMessage()`
* `c/planarityApp/planarityHelp.c` - To get around the compiler complaining that the string returned by `GetProjectTitle()`, `GetAlgorithmFlags()`, `GetAlgorithmSpecifiers()`, `GetSupportedOutputChoices()` being nonliteral, I created a format string that will just be populated with whatever is returned from these functions. Any `Prompt()` is now a `Message()`.
* `c/planarityApp/planarityCommandLine.c` - removed need for `messageFormat`, `messageContents`, `charsAvailForXXX` due to refactor with new variadic `Message()` and `ErrorMessage()`
* `c/planarityApp/planarityHelp.c` - To get around the compiler complaining that the string returned by `GetProjectTitle()`, I created a format string that will just be populated with whatever is returned from these functions.
* `c/planarityApp/planarityMenu.c` - To get around the compiler complaining that the string returned by `GetProjectTitle()`, `GetAlgorithmSpecifiers()`, and `GetSupportedOutputChoices()`, I created a format string that will just be populated with whatever is returned from these functions. All `Prompt()` are now `Message()`
* `c/planarityApp/planarityRandomGraphs.c` - removed need for `messageFormat`, `messageContents`, `charsAvailForXXX` due to refactor with new variadic `Message()` and `ErrorMessage()`
* `c/planarityApp/planarityTestAllGraphs.c` - In addition to the above changes, removed need for `messageFormat`, `messageContents`, `charsAvailForXXX` due to refactor with new variadic `Message()` and `ErrorMessage()`
* `c/planarityApp/planarityUtils.c` - All `Prompt()` are now `Message()`, removed `Prompt()` definition, and removed need for `messageFormat`, `messageContents`, `charsAvailForXXX` due to refactor with new variadic `Message()` and `ErrorMessage()`

### Removed

* N/A

## Testing

- [x] Re-ran `planarity_testAllGraphs_orchestrator.py` for `N=10` of VSCode Release build (1-based arrays) and verified that the `OK`/`NONEMBEDDABLE` counts did not change (observed a 1-2% slowdown, but without testing higher graph orders, this seems within a margin of error) 
- [ ] Upload logs or provide relevant snippets of terminal output to demo functionality

<details><summary>Replace with high-level description of log snippets</summary>

```
Paste terminal log snippets here
```

<details>

### DIFF
--- a/c/planarityApp/planarityTestAllGraphs.c
+++ b/c/planarityApp/planarityTestAllGraphs.c
@@ -113,8 +113,9 @@ int TestAllGraphs(char const *const commandString, char const *const infileName,
 int testAllGraphs(char command, char modifier, char const *const infileName, testAllStatsP stats)
 {
     int Result = OK;
-    graphP theGraph = NULL;
-    graphP copyOfOrigGraph = NULL;
+
+    graphP origGraphRead = NULL;
+    graphP graphForEmbedding = NULL;
     int embedFlags = 0, numOK = 0, numNONEMBEDDABLE = 0, errorFlag = FALSE;
 
     G6ReadIteratorP theG6ReadIterator = NULL;
@@ -132,9 +133,9 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         return Result;
     }
 
-    theGraph = gp_New();
+    origGraphRead = gp_New();
 
-    if (theGraph == NULL)
+    if (origGraphRead == NULL)
     {
         ErrorMessage("Unable to allocate graph.\n");
 
@@ -143,11 +144,11 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         return NOTOK;
     }
 
-    if ((Result = g6_NewReader((&theG6ReadIterator), theGraph)) != OK)
+    if ((Result = g6_NewReader((&theG6ReadIterator), origGraphRead)) != OK)
     {
         ErrorMessage("Unable to allocate G6ReadIterator.\n");
 
-        gp_Free(&theGraph);
+        gp_Free(&origGraphRead);
         stats->errorFlag = TRUE;
 
         return Result;
@@ -159,38 +160,77 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
                      "G6ReadIterator.\n");
 
         g6_FreeReader((&theG6ReadIterator));
-        gp_Free(&theGraph);
+        gp_Free(&origGraphRead);
         stats->errorFlag = TRUE;
 
         return Result;
     }
 
-    order = gp_GetN(theGraph);
+    order = gp_GetN(origGraphRead);
+
+    graphForEmbedding = gp_New();
+    if (graphForEmbedding == NULL)
+    {
+        ErrorMessage("Unable to allocate graph to store copy of original graph before embedding.\n");
+
+        g6_FreeReader((&theG6ReadIterator));
+        gp_Free(&origGraphRead);
+        stats->errorFlag = TRUE;
+
+        return NOTOK;
+    }
+
+    if (gp_InitGraph(graphForEmbedding, order) != OK)
+    {
+        ErrorMessage("Unable to initialize graph datastructure to store copy of original graph before embedding.\n");
+
+        g6_FreeReader((&theG6ReadIterator));
+        gp_Free(&origGraphRead);
+        gp_Free(&graphForEmbedding);
+        stats->errorFlag = TRUE;
+
+        return Result;
+    }
+
     // We have to set the maximum edge capacity (i.e. (N * (N - 1) / 2)) because some of the test files
     // can contain complete graphs, and the graph drawing, K_{3, 3} search, and K_4 search extensions
     // don't support expanding the edge capacity after being attached.
-    if (strchr("d34", command) != NULL)
+    if ((Result = gp_EnsureEdgeCapacity(graphForEmbedding, (order * (order - 1) / 2))) != OK)
     {
-        if ((Result = gp_EnsureEdgeCapacity(theGraph, (order * (order - 1) / 2))) != OK)
-        {
-            ErrorMessage("Unable to ensure sufficient edge capacity of the G6ReadIterator's graph struct.\n");
+        ErrorMessage("Unable to ensure sufficient edge capacity of the graph for embedding.\n");
 
-            g6_FreeReader((&theG6ReadIterator));
-            gp_Free(&theGraph);
-            stats->errorFlag = TRUE;
+        g6_FreeReader((&theG6ReadIterator));
+        gp_Free(&origGraphRead);
+        gp_Free(&graphForEmbedding);
+        stats->errorFlag = TRUE;
 
-            return Result;
-        }
+        return Result;
+    }
+    // We have to set the maximum edge capacity because otherwise gp_CopyGraph()
+    // will fail due to the destination graph (graphForEmbedding) having a greater
+    // edge capacity than the source graph (origGraphRead)
+    if ((Result = gp_EnsureEdgeCapacity(origGraphRead, (order * (order - 1) / 2))) != OK)
+    {
+        ErrorMessage("Unable to ensure sufficient edge capacity of the original graph read.\n");
+
+        g6_FreeReader((&theG6ReadIterator));
+        gp_Free(&origGraphRead);
+        gp_Free(&graphForEmbedding);
+        stats->errorFlag = TRUE;
+
+        return Result;
     }
 
-    if ((Result = ExtendGraph(theGraph, command)) != OK)
+    // We must extend the original graph so that in the loop body when we
+    // gp_CopyGraph(), the extension structures will be copied over.
+    if ((Result = ExtendGraph(origGraphRead, command)) != OK)
     {
         char commandStr[3];
         commandStr[0] = command;
         commandStr[1] = modifier;
         commandStr[2] = '\0';
 
-        messageFormat = "Unable to extend graph with command %s\n";
+        messageFormat = "Unable to extend graph for embedding with command %s\n";
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wformat-nonliteral"
         sprintf(messageContents, messageFormat, commandStr);
@@ -199,31 +239,8 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         ErrorMessage(messageContents);
 
         g6_FreeReader(&theG6ReadIterator);
-        gp_Free(&theGraph);
-        stats->errorFlag = TRUE;
-
-        return Result;
-    }
-
-    copyOfOrigGraph = gp_New();
-    if (copyOfOrigGraph == NULL)
-    {
-        ErrorMessage("Unable to allocate graph to store copy of original graph before embedding.\n");
-
-        g6_FreeReader((&theG6ReadIterator));
-        gp_Free(&theGraph);
-        stats->errorFlag = TRUE;
-
-        return NOTOK;
-    }
-
-    if (gp_InitGraph(copyOfOrigGraph, order) != OK)
-    {
-        ErrorMessage("Unable to initialize graph datastructure to store copy of original graph before embedding.\n");
-
-        g6_FreeReader((&theG6ReadIterator));
-        gp_Free(&theGraph);
-        gp_Free(&copyOfOrigGraph);
+        gp_Free(&origGraphRead);
+        gp_Free(&graphForEmbedding);
         stats->errorFlag = TRUE;
 
         return Result;
@@ -248,9 +265,13 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         if (g6_EndReached(theG6ReadIterator))
             break;
 
-        gp_CopyGraph(copyOfOrigGraph, theGraph);
+        // NOTE: This will gp_CopyExtensions() from the origGraphRead to the
+        // graphForEmbedding, which will free any existing extensions on the
+        // graphForEmbedding before malloc and populating with the extensions
+        // from the origGraphRead
+        gp_CopyGraph(graphForEmbedding, origGraphRead);
 
-        Result = gp_Embed(theGraph, embedFlags);
+        Result = gp_Embed(graphForEmbedding, embedFlags);
         if (Result != OK && Result != NONEMBEDDABLE)
         {
             messageFormat = "Failed to embed graph on line %d for command '%c'.\n";
@@ -265,7 +286,7 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
             break;
         }
 
-        if (gp_TestEmbedResultIntegrity(theGraph, copyOfOrigGraph, Result) != Result)
+        if (gp_TestEmbedResultIntegrity(graphForEmbedding, origGraphRead, Result) != Result)
         {
             messageFormat = "Embed integrity check failed for graph on line %d for command '%c'.\n";
 #pragma GCC diagnostic push
@@ -316,8 +337,8 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
     stats->errorFlag = errorFlag;
 
     g6_FreeReader((&theG6ReadIterator));
-    gp_Free(&theGraph);
-    gp_Free(&copyOfOrigGraph);
+    gp_Free(&origGraphRead);
+    gp_Free(&graphForEmbedding);
 
     return Result;
 }
--- a/c/planarityApp/planarityTestAllGraphs.c
+++ b/c/planarityApp/planarityTestAllGraphs.c
@@ -47,15 +47,14 @@ int TestAllGraphs(char const *const commandString, char const *const infileName,
 
     if (GetCommandAndOptionalModifier(commandString, &command, &modifier) != OK)
     {
-        ErrorMessage("Unable to determine command (and optional modifier) from command string.\n");
-
+        ErrorMessage("Unable to determine command (and optional modifier) from "
+                     "command string.\n");
         return NOTOK;
     }
 
     if (infileName == NULL)
     {
         ErrorMessage("No input file provided.\n");
-
         return NOTOK;
     }
 
@@ -85,7 +84,6 @@ int TestAllGraphs(char const *const commandString, char const *const infileName,
         sprintf(messageContents, messageFormat, command, charsAvailForFilename, infileName);
 #pragma GCC diagnostic pop
         ErrorMessage(messageContents);
-
         Result = NOTOK;
     }
     else
@@ -103,7 +101,6 @@ int TestAllGraphs(char const *const commandString, char const *const infileName,
         sprintf(messageContents, messageFormat, command, charsAvailForFilename, infileName);
 #pragma GCC diagnostic pop
         ErrorMessage(messageContents);
-
         Result = NOTOK;
     }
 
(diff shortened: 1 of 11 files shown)


c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
-        ErrorMessage(messageContents);
+        return Result;
+    }
 
-        g6_FreeReader(&theG6ReadIterator);
-        gp_Free(&theGraph);
+    // We have to set the maximum edge capacity (i.e. (N * (N - 1) / 2)) because some of the test files
+    // can contain complete graphs, and the graph drawing, K_{3, 3} search, and K_4 search extensions
+    // don't support expanding the edge capacity after being attached.
+    if ((Result = gp_EnsureEdgeCapacity(graphForEmbedding, (order * (order - 1) / 2))) != OK)
```
john-boyer-phd: It'd be good to put both EnsureCapacity calls here, then just have a more generic error message like "Unable to ensure sufficient edge capacity for requested graph embedding operation". Later, we will be able to remove this if block, once K3,3, K4, and drawing have ensure edge capacity function implementations.

Also, remove the assignment to Result.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
         stats->errorFlag = TRUE;
 
-        return NOTOK;
+        return Result;
     }
 
-    if (gp_InitGraph(copyOfOrigGraph, order) != OK)
+    // We must extend the original graph so that in the loop body when we
+    // gp_CopyGraph(), the extension structures will be copied over.
+    if ((Result = ExtendGraph(origGraphRead, command)) != OK)
```
john-boyer-phd: It would be better to extend both graphs here, to show app developers how we would like it to be done. Since gp_Embed() is called on graphForEmbedding, it makes the most sense to extend that graph so that it is of the subclass that supports the algorithm gp_Embed() will be asked to run.

It is true that the current gp_CopyGraph() implementation wipes out extensions for graphForEmbedding and then copies over the ones from origGraphRead, but this is suboptimal, and will be fixed soon. When that happens, it will have to be graphForEmbedding that needs to be extended (it would be optional to also extend origGraphRead, and in fact preferable not to, once gp_CopyGraph() has a better implementation).

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
     {
-        ErrorMessage("Unable to initialize graph datastructure to store copy of original graph before embedding.\n");
+        char commandStr[3];
```
john-boyer-phd: Let's get rid of commandStr and just use a static error message like "Unable to extend graph to support requested graph embedding operation." It will tighten up this error code and also eliminate the ugly pragma we are forced to use by the beyond pedantic warnings as errors.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
 
         return Result;
     }
 
-    order = gp_GetN(theGraph);
-    // We have to set the maximum edge capacity (i.e. (N * (N - 1) / 2)) because some of the test files
-    // can contain complete graphs, and the graph drawing, K_{3, 3} search, and K_4 search extensions
-    // don't support expanding the edge capacity after being attached.
-    if (strchr("d34", command) != NULL)
+    order = gp_GetN(origGraphRead);
```
john-boyer-phd, line 132: Let's please put a comment here that says something like "The order of the graphs in the G6 source file or string was determined by g6_NewReader() and we obtain it to initialize the graph for embedding".
Then, it might be better to move this down so that it comes right before the gp_InitGraph(), rather than being separated from it by the gp_New() call.
Edit: Once gp_InitGraph call is moved up to be with gp_New in the same if stmt, then this can stay here. See next comment.

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: This annoying pragma that we are forced to use is so disruptive to the flow of the loop (because the code formatter forces it to be on the left margin). To remediate, please refactor this error code to emit a first error message that has no linefeed, like "Unable to read graph from G6 source on line ". Then, create `char lineNumber[MAXCHARSFOR32BITINT + 3]` just below curly brace (to avoid another warning as error), then `sprintf("%d\n", theG6ReadIterator->numGraphsRead+1)`, and emit a second `ErrorMessage(lineNumber)`. 

Emitting two error messages isn't perfect, but perfection is the enemy of excellence, and the pragmas are the enemy.

P.S. I know I am the one who originally added these because the beyond-pedantic-warnings-as-errors were whining about a perfectly legit thing to do in C, but at some point we're going to have to go on a tear of seeing how many of those pragmas could be replaced by techniques like the above (rewording to put the variable part at the end of the line, then doing more than one error message). The reader of the code needs to be able to compartmentalize and forget about the error handling code, and the break in the indenting for the pragmas literally gets in the way of that.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -248,9 +265,13 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         if (g6_EndReached(theG6ReadIterator))
             break;
 
-        gp_CopyGraph(copyOfOrigGraph, theGraph);
+        // NOTE: This will gp_CopyExtensions() from the origGraphRead to the
```
john-boyer-phd: Let's delete this comment because it assumes a way of implementing copy graph that is not only suboptimal and will be replaced, but the implementation is not something the app developer has to be aware of about the API once graphForEmbedding is properly extended above.

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: Just to tighten up a bit more, let's remove the linefeeds here, before and after `errorFlag = TRUE`.

john-boyer-phd, line 228: Lets tighten this up to remove distinction of modifier versus non-modifier and especially to remove the pragmas. By the time you're getting to this error, the user will undoubtedly know what algorithm they asked for, and the app developer could use a debugger if they somehow didn't know. So, this could be as short as doing two error messages with the line number trick above.

This is likely to be read more than most of the code in the library, so we're just trying to eliminate complexity to maximize its readability, especially in the happy path case.

john-boyer-phd: Let's remove linefeeds to tighten up error cases.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -132,9 +133,9 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         return Result;
     }
 
-    theGraph = gp_New();
+    origGraphRead = gp_New();
```
john-boyer-phd: It's normal in C to say `if ((origGraphRead = gp_New()) == NULL)` and it would tighten this up a bit.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -132,9 +133,9 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
 
-    if (theGraph == NULL)
+    if (origGraphRead == NULL)
     {
         ErrorMessage("Unable to allocate graph.\n");
```
john-boyer-phd: Please remove extra linefeeds here, too.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -143,11 +144,11 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         return NOTOK;
     }
 
-    if ((Result = g6_NewReader((&theG6ReadIterator), theGraph)) != OK)
+    if ((Result = g6_NewReader((&theG6ReadIterator), origGraphRead)) != OK)
     {
         ErrorMessage("Unable to allocate G6ReadIterator.\n");
 
-        gp_Free(&theGraph);
+        gp_Free(&origGraphRead);
```
john-boyer-phd, line 123: Really good catch, adding `gp_Free()`

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -143,11 +144,11 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
         stats->errorFlag = TRUE;
 
         return Result;
```
john-boyer-phd: Returning NOTOK here would be better because it is NOTOK and because it causes debug mode to report this location as the function call stack unrolls. Also, the if statement should just call the function without assigning to Result.

Please also remove the extra linefeeds.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
 
-    order = gp_GetN(theGraph);
-    // We have to set the maximum edge capacity (i.e. (N * (N - 1) / 2)) because some of the test files
-    // can contain complete graphs, and the graph drawing, K_{3, 3} search, and K_4 search extensions
-    // don't support expanding the edge capacity after being attached.
-    if (strchr("d34", command) != NULL)
+    order = gp_GetN(origGraphRead);
+
+    graphForEmbedding = gp_New();
+    if (graphForEmbedding == NULL)
```
john-boyer-phd: Let's compress the new/init logic here:

`if ((graphForEmbedding = gp_New()) == NULL || gp_InitGraph(graphFormEmbedding, order) != OK)`

In the error handling body below, you can add `gp_Free(&graphForEmbedding)` because it's needed if the init graph fails and it no-ops if the gp_New failed.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
     {
-        if ((Result = gp_EnsureEdgeCapacity(theGraph, (order * (order - 1) / 2))) != OK)
-        {
-            ErrorMessage("Unable to ensure sufficient edge capacity of the G6ReadIterator's graph struct.\n");
+        ErrorMessage("Unable to allocate graph to store copy of original graph before embedding.\n");
```
john-boyer-phd: The logic has been changed so that we're working on the graphForEmbedding, and we're also going for new and init in the if condition, so best to change the message to something like "Failed to allocate and initialize a graph for the embedding operation."

And please remove the linefeeds.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
-            g6_FreeReader((&theG6ReadIterator));
-            gp_Free(&theGraph);
-            stats->errorFlag = TRUE;
+        g6_FreeReader((&theG6ReadIterator));
+        gp_Free(&origGraphRead);
+        stats->errorFlag = TRUE;
 
-            return Result;
-        }
+        return NOTOK;
```
john-boyer-phd: Good returning NOTOK here.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
-        if ((Result = gp_EnsureEdgeCapacity(theGraph, (order * (order - 1) / 2))) != OK)
-        {
-            ErrorMessage("Unable to ensure sufficient edge capacity of the G6ReadIterator's graph struct.\n");
+        ErrorMessage("Unable to allocate graph to store copy of original graph before embedding.\n");
 
-            g6_FreeReader((&theG6ReadIterator));
-            gp_Free(&theGraph);
-            stats->errorFlag = TRUE;
+        g6_FreeReader((&theG6ReadIterator));
+        gp_Free(&origGraphRead);
```
john-boyer-phd: Again, very good catch here.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
-#pragma GCC diagnostic ignored "-Wformat-nonliteral"
-        sprintf(messageContents, messageFormat, commandStr);
-#pragma GCC diagnostic pop
+        g6_FreeReader((&theG6ReadIterator));
+        gp_Free(&origGraphRead);
+        gp_Free(&graphForEmbedding);
+        stats->errorFlag = TRUE;
 
-        ErrorMessage(messageContents);
+        return Result;
```
john-boyer-phd: I think this would have returned the wrong result, because it wasn't assigned the result of gp_InitGraph, so this is a good extra example reason why it's better to return NOTOK when we know that it really can only be NOTOK anyway.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
+    if ((Result = gp_EnsureEdgeCapacity(graphForEmbedding, (order * (order - 1) / 2))) != OK)
+    {
+        ErrorMessage("Unable to ensure sufficient edge capacity of the graph for embedding.\n");
+
+        g6_FreeReader((&theG6ReadIterator));
+        gp_Free(&origGraphRead);
+        gp_Free(&graphForEmbedding);
         stats->errorFlag = TRUE;
 
         return Result;
```
john-boyer-phd: Just return NOTOK here, and remove the preceding linefeed.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
+        gp_Free(&graphForEmbedding);
+        stats->errorFlag = TRUE;
 
-        ErrorMessage(messageContents);
+        return Result;
+    }
 
-        g6_FreeReader(&theG6ReadIterator);
-        gp_Free(&theGraph);
+    // We have to set the maximum edge capacity (i.e. (N * (N - 1) / 2)) because some of the test files
```
john-boyer-phd: Let's say "We set the maximum... because many test files have more than edges than the default of 3N". No reason to document a current API limitation on some algorithms, which will go away at some point, when the step makes a certain amount of sense on its own.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
 
         return Result;
     }
-
-    copyOfOrigGraph = gp_New();
-    if (copyOfOrigGraph == NULL)
+    // We have to set the maximum edge capacity because otherwise gp_CopyGraph()
+    // will fail due to the destination graph (graphForEmbedding) having a greater
+    // edge capacity than the source graph (origGraphRead)
+    if ((Result = gp_EnsureEdgeCapacity(origGraphRead, (order * (order - 1) / 2))) != OK)
```
john-boyer-phd: When this moves up, remove assignment to Result.

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: Also, let's remove the assignments to Result, just call the function and tests for != OK.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
+#pragma GCC diagnostic pop
+
+        ErrorMessage(messageContents);
+
+        g6_FreeReader(&theG6ReadIterator);
+        gp_Free(&origGraphRead);
+        gp_Free(&graphForEmbedding);
         stats->errorFlag = TRUE;
 
         return Result;
```
john-boyer-phd: Just return NOTOK here, and remove linefeeds to tighten.

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -316,8 +337,8 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
     stats->errorFlag = errorFlag;
 
     g6_FreeReader((&theG6ReadIterator));
-    gp_Free(&theGraph);
-    gp_Free(&copyOfOrigGraph);
+    gp_Free(&origGraphRead);
+    gp_Free(&graphForEmbedding);
 
     return Result;
```
john-boyer-phd, line 255: This will sometimes return the wrong result, such as if `g6_ReadGraph()` fails rather than `gp_Embed()` or, I think, if the last `gp_Embed()` returns NONEMBEDDABLE.

It's also better to return a direct NOTOK in the error case, so let's change this to `return !errorFlag ?  OK : NOTOK`

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: Death to pragmas!

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -265,7 +286,7 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
             break;
         }
 
-        if (gp_TestEmbedResultIntegrity(theGraph, copyOfOrigGraph, Result) != Result)
+        if (gp_TestEmbedResultIntegrity(graphForEmbedding, origGraphRead, Result) != Result)
         {
             messageFormat = "Embed integrity check failed for graph on line %d for command '%c'.\n";
 #pragma GCC diagnostic push
```
john-boyer-phd: Please remediate this pragma situation too. Focus on line number of failure and not command.

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: It would be tighter to promote this call into the above if stmt with g6_NewReader():

`if (g6_NewReader((&theG6ReadIterator), origGraphRead) != OK || g6_InitReaderWithFileName(theG6ReadIterator, infileName)) != OK)`

Then, the error message can be something like "Failed to initialize for reading all graphs from G6 source"

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -159,71 +160,87 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
                      "G6ReadIterator.\n");
 
         g6_FreeReader((&theG6ReadIterator));
-        gp_Free(&theGraph);
+        gp_Free(&origGraphRead);
         stats->errorFlag = TRUE;
 
         return Result;
```
john-boyer-phd: this should turn into a return NOTOK when the the initializer joins the NewReader call.

john-boyer-phd (changes requested):
May seem like a lot of comments, but they're all really small things that will hopefully help make this shorter and easier on the eyes. Coming along well!

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Just a stylistic question: I tried to separate each error block throughout the code that I've touched so that the `ErrorMessage()` comes first, then whatever cleanup actions need to happen, then the `return` statement... If you want me to address all the comments throughout `planarityTestAllGraphs.c`, do you want me to go one step further and update all error cases throughout the project so that there's no linefeeds between these three sections? If so, I'll make the requested changes to remove linefeeds in error cases on this PR, but then will add an issue to do the same throughout the project to make it consistent (also so that I'm not polluting this PR with a billion unrelated changes)

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Absolutely, great suggestion! There's been a bit of back-and-forth on this pattern throughout the project; it might be helpful to define a style guide to demonstrate "how we do things over here" (including the standard for what error blocks should look like; see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/206/changes#r3149568535)). (I could'a swore I've brought this up before, but I double-checked both the EAPS issues and my TODO list, and it doesn't seem like I actually created an issue for this LOL. Oops 🙏🏼 😭 )

edit: Created #207 and #208

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Will resolve this once we come to a concurrence about error block style! (see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/206/changes#r3149568535))

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Freakin' right on, yeah, that's another pattern that we should probably document in the style guide (see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/206/changes#r3149587305) re: style guide discussion) But yeah, I am pretty sure there's other places where we `return Result;` rather than explicitly `return NOTOK;` so the location doesn't actually show up in the "call stack" 😭 My bad!!!!! 🙏🏼 

Re: spacing, once we figure out error block formatting, we can make this change consistently! (see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/206/changes#r3149568535))

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: This is an interesting suggestion, but I'm leery to combine two separate actions into one statement (allocating vs. initialization), even if it does slim down the sourcecode a bit. Also, if we change this here (which I don't think we should, because we've been following the single responsibility principle by breaking these two tasks off into their own functions), we'd likely have to change all cases where we're allocating vs. initializing the datastructures.

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Like I mentioned in the [above comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/206/changes#r3150307295) about allocation vs. initialization of the `G6ReadIterator`, I don't know if I want to combine these two separate calls to allocate vs. initialize the `graphP` into one if statement 😭

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: OH SHOOT, good freakin' catch!!

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: I think that since we're `gp_EnsureEdgeCapacity()` for two different `graphP`, it *feels* better to keep the error-handling specific to the possible failure... However, this doesn't have the same level of "feels-bad-ness" as the decision to include allocation and initialization calls in the same if statement with a more generic error handling block (see [this comment](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/206/changes#r3150307295)).

Regarding the latter comment, I've actually combed over all the places where I'm assigning to `Result`, and I've removed a bunch of instances (basically everything other than when we embed and check embed result integrity!).

I also just double-checked that the places where I'm assigning `Result = NOTOK;` in the scope of an error block, that the `#define NOTOK (printf("NOTOK on Line %d of %s\n", __LINE__, __FILE__), debugNOTOK())` macro actually does show that particular line number in the "call stack"!

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Circling back to this, as per text convo, I'll be going ahead and removing the whitespace in between statements within error handling blocks within `planarityTestAllGraphs.c`, and will make a follow-up issue (a sub-issue of #208) to make this consistent throughout the `planarityApp` layer!

wbkboyer: As per conversation with @john-boyer-phd , will be proceeding with this change (captured in #207 )!

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: Should I remove the comment indicating why I'm having to `ExtendGraph(origGraphRead, command)` as well as `ExtendGraph(graphForEmbedding, command)`?

Also, regarding the upcoming changes to `gp_CopyGraph()`, will one of the requirements be that if `destGraph` has extensions attached to it but `srcGraph` doesn't, we should not `gp_FreeExtensions()` on the `destGraph`, but rather just reinitialize those structures (or would it be safe to just leave them to be overwritten)? That way, we won't have to `ExtendGraph()` in the below loop...

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: The only problem is that we still need to populate the error message with the line number, and this dynamic message requires a `messageFormat` string for the `messageContents`. Unless we create a variadic version of `ErrorMessage()` that would allow format specifiers, I don't think we'll be able to get around this :(

edit: this also impacts the above if-statement for the result of embedding as well

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: I made sure to add `Result = NOTOK` (which will show up in the callstack) for the above error handling blocks; do I still need to make this change? If so, then should the return value be defined in terms of the `Result` as well as the `errorFlag`, just in case one forgets to set `errorFlag = TRUE` when `Result` is set to `NOTOK`?

```
return ((Result == OK || Result = NONEMBEDDABLE) && !errorFlag) ? OK : NOTOK;
```

Honestly, I'm half-tempted to remove the `errorFlag` assignments and to do that at the very end based on the `Result` value... It would simplify the error handling blocks, and would remove this ternary return code calculation!

john-boyer-phd (approved):
This is a wonderful improvement, especially the variadic update to ErrorMessage() and Message(), which have been a nuisance for a very long time.

Issue #211, opened 2026-04-30
Build failure for MSVC cl due to unsupported `__attribute__()` syntax


wbkboyer:
## Describe the bug

On #206 , I introduced the variadic `Message()` and `ErrorMessage()` functions. When the project is compiled using MSVC `cl` on Windows, the syntax

```
    extern void Message(char const *message, ...) __attribute__((format(printf, 1, 2)));
    extern void ErrorMessage(char const *message, ...) __attribute__((format(printf, 1, 2)));
```

Causes a significant number of compilation errors to be thrown, e.g. `C2061`, `C2059`, `C4131`, `C4431`, `C4218`, and `C4013`.

## Environment

 - **OS** Windows MSVC `cl`

wbkboyer:
I think I've got the fix: basically, if we're using MSVC cl (or don't have function attributes, i.e. `__attribute__` not defined), then don't apply the format attribute. Otherwise, if you're using `gcc`, use `gnu_printf` archetype (rather than `printf`), or just use `printf`.

I was able to build using MSVC cl locally *and* for different architectures for the new GHA job (#210), as well as MSYS2 UCRT64 `gcc`, and finally (still) `clang` on MacOS

https://gcc.gnu.org/onlinedocs/gcc-4.4.7/gcc/Function-Attributes.html

PR #212, opened 2026-05-05
Issue #210 & #211 - Add MSVC `cl` job to GitHub action workflow file, and fix MSVC cl break for variadic functions

### LINKED ISSUE #210: Create GitHub Action to test compilation for MSVC `cl`
## Problem statement

On #206 , I introduced the variadic `Message()` and `ErrorMessage()` functions. However, this syntax doesn't appear to be supported by MSVC `cl`, but we didn't discover this on the pull request because there's no job for this compiler toolchain.

## Proposed solution

Either add a new job, or add MSVC `cl` on Windows to some platform/compiler matrix.

wbkboyer:
Resolves #210 and #211

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Added

* N/A

### Updated

* `.github/workflows/compiler-versions.yml` - renamed from `compilers.yml`
* `.github/workflows/dev-platforms.yml` - renamed from `os.yml`. Organized jobs alphabetically, and added `msvc` job to build for `x64`, `x86`, `x86_amd64`, and `amd64_x86` architectures
* `c/graphLib/lowLevelUtils/apiutils.c` - removed `#pragma` to ignore `-Wsuggest-attribute=format`, since we now use `gnu_printf` archetype for `gcc` (see [`gcc` Docs - Function-Attributes](https://gcc.gnu.org/onlinedocs/gcc-4.4.7/gcc/Function-Attributes.html))
* `c/graphLib/lowLevelUtils/apiutils.h` - If the compiler is MSVC cl (or any compiler that is not `gcc` nor `clang` and doesn't have macro `__has_attribute` defined; see [this](https://stackoverflow.com/a/54032648/5456879) stackoverflow response), then don't apply the `format` attribute. If the compiler has the `__attribute__` keyword (or, explicitly calling out `gcc` and `clang` versions where the `__has_attribute` macro may not be defined), then use `gnu_printf` archetype for `gcc` and use the `printf` archetype otherwise.
    * __**NOTE:**__ Perhaps this logic can be simplified, but I wanted to be explicit that MSVC cl would not apply the format, and that in all other reasonable cases we'd apply the correct format with the right archetype.
 
### Removed

* N/A

## Testing

- [x] Manually built on MacOS with `clang`, and on Windows with `gcc` via MSYS2 and MSVC `cl`, then ran `-test`; also used autotools on Windows with `--enable-compile-warnings` to make sure no warnings raised (didn't install or run tests)
- [x] [Successful job run on fork](https://github.com/wbkboyer/edge-addition-planarity-suite/actions/runs/25396575992)


wbkboyer:
No idea why [`clang-18` job for `compiler-warnings.yml`](https://github.com/graph-algorithms/edge-addition-planarity-suite/actions/runs/25399617646/job/74495730186?pr=212) is failing, since [`clang-18` job for `compiler-versions.yml`](https://github.com/graph-algorithms/edge-addition-planarity-suite/actions/runs/25399617682/job/74495409866?pr=212) and [`macos` job for `dev-platforms.yml`](https://github.com/graph-algorithms/edge-addition-planarity-suite/actions/runs/25399617679/job/74495409713?pr=212) are working... Will re-run one more time in a while, then continue to troubleshoot if the issue persists.

edit: One thing I notice is that it's failing on `clang` on `ubuntu-latest` runner (although the `compiler-versions.yml` is the same, it's a *warning* being treated as an error??), which might mean that [`__GNUC__`](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html#Common-Predefined-Macros-1) is defined for `clang` on `ubuntu-latest`, and so we're using an unsupported archetype... so maybe I need to add [`!defined(__clang__)`](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-macros)????

john-boyer-phd (approved):
That's really an excellent solution to extend the variadic solution to MSVC cl while still getting gcc and clang to do the right things.

PR #215, opened 2026-05-12
Fixes to hide internal data structures of G6 read and write iterators

### LINKED ISSUE #167: Create accessor functions for `G6(Read|Write)Iterator` attributes used at the `planarityApp` layer
During #165 , I noticed that there are instances where I'm directly accessing members of the `G6(Read|Write)Iterator` struct (use regex `g6(Read|Write)Iterator->` to search through codebase for all examples), which is a Bad Thing™. These should *not* be macros, since performance using the iterators is dominated by IO operations, and it's more helpful to get the stack trace with appropriate error messaging/handling.

john-boyer-phd:
Resolves #167 

## Changes

1. Hid internal data structures for G6 read/write iterators by moving the declarations from the public header files to the c source files.
2. Replaced read iterator data member accesses in test all graphs with calls to the getter method.
3. Removed assignment of NULL to `currGraph` when `endReached` during `g6_ReadGraph()` because assigning NULL causes the iterator to report it is not initialized, which prevents success on the last call of `g6_GetNumGraphsRead()` in test all graphs.

## Testing

Ran test all graphs with N5-all.g6 from samples directory.

(The last line of this PR template needs to have a details end tab, so please add a slash to it)

### DIFF
--- a/c/graphLib/io/g6-read-iterator.c
+++ b/c/graphLib/io/g6-read-iterator.c
@@ -37,6 +37,32 @@ int _g6_DecodeGraph(char *graphBuff, const int order, const int numChars, graphP
 int _g6_ReadGraphFromFile(graphP theGraph, char *pathToG6File);
 int _g6_ReadGraphFromString(graphP theGraph, char *g6EncodedString);
 
+/********************************************************************
+ Package private structure declaration for read iterator
+ ********************************************************************/
+    typedef struct strOrFileStruct strOrFileStruct;
+    typedef strOrFileStruct *strOrFileP;
+
+    struct G6ReadIteratorStruct
+    {
+        strOrFileP inputContainer;
+        int numGraphsRead;
+
+        int order;
+        int numCharsForOrder;
+        size_t numCharsForGraphEncoding;
+        size_t currGraphBuffSize;
+        char *currGraphBuff;
+
+        graphP currGraph;
+
+        bool endReached;
+    };
+
+/********************************************************************
+ Public and package private method implementations for read iterator
+ ********************************************************************/
+
 int g6_NewReader(G6ReadIteratorP *pG6ReadIterator, graphP theGraph)
 {
     int exitCode = OK;
@@ -614,7 +640,6 @@ int g6_ReadGraph(G6ReadIteratorP theG6ReadIterator)
     }
     else
     {
-        theG6ReadIterator->currGraph = NULL;
         theG6ReadIterator->endReached = true;
     }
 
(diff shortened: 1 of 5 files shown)


c/graphLib/io/g6-read-iterator.c
```diff
@@ -37,6 +37,32 @@ int _g6_DecodeGraph(char *graphBuff, const int order, const int numChars, graphP
 int _g6_ReadGraphFromFile(graphP theGraph, char *pathToG6File);
 int _g6_ReadGraphFromString(graphP theGraph, char *g6EncodedString);
 
+/********************************************************************
+ Package private structure declaration for read iterator
+ ********************************************************************/
+    typedef struct strOrFileStruct strOrFileStruct;
+    typedef strOrFileStruct *strOrFileP;
+
+    struct G6ReadIteratorStruct
```
wbkboyer, line 46: This is a great pattern, moving the `struct` definition into the source file and only exposing the type and relevant getters through the public header!

c/graphLib/io/g6-read-iterator.c
```diff
@@ -614,7 +640,6 @@ int g6_ReadGraph(G6ReadIteratorP theG6ReadIterator)
     }
     else
     {
-        theG6ReadIterator->currGraph = NULL;
```
wbkboyer, line 617: Great catch!!! `endReached` is the only flag we should be looking at to determine if there's no more graphs to read from the input container 🎉 Otherwise, in `planarityTestAllGraphs.c`, since you're now using the "getter" function for `numGraphsRead` on line 256, you're going to get a `NOTOK` result because `g6_GetNumGraphsRead()` checks `_g6_IsReaderInitialized()` as one of the steps before accessing the private value. Also, when we go to `g6_FreeReader()`, we already set the `currGraph` to `NULL` to show that the reader relinquishes ownership, so there's absolutely no need to `NULL` it out here!

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -227,23 +233,27 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
-                                theG6ReadIterator->numGraphsRead + 1);
+                                command, modifier, numGraphsRead + 1);
             }
             Result = NOTOK;
             break;
         }
     }
 
-    stats->numGraphsRead = theG6ReadIterator->numGraphsRead;
+    stats->numGraphsRead = 0;
```
wbkboyer, line 255: Another good call to set `stats->numGraphsRead` to 0 here, because if we encounter an error in `g6_GetNumGraphsRead()`, this value might subsequently contain garbage (although in `TestAllGraphs()`, I `memset()` the `testAllStats` struct).

wbkboyer (approved):
Great changes to hide the reader/writer internals from the public API, as proved out in the `planarityApp` layer with `TestAllGraphs()` using the getter functions! 🚀

PR #217, opened 2026-05-22
Issue #209 - dynamic edge capacity for extensions

### LINKED ISSUE #209: Implement Ensure Edge Capacity for Extensions
The extensions for planar graph drawing and for K_{3,3} and K_4 homeomorph search algorithms currently have `EnsureEdgeCapacity` function overloads that return `NOTOK`. With newer capabilities, like testing all graphs in a G6 file, this is causing us to have to ensure edge capacity for a complete graph on N vertices because we don't know how many edges will be in the graphs in a G6 file until after the first one is read. The reader itself uses `gp_DynamicAddEdge()`, so it can dynamically expand edge capacity as needed, except that the three extensions mentioned above don't allow it.

The goal for this issue is to implement the `EnsureEdgeCapacity` functions for the three extensions. It would be better to use `realloc()` because it can often expand the currently allocated memory without having to copy the contents to a new location. However, `realloc()` may have to allocate new memory and copy the old content over, then free the old memory, so the pointer it returns must be placed in the extension data structure as its new parallel edge array.

If `realloc()` fails, it will return NULL, in which case don't replace the E array pointer in the extension data structure and instead return NOTOK.

Whether or not `realloc()` changes the content location, the extra array space in the new array must be initialized via a `memset()` call. An example of the `memset` call needed for each can be found in the extensions C code file for each extension in the `_InitStructures()` methods, except that the first and third parameters must be amended so that  only the extra array space is initialized. For K_{3,3} and K_4 homeomorph search algorithms, the extra space in the parallel edge array should be set to `NIL_CHAR`. For planar graph drawing, the extra space should always be set to 0.

In DrawPlanar, the `_InitStructures()` method gives NIL_CHAR to the `memset` in the 1-based array case. This should be changed to 0 because that is what NIL_CHAR is, and if you look at the 0-based array case, in `_InitEdgeRec()`, it will be clear that 0 is what it should always be, not NIL_CHAR. 

Finally, once the three extensions have working `EnsureEdgeCapacity` methods that return OK, then amend the test all graphs method to eliminate the EnsureEdgeCapacity calls that appear before the main loop, then test to ensure that test all graphs still works on G6 files that contain graphs with more than 3N edges. Ideally, this should speed up testing all graphs for a given order N because the edge capacity will remain at 3N for all G6 files containing graphs with 3N or fewer edges.

john-boyer-phd:
Resolves #209

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Changes

Main change was adding implementations of EnsureEdgeCapacity for Drawing, K3,3 search, and K4 search so that the direct ensure edge capacity calls could be removed from testAllGraphs. So, when reading all graphs from a file, the G6 read iterator will implicitly grow the edge capacity only when needed.

This occurs for some graphs at N=8 and larger. So, added the N=8 all edges G6 file to the samples, and tests of all graphs in it for all 6 main graph algorithms to the -test quick regression tests, so that the sanitizers will also run on all algorithms for all N=8 graphs from now on.

Added a private getter function for the DRAWPLANAR_ID.

Rearranged the definitions of edge array iteration methods to ensure that the primary expected method is first.

A few other insignificant and non-functional changes.

## Testing

Compiled with gcc and cl on Windows. Ran Warnings GA. Ran Sanitisers GA.

<details>


c/graphLib/homeomorphSearch/graphK33Search.c
```diff
@@ -362,7 +362,7 @@ int _RunExtraK33Tests(graphP theGraph, K33SearchContext *context)
     // Now mark the lowest x-y path so that we can test whether _any_ x-y path
     // has points of attachment, px or py, below x or y, respectively (where
     // below means closer to W than x or y, respectively, along the external face).
-    if (_MarkLowestXYPath(theGraph) != TRUE)
+    if (_MarkLowestXYPath(theGraph) != OK)
```
wbkboyer, line 365: Good catch! Do you need to add check if `IC->py == NIL` after marking the lowest xy path?

c/graphLib/io/strOrFile.c
```diff
@@ -500,7 +507,14 @@ char sf_ungetc(char theChar, strOrFileP theStrOrFile)
         sp_GetCurrentSize(theStrOrFile->ungetBuf) >= sp_GetCapacity(theStrOrFile->ungetBuf))
         return EOF; // Acceptable downcast, allowing char rather than int return type
 
+#ifndef DEBUG
     sp_Push(theStrOrFile->ungetBuf, theChar);
+#else
+    // In DEBUG, MSVC cl is misreading the printf() generated by sp_Pop(),
```
wbkboyer: Should this comment be referring to `sp_Push()`?

---

Also, I need some help understanding these changes to `strOrFile.c`'s `sf_getc()` and `sf_ungetc()`! <3 Looking at `sp_Push()` macro in `stack.h`:

```
#define sp_Push(theStack, a) theStack->S[theStack->size++] = a
```

vs. `sp__Push()` function in `stack.c`:

```
int sp__Push(stackP theStack, int a)
{
    if (theStack->size >= theStack->capacity)
        return NOTOK;

    theStack->S[theStack->size++] = a;
    return OK;
}
```

Isn't the opposite true, that the `NOTOK` macro expansion for Debug builds has the `printf()` that offends MSVC cl's sensibilities?

I actually commented-out the changes to `c/graphLib/io/strOrFile.c` and compiled the Debug build using MSVC cl (19.44.35224), and no warnings/errors about `printf()` were emitted (only `'return': conversion from 'int' to 'char', possible loss of data`)!

c/graphLib/homeomorphSearch/graphK33Search.c
john-boyer-phd: The code right below essentially does that test.

c/graphLib/io/strOrFile.c
john-boyer-phd: In DEBUG, the single underscore sp_Push() resolves to something else that includes a NOTOK and then calls sp__Push(). MSVC is misreading the printf in the single underscore one and thinking that there is an in to char loss of data when there isn't. The same compiler is not getting choked up about the printf in the NOTOK (nor anyplace else).

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -135,28 +133,7 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
-    // Additionally, we have to set the maximum edge capacity because otherwise
-    // gp_CopyGraph() will fail due to the destination graph (graphForEmbedding)
-    // having a greater edge capacity than the source graph (origGraphRead)
-    if (
-        gp_EnsureEdgeCapacity(origGraphRead, (order * (order - 1) / 2)) != OK ||
-        gp_EnsureEdgeCapacity(graphForEmbedding, maxNumEdgesForOrder) != OK)
-    {
-        gp_ErrorMessage("Unable to ensure sufficient edge capacity of the "
-                        "original graph read or the graph for embedding.\n");
+        gp_ErrorMessage("Unable initialize graph for embedding.\n");
```
wbkboyer, line 136: [Suggestion] Should this `gp_ErrorMessage()` be changed back to

```
gp_ErrorMessage("Unable to allocate or initialize graph for embedding "
                        "operation.\n");
```

Since we're performing both the allocation (`gp_New()`) and initialization (`gp_InitGraph()`) in the same block, or is it just too wordy?

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -182,8 +158,7 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         {
             int numGraphsRead = 0;
             g6_GetNumGraphsRead(theG6ReadIterator, &numGraphsRead);
-            gp_ErrorMessage("Unable to read graph on line %d from .g6 read "
-                            "iterator.\n",
+            gp_ErrorMessage("Unable to read graph on line %d.\n",
```
wbkboyer, line 161: [Suggestion] Might be useful to still specify the read operation failed for the `G6ReadIterator`, although perhaps it was too wordy...

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -235,16 +209,14 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
+                gp_ErrorMessage("Command '%c' error on graph on line %d.\n",
                                 command, numGraphsRead + 1);
             }
             else
             {
                 int numGraphsRead = 0;
                 g6_GetNumGraphsRead(theG6ReadIterator, &numGraphsRead);
-                gp_ErrorMessage("Error applying algorithm '%c' with modifier '%c' "
-                                "to graph on line %d.\n",
+                gp_ErrorMessage("Command '%c%c' error on graph on line %d.\n",
```
wbkboyer, line 219: That's much more concise! Nice!

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -279,7 +251,7 @@ int outputTestAllGraphsResults(char command, char modifier, testAllStatsP stats,
     int headerStrLen = 0, resultStrLen = 0;
     char *resultsStr = NULL;
 
-    if (outfileName == NULL && (pOutputStr == NULL || *pOutputStr == NULL))
+    if (outfileName == NULL && (pOutputStr == NULL || *pOutputStr != NULL))
```
wbkboyer, line 254: WHOOOOOOPS GOOD CATCH: if `(*pOutputStr)` was allocated, we'd leak memory :(

c/graphLib/homeomorphSearch/graphK33Search.c
```diff
@@ -918,7 +918,7 @@ int _FindK33WithMergeBlocker(graphP theGraph, K33SearchContext *context, int v,
  The function returns NOTOK on internal error, OK otherwise.
 
  Preconditions: The X-Y path is marked visited by a prior invocation of
- the meothd _MarkLowestXYPath() above.
+ the method _MarkLowestXYPath() above.
```
wbkboyer, line 921: Geeze, no idea why this wasn't caught by previous runs of `spelling.yml`!! Good catch <3

c/graphLib/homeomorphSearch/graphK33Search_Extensions.c
```diff
@@ -650,13 +685,9 @@ int _K33Search_HandleBlockedBicomp(graphP theGraph, int v, int RootVertex, int R
...
     }
-    else
-    {
-        return context->functions.fpHandleBlockedBicomp(theGraph, v, RootVertex, R);
-    }
 
-    // No way to get here in current implementation, but this protects against future mistakes
-    return NOTOK;
+    // else if we are not doing a K3,3 homeomorph search, then call the superclass to handle
+    return context->functions.fpHandleBlockedBicomp(theGraph, v, RootVertex, R);
```
wbkboyer, line 690: Nice!!! This addresses the unreachable code warning for MSVC cl!

c/graphLib/planarityRelated/graphDrawPlanar_Extensions.c
```diff
@@ -170,6 +170,16 @@ int gp_Detach_DrawPlanar(graphP theGraph)
...
 }
 
+/********************************************************************
+ gp_GetDrawPlanarExtensionIdentifier()
+ A private function that returns the DRAWPLANAR_ID.
+ ********************************************************************/
+
+int gp_GetDrawPlanarExtensionIdentifier(void)
+{
+    return DRAWPLANAR_ID;
```
wbkboyer, line 180: Suhweeeet, this should help at the Cython layer in the `classic` subpackage's `PGraph` class methods `nodes()` and `edges()`, where we're currently hard-coding `DRAWPLANAR_ID = 1` 🎉

c/planarityApp/planarityRandomGraphs.c
```diff
@@ -135,7 +135,7 @@ int RandomGraphs(char const *const commandString, int NumGraphs, int SizeOfGraph
     // Seed the random number generator with "now". Do it after any prompting
     // to tie randomness to human process of answering the prompt.
     // Acceptable downcast of time_t to unsigned int (seeding benefits from the lower bits of now)
-    srand(time(NULL));
+    srand((unsigned int) time(NULL));
```
wbkboyer, line 138: Good catch with this cast: [`srand()`](https://cplusplus.com/reference/cstdlib/srand/) is indeed looking for an `unsigned int`!! 🚀

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -95,21 +95,20 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
-    int order = 0, maxNumEdgesForOrder = 0;
+    int order = 0;
 
     G6ReadIteratorP theG6ReadIterator = NULL;
 
     if (GetEmbedFlags(command, modifier, &embedFlags) != OK)
     {
-        gp_ErrorMessage("Unable to derive embedFlags from command and modifier "
-                        "characters.\n");
+        gp_ErrorMessage("Invalid command or modifier.\n");
```
wbkboyer, line 104: [Suggestion] Might still be helpful to know what operation failed due to invalid command or modifier!

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -118,8 +117,7 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         g6_NewReader((&theG6ReadIterator), origGraphRead) != OK ||
         g6_InitReaderWithFileName(theG6ReadIterator, infileName) != OK)
     {
-        gp_ErrorMessage("Unable to test all graphs due to failure to allocate or "
-                        "initialize G6ReadIterator.\n");
+        gp_ErrorMessage("Unable to allocate or initialize G6 read iterator.\n");
```
wbkboyer, line 120: [Suggestion] Might still be helpful to indicate what process failed (i.e. `testAllGraphs()`) as a result of the failure to initialize the reader (but also might be too wordy)

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -95,21 +95,20 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
...
-                        "characters.\n");
+        gp_ErrorMessage("Invalid command or modifier.\n");
         stats->errorFlag = TRUE;
         return NOTOK;
     }
 
     if ((origGraphRead = gp_New()) == NULL)
     {
-        gp_ErrorMessage("Unable to allocate graph.\n");
+        gp_ErrorMessage("Unable to allocate graph for reading.\n");
```
wbkboyer, line 111: [Suggestion] Maybe indicate `testAllGraphs()` failed as a result of failure to allocate `origGraphRead`?

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: Just trying to avoid wraps. And one reason you can't init is if you can't allocate.

c/planarityApp/planarityCommandLine.c
```diff
@@ -444,29 +418,122 @@ int runSpecificGraphTests(char const *samplesDir)
...
+        case '4':
+            theValidationStr = K4SearchValidationStr;
+            break;
+        default:
+            Result = NOTOK;
+            break;
+        }
+
+        if (theValidationStr != NULL)
+            Result = strstr(outputStr, theValidationStr) ? OK : NOTOK;
```
wbkboyer, line 525: Nice use of [`strstr()`](https://en.cppreference.com/c/string/byte/strstr) to get around the fact that the `DURATION` field on the first line of the `outputStr` might vary between runs!! 💪🏼

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: Yeah, it's the wrap. no need to know G6ReadIterator in the msg. Programmer still has to find the same stuff to debug.

c/graphLib/homeomorphSearch/graphK33Search.c
john-boyer-phd: me either! My bad.

c/graphLib/planarityRelated/graphDrawPlanar_Extensions.c
john-boyer-phd: Yes indeed. (At least until we add the extra APIs that the Planarity Python package's classic API really needs)

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: Not really, still programmer problem. One of these days, we're going to exploit your new variadic gp_ErrorMessage everywhere so that both release and debug will get line numbers and files. Until then, the shorter string avoids the wrap and doesn't contain a lot less info.

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: just trying to kill wraps. This will never really happen; if it did, the user will generally know what they did, and they will be a programmer who will still have to hop into debug mode, at which point the NOTOK will say even more.

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: I'm up for little changes that don't cause whatever wrap formatter you're using to make two lines :-)

c/planarityApp/planarityCommandLine.c
john-boyer-phd: Yeah that was a good call you made!

c/graphLib/io/strOrFile.c
wbkboyer: Whoops, I was looking at `sp_Push()` definition on line 107 (which is what's used when `SPEED_MACROS` is defined!!!!) rather than the definition on line 46 (line 114 vs. line 60 for `sp_Pop()`! Now I see the `NOTOK` you're referring to ❤️ 

Also, now I understand that those two warnings were what you were trying to resolve with this change. Looks good to me, thanks for walking me through this!!

c/planarityApp/planarityTestAllGraphs.c
john-boyer-phd: For all of the "might be better" comments, tweaks that keep it to one line are OK :-)

wbkboyer (approved):
Looks great! Lots of excellent catches alongside the implementation of ensure edge capacity for draw planar, $K_{3,3}$ search, $K_4$ search!

PR #220, opened 2026-05-25
Refactor quick regression test groups

### LINKED ISSUE #219: Minor refactor of runQuickRegressionTests
Currently, `runQuickRegressionTests` ends by calling `runSpecificGraphTests`, which then runs all the specific graph tests, transformation tests, and test all graphs tests.

This should be refactored to have two new extra methods, one that runs all transform tests and one that runs all test all graphs tests. Then,  `runQuickRegressionTests` should call `runSpecificGraphTests` first and if its return is != OK, then return NOTOK. Then, use the same pattern to call the two new extra methods for running all transformation and all test all graphs tests, return NOTOK if each of their returns is != OK. Finally, return OK if all three are successful.

shivam2931120:
## Summary
- split quick regression orchestration so `runQuickRegressionTests` calls specific graph, transformation, and all-graphs test groups separately
- keep sample directory setup/restoration and final success/failure reporting in `runQuickRegressionTests`
- add dedicated `runGraphTransformationTests` and `runTestAllGraphsTests` helpers

Fixes #219.

## Validation
- `git diff --check`
- `gcc -std=c99 -Wall -Wextra -Ic -Ic/graphLib -Ic/graphLib/lowLevelUtils -Ic/graphLib/io -Ic/planarityApp -c c/planarityApp/planarityCommandLine.c -o /tmp/planarityCommandLine.o`
- `gcc -std=c99 -Ic -Ic/graphLib -Ic/graphLib/lowLevelUtils -Ic/graphLib/io -Ic/graphLib/extensionSystem -Ic/graphLib/planarityRelated -Ic/graphLib/homeomorphSearch -Ic/planarityApp $(rg --files c/graphLib c/planarityApp -g "*.c") -lm -o /tmp/planarity`
- `/tmp/planarity -test c/samples`

Note: I could not run the documented `autoreconf -fi && ./configure && make` flow because `autoreconf` is not installed in this environment.

### DIFF
--- a/c/planarityApp/planarityCommandLine.c
+++ b/c/planarityApp/planarityCommandLine.c
@@ -23,7 +23,9 @@ int callRandomNonplanarGraph(int argc, char *argv[]);
 int callTestAllGraphs(int argc, char *argv[]);
 int callTransformGraph(int argc, char *argv[]);
 
-int runSpecificGraphTests(char const *);
+int runSpecificGraphTests(void);
+int runGraphTransformationTests(void);
+int runTestAllGraphsTests(void);
 int runSpecificGraphTest(char const *command, char const *infileName, int inputInMemFlag);
 int runGraphTransformationTest(char const *command, char const *infileName, int inputInMemFlag);
 int runTestAllGraphsTest(char const *commandString, char const *infileName);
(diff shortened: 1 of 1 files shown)


c/planarityApp/planarityCommandLine.c
```diff
@@ -202,6 +197,7 @@ int runSpecificGraphTests(char const *samplesDir)
             {
                 // Give success result, but Warn if no samples (except no warning if in quiet mode)
                 gp_Message("WARNING: Unable to change to samples directory to run tests on samples.\n");
+                chdir(origDir);
```
john-boyer-phd, line 200: Very nice catch.

john-boyer-phd (approved):
Thank you for this excellent contribution, Shivam.

Issue #218, opened 2026-05-22
`spelling.yml` `actions-codespell` failure due to misconfigured `skip` option


wbkboyer:
## Describe the bug

Looking at the logs for the [failed run](https://github.com/graph-algorithms/edge-addition-planarity-suite/actions/runs/26304848856/job/77438886864?pr=217) of [`spelling.yml`](https://github.com/graph-algorithms/edge-addition-planarity-suite/blob/master/.github/workflows/spelling.yml) workflow file:

```
name: Spelling
on: [pull_request, workflow_dispatch]

jobs:
  codespell:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: codespell-project/actions-codespell@v2
        with:
          skip: c/samples
```

I see:
```
Running codespell on '' with the following options...
...
Skipping 'c/samples'
...
Resulting CLI options  --skip c/samples
```

And so the job chokes on files under `c/samples`. Looking at the [`codespell` repo](https://github.com/codespell-project/codespell), it seems like the flag should actually be `--skip="c/samples"`, not `--skip c/samples` (I also tried `skip: c/samples/` and `skip: c/samples/**` before I checked the `codespell` docs)

## Remediation steps

Coordinate with the `codespell-project` maintainers to troubleshoot and fix the issue, then if a new version of the action is released, update the `spelling.yml` file `uses` line for `actions-codespell`.

In the meanwhile, disable the workflow.

## Additional context

* See [`codespell-project/actions-codespell` Issue #96](https://github.com/codespell-project/actions-codespell/issues/96) for description of issue; [`codespell-project/codespell` repo](https://github.com/codespell-project/codespell) gives `skip` flag syntax as `--skip="<comma-separated filenames/dirs>"

john-boyer-phd:
Contrary to documentation, it needs a leading `./`. At that point, the whole directory can be skipped, not just the one file.

PR #221, opened 2026-05-26
Issue216 improve copy extensions and Issue218 spell checker fix

### LINKED ISSUE #216: Improve Copy Extensions logic in gp_CopyGraph()
When `gp_CopyGraph()` copies from a srcGraph to a dstGraph, the graphs may be different subclasses of the base Graph. The subclass extensions have to be copied from srcGraph to dstGraph. The way this is being done should be improved. Currently, the logic starts by freeing all extensions of the dstGraph, then copying over each one from the srcGraph. But if the srcGraph is a base Graph used for reading input only and the dstGraph is a subclass that will be used to perform an advanced function of `gp_Embed()`, for example, then the subclassing of the dstGraph is removed. 

Instead, we want the dstGraph to receive the copy of all the content it can get from the srcGraph, and then initialize/reinitialize any extension data structures that the srcGraph doesn't have. In this way, we copy the base srcGraph into the subclass dstGraph without losing the subclassiness of the dstGraph.

Specifically,
1. Any extension in dstGraph not in srcGraph should be initialized/reinitialized
2. Any extension in srcGraph not in dstGraph should be added to dstGraph (current behavior)
3. Remove unnecessary ExtendWith in test all graphs logic and retest

Note that Planarity and DFSUtils are now coded as extensions to the base graph, so 1 and 2 above should be coded to ensure that those extensions are reinitialized in dstGraph if srcGraph doesn't have them to copy (1), or added to dstGraph the are absent but srcGraph has them (2).

john-boyer-phd:
Resolves #216 (and #218)

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Updated

* Updated the `gp_CopyExtensions()` method and dependent code to copy extension data rather than freeing destination extensions and then duplicating the source extensions.
* Fixed the spell-checker to ignore the data files in the samples directory

## Testing

Release and Debug, gcc and cl on windows, test all graphs and random graphs for all extensions

### DIFF
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -8,4 +8,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: codespell-project/actions-codespell@v2
         with:
-          skip: ../../c/samples/n8.mALL.g6
+          skip: ./c/samples
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -8,4 +8,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: codespell-project/actions-codespell@v2
         with:
-          skip: c/samples/n8.mALL.g6
+          skip: ../../c/samples/n8.mALL.g6
--- a/c/graphLib/extensionSystem/graphExtensions.c
+++ b/c/graphLib/extensionSystem/graphExtensions.c
@@ -53,19 +53,10 @@ static int moduleIDGenerator = 0;
      An instance of this context structure is passed to the "context"
      parameter of gp_AddExtension().
 
-  3) Define a function capable of duplicating your context data
-     structure.  It receives a void pointer indicating the context
-     to duplicate and a void pointer that can be cast to a graph
-     pointer indicating the graph for which the context is being
-     duplicated.  The void pointer returned indicates the newly
-     allocated context structure.  The pointer to this function is
-     passed to the "dupContext" parameter of gp_AddExtension()
-
-     Note: It is useful to store in your context structure a pointer
-     to the graph that the context is extending.  There are certain
-     function overloads you will perform that will only receive
-     the context, and you may need to know things about the graph,
-     such as the number of vertices or edges.
+  3) Define a function capable of copying your context data. It receives 
+     void pointers to destination and source contexts, and then copies the
+     extension-specific data from source to destination. This function's
+     pointer is passed to the "copyData" parameter of gp_AddExtension()
 
   4) Define a function that can free the memory used by your context
      data structure.  It will receive a void pointer indicating the
@@ -132,7 +123,8 @@ static int moduleIDGenerator = 0;
         of fpReadPostprocess() and fpWritePostprocess() are needed.
 
   7) Define internal functions for _Feature_ClearStructures(),
-     _Feature_CreateStructures() and _Feature_InitStructures();
+     _Feature_CreateStructures(), _Feature_InitStructures(), and
+     _Feature_CopyStructures();
 
      a) The _Feature_ClearStructures() should simply null out pointers
         to extra structures on its first invocation, but thereafter it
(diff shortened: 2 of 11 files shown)


.github/workflows/spelling.yml
```diff
@@ -8,4 +8,4 @@ jobs:
       - uses: actions/checkout@v4
       - uses: codespell-project/actions-codespell@v2
         with:
-          skip: c/samples/n8.mALL.g6
+          skip: ./c/samples
```
wbkboyer, line 11: Great fix! 🎉  D'OH at prepending with `./`, since the [`actions-codespell`](https://github.com/codespell-project/actions-codespell) repo `README` gives the wrong syntax 😭 (see [`actions-codespell` Issue #96](https://github.com/codespell-project/actions-codespell/issues/96))

Image

c/graphLib/extensionSystem/graphExtensions.c
```diff
@@ -459,39 +458,51 @@ graphExtensionP _FindNearestOverload(graphP theGraph, graphExtensionP target, in
  gp_CopyExtensions()
  ********************************************************************/
 
+// This number can just be made bigger if ever needed
+#define MAXNUMSUPPORTEDEXTENSIONS 32
```
wbkboyer, line 462: As per our convo, we'd only need to update this constant if there's more than 32 extensions in EAPS (and we're currently sitting at 6) 🎉 , so this seems like a sensible limit (since the `embedFlags` is an `unsigned int` comprising 32 bits).

c/graphLib/extensionSystem/graphExtensions.c
```diff
@@ -459,39 +458,51 @@ graphExtensionP _FindNearestOverload(graphP theGraph, graphExtensionP target, in
...
+
 int gp_CopyExtensions(graphP dstGraph, graphP srcGraph)
 {
-    graphExtensionP next = NULL, newNext = NULL, newLast = NULL;
+    graphExtensionP dstExtension = NULL, srcExtension = NULL;
+    graphExtensionP srcExtensionIDMap[MAXNUMSUPPORTEDEXTENSIONS+1];
```
wbkboyer, line 467: (Just for my own comprehension) The `srcExtensionIDMap` is 1-indexed because the `moduleIDGenerator` starts off at 0, but then gets incremented as soon as you attach an extension to a graph, and so that very first extension's `moduleID` is 1.

c/graphLib/extensionSystem/graphExtensions.c
```diff
@@ -459,39 +458,51 @@ graphExtensionP _FindNearestOverload(graphP theGraph, graphExtensionP target, in
...
+    // by the extension module ID. This mapping is needed because the srcGraph
+    // may have been extended with the extensions in a different order than the
+    // dstGraph, but the moduleID for each extension is the same across graphs. 
+    srcExtension = srcGraph->extensions;
+    while (srcExtension != NULL)
     {
-        if ((newNext = (graphExtensionP)malloc(sizeof(graphExtensionStruct))) == NULL)
-        {
-            gp_FreeExtensions(dstGraph);
+        if (srcExtension->moduleID < 0 || srcExtension->moduleID > MAXNUMSUPPORTEDEXTENSIONS)
```
wbkboyer, line 486: Should this be turned into a macro, something like `gp_IsValidModuleID()`, to bounds-check the `moduleID` of the current extension being copied from the the `srcGraph`? (Although glancing at the other sourcefiles, I don't know if this check is really done anywhere else, so might not be useful to encapsulate)

c/graphLib/extensionSystem/graphExtensions.c
```diff
@@ -459,39 +458,51 @@ graphExtensionP _FindNearestOverload(graphP theGraph, graphExtensionP target, in
...
             return NOTOK;
-        }
 
-        newNext->moduleID = next->moduleID;
-        newNext->context = next->dupContext(next->context, dstGraph);
-        newNext->dupContext = next->dupContext;
-        newNext->freeContext = next->freeContext;
-        newNext->functions = next->functions;
-        newNext->next = NULL;
+        srcExtensionIDMap[srcExtension->moduleID] = srcExtension;
```
wbkboyer, line 489: Neat little trick to map this correspondence by `moduleID`, so that no matter the order in which the extensions were attached to the `srcGraph` vs. the order in which they were (optionally already) attached to the `dstGraph`, you'll be able to easily do this lookup!!! 🚀

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -141,8 +141,7 @@ int testAllGraphs(char command, char modifier, char const *const infileName, tes
         return NOTOK;
     }
 
-    if (ExtendGraph(origGraphRead, command) != OK ||
-        ExtendGraph(graphForEmbedding, command) != OK)
+    if (ExtendGraph(graphForEmbedding, command) != OK)
```
wbkboyer, line 144: Very cool to see the `ExtendGraph()` on the `origGraphRead` (i.e. the one associated with the `G6ReadIterator`) go away, since that `gp_CopyGraph()` in the loop after reading the next graph is now going to see that there's no extensions attached to the `origGraphRead`, and will simply reinitialize the structures associated with the extension attached to the `graphForEmbedding`!!!!

wbkboyer (approved):
This looks great to me! 🚀

Issue #207, opened 2026-04-27
Write EAPS style guide


wbkboyer:
## Background

During the course of #206 some suggestions were made to adjust the style of the sourcecode, including:
* Remove whitespace from error handling blocks that separate `ErrorMessage()` from cleanup and return statement
* Group allocation and initialization calls for `graphP` and `G6(Read|Write)Iterator`s (taking advantage of short-circuit evaluation) with more generic error messaging to reduce boilerplate 
* When checking the result of some operations, don't assign return value to `Result`; in error case, explicitly `return NOTOK;` or assign `Result = NOTOK` if subsequent cleanup steps need to be performed
    ```
    if (someOperation() != OK)
    {
        ErrorMessage("Some operation failed.\n");
        return NOTOK;
    }
    ```

    or

    ```
    int Result = OK;
    ...
    if (someOperation() != OK)
    {
        ErrorMessage("Some operation failed.\n");
        Result = NOTOK;
    }
    ```
    vs. 
    ```
    int Result = OK;
    ...
    if ((Result = someOperation()) != OK)
    {
        ErrorMessage("Some operation failed.\n");
        return Result;
    }
    ```
* In `graphLib`, when returning `NOTOK`, make sure an `ErrorMessage()` is emitted to indicate the cause of the issue (`QuietMode` will be `TRUE` by default come 5.0.0)
* Curly-braces opening block should appear on the next line, e.g.
    ```
    if (<statement>)
    {
        ...
    }
    ```

## Proposed solution

Create wiki page to capture these patterns so that any contributors (or downstream consumers of the `graphLib` via `libplanarity` or the [`planarity`](https://github.com/graph-algorithms/planarity) Cython wrapper) will maintain consistency.

Issue #188, opened 2025-11-05
Add parameter names to all function declarations that don't have them


john-boyer-phd:
Many header file function declarations include parameter names, but some don't. During initial development, it's true that it can be an inconvenience to have to keep parameter names in header declarations synchronized with parameter names at the start of function implementations, but long after the function has been finished, the developer benefits greatly from the parameter names being available in code hover hints, and that long term benefit outweighs the small short term inconvenience to the initial developer.

The goal of solving this issue is to systematically go through all header files and add the parameter names if they are missing.

WilliamK112:
I would like to take this. I checked current master: the remaining unnamed header parameters are in the graph function-table callbacks. I will match their names to the corresponding implementations, audit the other headers, and verify the change with the build and tests. This will only add parameter names, without changing function types or behavior.

PR #230, opened 2026-05-29
Fine tuning 50

### LINKED ISSUE #223: Minor refactor of quiet mode functions
In `apiutils` (.c/.h), we currently have two methods called `gp_GetQuietModeFlag()` and `gp_SetQuietModeFlag()`, which manipulate an underlying private library variable `quietModeFlag`. It's become evident that we need more than just a flag because we would like to separately control whether to quiet only `gp_ErrorMessage` (versus the current behavior of quieting both `gp_Message` and `gp_ErrorMessage`).

First, we need to change to the following names:

* `unsigned gp_GetQuietMode(void)`
* `void gp_SetQuietMode(unsigned)` 
*  `unsigned quietMode` for the private library variable

Once this compiles and runs, then we need quiet mode macro definitions in `apiutils.h`:

* `QUIETMODE_NONE 0`
* `QUIETMODE_ERRORS 1`
* `QUIETMODE_MESSAGES 2`
* `QUITEMODE_ALL 0XFFFFFFFF`

Then, the following alternative `gp_GetQuietMode()` tests can be performed:

* In `gp_Message()`, test `if (!(gp_GetQuietMode() & QUIETMODE_MESSAGES))`
* In `gp_ErrorMessage()`, test `if (!(gp_GetQuietMode() & QUIETMODE_ERRORS))`
* Both tests in `planarityRandomGraphs.c` should do the test for messages (because they write to `stdout`)

And finally, the `gp_SetQuietMode()` calls should change as follows:

* In `planarity.c`, call `gp_SetQuietMode(QUIETMODE_NONE)`
* In `planarityCommandLine.c`, call  `gp_SetQuietMode(QUIETMODE_ALL)`

To "reality check" the functioning of the revisions, start with changing `gp_SetQuietMode(QUIETMODE_NONE)` to `gp_SetQuietMode(QUIETMODE_ERRORS)` in `planarity.c`. Then, run `planarity` in menu mode (such as from the debugger). Type X for transforming a graph, and then type filenames that don't exist (like `blah`) and choose any format. You should *not* see "Transform Graph Menu emitted an error.".  Then, change back to `gp_SetQuietMode(QUIETMODE_NONE)`, run the same transform, and you should see  "Transform Graph Menu emitted an error.".

To test before submitting a PR, just commit your code to your branch in your fork, then use the Github web interface to go to the Actions menu and select each GitHub action on the left then press the "Run Workflow" button on the right, and select the branch for the code you want to submit as a PR. If these all run correctly with no errors, then you're good to go (or if you have any trouble running them, please drop a comment on this issue to let us know).

john-boyer-phd:
Resolves #223 

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Changes

Minor updates to comments, parameter types, return types and function order. 

Fix for #223 

## Testing

Compiles for gcc and cl, release and debug. Random graphs for planarity and K3,3 search. Test like the one described in #223, except with Specific Graph rather than transform.

### DIFF
--- a/c/graphLib/graphLib.c
+++ b/c/graphLib/graphLib.c
@@ -13,7 +13,7 @@ See the LICENSE.TXT file for licensing information.
  Return full major.minor.maint.tweak version string for the graph planarity project
  ********************************************************************/
 
-char *gp_GetProjectVersionFull(void)
+const char *gp_GetProjectVersionFull(void)
 {
     static char projectVersionStr[MAXLINE + 1];
     sprintf(projectVersionStr, "%d.%d.%d.%d",
@@ -29,7 +29,7 @@ char *gp_GetProjectVersionFull(void)
  Returns full current:revision:age version string for the graph planarity shared library
  ********************************************************************/
 
-char *gp_GetLibPlanarityVersionFull(void)
+const char *gp_GetLibPlanarityVersionFull(void)
 {
     static char libPlanarityVersionStr[MAXLINE + 1];
     sprintf(libPlanarityVersionStr, "%d:%d:%d",
--- a/c/graphLib/homeomorphSearch/graphK23Search_Extensions.c
+++ b/c/graphLib/homeomorphSearch/graphK23Search_Extensions.c
@@ -28,7 +28,7 @@ int _K23Search_CheckObstructionIntegrity(graphP theGraph, graphP origGraph);
 
 /* Forward declarations of functions used by the extension system */
 
-int  _K23Search_CopyData(void *, void *);
+int _K23Search_CopyData(void *, void *);
 void _K23Search_FreeContext(void *);
 
 /****************************************************************************
@@ -176,7 +176,7 @@ int _K23Search_EmbedPostprocess(graphP theGraph, int v, int edgeEmbeddingResult)
             // to ensure post-processing continues as expected.
             savedEmbedFlags = gp_GetEmbedFlags(theGraph);
             savedZEROBASEDIO = gp_GetGraphFlags(theGraph) & GRAPHFLAGS_ZEROBASEDIO;
-            gp_ReinitGraph(theGraph);
+            gp_ResetGraphStorage(theGraph);
             theGraph->embedFlags = savedEmbedFlags;
             theGraph->graphFlags &= savedZEROBASEDIO;
         }
--- a/c/graphLib/io/g6-read-iterator.c
+++ b/c/graphLib/io/g6-read-iterator.c
@@ -40,24 +40,24 @@ int _g6_ReadGraphFromString(graphP theGraph, char *g6EncodedString);
 /********************************************************************
  Package private structure declaration for read iterator
  ********************************************************************/
-    typedef struct strOrFileStruct strOrFileStruct;
-    typedef strOrFileStruct *strOrFileP;
+typedef struct strOrFileStruct strOrFileStruct;
+typedef strOrFileStruct *strOrFileP;
 
-    struct G6ReadIteratorStruct
-    {
-        strOrFileP inputContainer;
-        int numGraphsRead;
+struct G6ReadIteratorStruct
+{
+    strOrFileP inputContainer;
+    int numGraphsRead;
 
-        int order;
-        int numCharsForOrder;
-        size_t numCharsForGraphEncoding;
-        size_t currGraphBuffSize;
-        char *currGraphBuff;
+    int order;
+    int numCharsForOrder;
+    size_t numCharsForGraphEncoding;
+    size_t currGraphBuffSize;
+    char *currGraphBuff;
 
-        graphP currGraph;
+    graphP currGraph;
 
-        bool endReached;
-    };
+    bool endReached;
+};
 
 /********************************************************************
  Public and package private method implementations for read iterator
@@ -404,7 +404,7 @@ int _g6_InitReader(G6ReadIteratorP theG6ReadIterator)
 
     if (gp_GetN(theG6ReadIterator->currGraph) == 0)
     {
-        if (gp_InitGraph(theG6ReadIterator->currGraph, order) != OK)
+        if (gp_EnsureVertexCapacity(theG6ReadIterator->currGraph, order) != OK)
         {
             gp_ErrorMessage("Unable to initialize reader due to failure "
                             "initializing graph datastructure with order %d for "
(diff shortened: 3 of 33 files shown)


c/graphLib/homeomorphSearch/graphK23Search_Extensions.c
```diff
@@ -28,7 +28,7 @@ int _K23Search_CheckObstructionIntegrity(graphP theGraph, graphP origGraph);
 
 /* Forward declarations of functions used by the extension system */
 
-int  _K23Search_CopyData(void *, void *);
+int _K23Search_CopyData(void *, void *);
```
wbkboyer, line 31: Should this be updated with the parameter names, i.e. `int _K23Search_CopyData(void *dstContext, void *srcContext);`? (see #188 )

c/graphLib/io/g6-read-iterator.c
```diff
@@ -40,24 +40,24 @@ int _g6_ReadGraphFromString(graphP theGraph, char *g6EncodedString);
 /********************************************************************
  Package private structure declaration for read iterator
  ********************************************************************/
-    typedef struct strOrFileStruct strOrFileStruct;
-    typedef strOrFileStruct *strOrFileP;
+typedef struct strOrFileStruct strOrFileStruct;
```
wbkboyer, line 43: Good call to de-indent these type declarations!

c/graphLib/io/g6-read-iterator.c
```diff
@@ -310,7 +310,7 @@ int g6_InitReaderWithFileName(G6ReadIteratorP theG6ReadIterator, char const *con
     if ((inputContainer = sf_NewInputContainer(NULL, infileName)) == NULL)
     {
         gp_ErrorMessage(
-            "Unable to initialize reader with filename, as we failed to "
+            "Unable to initialize reader with file name, as we failed to "
```
wbkboyer, line 313: I noticed you change "filename" to "file name", so I checked out your branch and made sure that there were no errant `\bfilename\b` kicking around; I'll try to remember this naming convention change (e.g. use `fileName` and not `filename`, `FileName`, nor `Filename`) Maybe add variable naming conventions to #207 ?

c/graphLib/lowLevelUtils/appconst.h
```diff
@@ -62,7 +62,7 @@ extern int debugNOTOK(void);
 #endif
 
 #ifndef NULL
-#define NULL 0L
+#define NULL ((void *)0)
```
wbkboyer, line 65: Great catch, as per our discussion!!

Related StackOverflow [response](https://stackoverflow.com/a/2599766/5456879)

c/graphLib/graphLib.c
```diff
@@ -13,7 +13,7 @@ See the LICENSE.TXT file for licensing information.
  Return full major.minor.maint.tweak version string for the graph planarity project
  ********************************************************************/
 
-char *gp_GetProjectVersionFull(void)
+const char *gp_GetProjectVersionFull(void)
```
wbkboyer, line 16: Nice const correctness catches!

c/planarityApp/planarityUtils.c
```diff
@@ -856,25 +852,18 @@ char *ConstructInputFilename(char const *infileName)
...
 
         if (Result == OK)
         {
-            if (strcpy(theFileName, infileName) == NULL)
-            {
-                gp_ErrorMessage("Copying infileName into theFileName using strcpy() failed.\n");
-
-                Result = NOTOK;
-            }
+            strcpy(theFileName, infileName);
```
wbkboyer, line 866: One wart I notice is that I use `theFileName` within the scope of a lot of functions in the `c/planarityApp` layer (trying to avoid global variables, e.g. in `RandomGraphs()`) Is there any issue that can arise from this?

c/planarityApp/planarityUtils.c
```diff
@@ -961,7 +950,7 @@ char *ConstructPrimaryOutputFilename(char const *infileName, char const *outfile
  * Returns non-NULL string
  ****************************************************************************/
 
-int ConstructTransformationExpectedResultFilename(char const *infileName, char **outfileName, char command, int baseFlag)
+int ConstructTransformationExpectedResultFileName(char const *infileName, char **outfileName, char command, int baseFlag)
```
wbkboyer, line 953: Should we change all instances of `infileName` to `inputFileName`, and likewise `outfileName` to `outputFileName`?

wbkboyer (commented):
Looking great so far! Just have a couple minor questions, but then should be good to merge!!!

c/graphLib/homeomorphSearch/graphK23Search_Extensions.c
john-boyer-phd: I would have preferred to include the names, but the function itself was included for consistency with the other extensions; however, the function implementation doesn't use the parameters because there is no data to copy. I was concerned that one of our beyond-pedantic-warnings-as-errors would get annoyed that we named a parameter without using it, so I left them out.

c/graphLib/io/g6-read-iterator.c
john-boyer-phd: I think the autoformatting made that choice.

c/graphLib/io/g6-read-iterator.c
john-boyer-phd: There's a lot of momentum in the world behind filename/Filename rather than file name/FileName, but we were doing capital N (which means space in english) about 80% of the time, and I personally think of it as the name of the file. For example, similar variables might be called fileSize. And we even have a place where we talk about doing something to the content of a variable containing a file name before we "write the file's name..." It seems best not to let tech invent yet another word when not needed.

c/planarityApp/planarityUtils.c
john-boyer-phd: I've kind of been doing infile and therefore infileName, and outfile and therefore outfileName, since like the 80s, man 🤣Let's leave them.

c/planarityApp/planarityUtils.c
wbkboyer: We'll circle back to getting rid of `theFileName` global variable (requires refactoring `ConstructInputFileName()` and `ConstructPrimaryOutputFileName()` to `malloc()` sufficient space for the constructed filename, and then requires refactoring of callers of these functions so that they must free the returned `char *`)

wbkboyer (approved):
LET'S FREAKIN GOOOOOO 🚀

PR #234, opened 2026-06-04
Fine tuning 50

john-boyer-phd:
* Move Planarity Graph and DFSUtils Graph data structures from the public graph structure to a private data structure within the graph so that they can be moved elsewhere later while preserving backwards compatibility.
* Added an alternative `testAllGraphsN8()` method to facilitate performance testing of the above.
* Add planar graph drawing methods that deliver visibility representation.

### DIFF
--- a/c/graphLib/graph.h
+++ b/c/graphLib/graph.h
@@ -442,36 +442,6 @@ extern "C"
     typedef struct graphFunctionTableStruct graphFunctionTableStruct;
     typedef graphFunctionTableStruct *graphFunctionTableP;
 
-    // DFS-RELATED and PLANARITY-RELATED ONLY
-
-    // Declaration of package-private data type for managing a
-    // collection of lists of integers
-    typedef struct listCollectionStruct listCollectionStruct;
-    typedef listCollectionStruct *listCollectionP;
-
-    // Declaration of package-private data type for managing additional 
-    // DFS--related information associated with each non-virtual vertex
-    typedef struct DFSUtils_VertexInfo DFSUtils_VertexInfo;
-    typedef DFSUtils_VertexInfo *DFSUtils_VertexInfoP;
-
-    // PLANARITY-RELATED ONLY
-
-    // Declaration of package-private data type for managing additional 
-    // planarity-related information associated with each non-virtual vertex
-    typedef struct Planarity_VertexInfo Planarity_VertexInfo;
-    typedef Planarity_VertexInfo *Planarity_VertexInfoP;
-
-    // Declaration of package private data type for optimizing management of
-    // the external face of a planar embedding as it is being built
-    typedef struct extFaceLinkRec extFaceLinkRec;
-    typedef extFaceLinkRec *extFaceLinkRecP;
-
-    // PLANARITY-RELATED ONLY
-    // Declaration of package private data type for isolating
-    // minimal subgraphs obstructing planarity-related embedding
-    typedef struct isolatorContextStruct isolatorContextStruct;
-    typedef isolatorContextStruct *isolatorContextP;
-
     /********************************************************************
          Graph structure definition
                 V : Array of vertex records (allocated size N + NV)
@@ -516,20 +486,12 @@ extern "C"
         stackP theStack;
 
         // Provides ability to subclass the base Graph, 
-        // including virtual function overlaods by subclasses
+        // including virtual function overloads by subclasses
         graphExtensionP extensions;
         graphFunctionTableP functions;
 
         void *privateData;
- 
-        listCollectionP BicompRootLists;
-        //DFSUtils_VertexInfoP DVI;
-
-        //Planarity_VertexInfoP PVI;
-        //listCollectionP sortedDFSChildLists;
-        //extFaceLinkRecP extFace;
-        //isolatorContextP IC;
-   };
+    };
 
     typedef struct graphStruct graphStruct;
     typedef graphStruct *graphP;
--- a/c/graphLib/graph.h
+++ b/c/graphLib/graph.h
@@ -462,12 +462,14 @@ extern "C"
 
                 extensions: an object-oriented hierarchy of graph classes is implemented
                             manually as a list of extensions for data of any subclasses
-                            with which a graph has been extended. 
-                functions: the object-oriented hierarchies includes virtual function 
-                           overloading, which is provided by a function pointer table.
+                            with which a graph has been extended.
+                extensionLookupTable: if not NULL, keeps an array of pointers to the 
+                                      extensions indexed by ID for constant-time lookup.
+                functions: object-oriented class hierarchies includes virtual function 
+                           overloading, which is provided by this function pointer table.
 
-                privateData: pointer to structure for package private data that can change
-                             while maintaining backwards compatibility.
+                privateData: pointer to a package private data structure that can be
+                             backwards compatibly changed in releases of graphLib.
         */
 
     struct graphStruct
@@ -488,6 +490,7 @@ extern "C"
         // Provides ability to subclass the base Graph, 
         // including virtual function overloads by subclasses
         graphExtensionP extensions;
+        graphExtensionP *extensionLookupTable;
         graphFunctionTableP functions;
 
         void *privateData;
--- a/c/graphLib/graph.h
+++ b/c/graphLib/graph.h
@@ -527,7 +527,7 @@ extern "C"
 
         Planarity_VertexInfoP PVI;
         listCollectionP sortedDFSChildLists;
-        extFaceLinkRecP extFace;
+        //extFaceLinkRecP extFace;
         //isolatorContextP IC;
    };
 
--- a/c/graphLib/graph.h
+++ b/c/graphLib/graph.h
@@ -523,9 +523,9 @@ extern "C"
         void *privateData;
  
         listCollectionP BicompRootLists;
-        DFSUtils_VertexInfoP DVI;
+        //DFSUtils_VertexInfoP DVI;
 
-        // Planarity_VertexInfoP PVI;
+        //Planarity_VertexInfoP PVI;
         //listCollectionP sortedDFSChildLists;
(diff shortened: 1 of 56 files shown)


c/graphLib/graph.private.h
```diff
@@ -18,6 +18,81 @@ extern "C"
...
+
+        PVI: package private pointer; if the graph is extended to Planarity,
+                then N instances of Planarity vertex info records are allocated
+        IC: contains additional useful variables for Kuratowski subgraph isolation.
+        sortedDFSChildLists: for Planarity graphs, storage for the sorted DFS child
+                lists of each vertex
+        extFace: For Planarity graphs, an array of (N + NV) external face
+                short circuit records
+     ********************************************************************/
+    struct graphPrivateDataStruct
```
wbkboyer, line 68: Very freakin' snazzy, encapsulating all these in a private `struct`... Is there a reason why the `privateData` in the `graphStruct` is a `void *` and not a `graphPrivateDataP`? I see in `gp_New()`:

```
    graphPrivateDataP theGraphPrivateData = (graphPrivateDataP)calloc(1, sizeof(graphPrivateDataStruct));
...
        theGraph->privateData = (void *)theGraphPrivateData;
```

Is this because you don't want to have to expose `graphPrivateDataP` in `graph.h`?

wbkboyer: @john-boyer-phd replied in private correspondence:

> Good question, since `stackP` is declared as a pointer to an undefined struct, so that could happen for `graphPrivateDataP` too. The reason for `void *` is that someday everything in `graphPrivateDataStruct` will instead be in `Planarity` or `DFSUtils` extension `struct`, at which point `graphPrivateDataStruct` will have to be deleted, until the next time it is necessary to temporarily extend the graph while preserving backwards compatibility.

c/graphLib/graph.h
```diff
@@ -522,19 +484,16 @@ extern "C"
...
-
-        Planarity_VertexInfoP PVI;
-        listCollectionP sortedDFSChildLists;
-        extFaceLinkRecP extFace;
-        isolatorContextP IC;
 
+        // Provides ability to subclass the base Graph, 
+        // including virtual function overloads by subclasses
         graphExtensionP extensions;
+        graphExtensionP *extensionLookupTable;
```
wbkboyer, line 493: Is this the array to enable constant time lookup to determine whether a graph is extended with an extension having been assigned a `moduleID`? i.e. one uses the `moduleID` as the index, the corresponding element is the pointer to the extension, so that one doesn't have to use `gp_FindExtension()` to iterate over all extensions of the graph?

I don't see `extensionLookupTable` being referenced anywhere yet, e.g. in `gp_AddExtension()` after you update the `extensions` pointer, and I still see calls to `gp_FindExtension()` in the `gp_ExtendWith_XXX()` functions and in `gp_AddExtension()`!

c/graphLib/planarityRelated/graphDrawPlanar.h
```diff
@@ -22,7 +22,15 @@ extern "C"
     int gp_DrawPlanar_RenderToFile(graphP theEmbedding, char *theFileName);
     int gp_DrawPlanar_RenderToString(graphP theEmbedding, char **pRenditionString);
 
-#ifdef __cplusplus
+    int gp_DrawPlanar_GetVertexPosition(graphP theEmbedding, int v);
```
wbkboyer, line 25: Right onnnn, these functions can be called from the [`planarity.classic`](https://github.com/graph-algorithms/planarity/blob/master/planarity/classic) subpackage of the [`planarity`](https://pypi.org/project/planarity/) EAPS Cython wrapper, specifically in the [`nodes()`](https://github.com/graph-algorithms/planarity/blob/c9bc67b8044a630e33200cbcd776e803040ec3d0/planarity/classic/planarity.pyx#L108) and [`edges()`](https://github.com/graph-algorithms/planarity/blob/c9bc67b8044a630e33200cbcd776e803040ec3d0/planarity/classic/planarity.pyx#L133) methods of the [`PGraph`](https://github.com/graph-algorithms/planarity/blob/c9bc67b8044a630e33200cbcd776e803040ec3d0/planarity/classic/planarity.pyx#L12) class!

I also notice these are accessor functions rather than macros; would there be any benefit to turning these into macros to be used in `graphDrawPlanar.c` (e.g. in `_CheckVisibilityRepresentationIntegrity()` when comparing values)?

c/graphLib/planarityRelated/graphDrawPlanar.c
```diff
@@ -1091,3 +1091,155 @@ int _CheckVisibilityRepresentationIntegrity(DrawPlanarContext *context)
...
+    DrawPlanarContext *context = NULL;
+    gp_FindExtension(theEmbedding, DRAWPLANAR_ID, (void *)&context);
+
+    if (theEmbedding == NULL || context == NULL ||
+        context->VI == NULL ||
+        v < gp_LowerBoundVertices(theEmbedding) ||
+        v >= gp_UpperBoundVertices(theEmbedding))
+    {
+        gp_ErrorMessage("Invalid parameters");
+        return -1;
```
wbkboyer, line 1114: Should this (and all other instances in `graphDrawPlanar.c`) return `NIL` (`0` if `USE_1BASEDARRAYS`, `-1` otherwise) instead of hard-coded to `-1`? (If so, this would require adding `NIL` back to the `planarity` Cython wrapper to test the return value)

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -327,3 +337,195 @@ int outputTestAllGraphsResults(char command, char modifier, testAllStatsP stats,
...
+        for (lineNum = 0; lineNum <= NUM8VERTEXTGRAPHS; lineNum++)
+            gp_Free(&origGraphRead[lineNum]);
+        gp_Free(&graphForEmbedding);
+        stats->errorFlag = TRUE;
+        return NOTOK;
+    }
+
+    // Read the graphs into memory
+    lineNum = 0;
+    while (lineNum <= NUM8VERTEXTGRAPHS)
```
wbkboyer, line 434: Should this loop be `while(TRUE)` and then check after the loop that we did indeed read `NUM8VERTEXTGRAPHS` graphs, i.e. check `(lineNum - 1) == NUM8VERTEXTGRAPHS` and emit error if not?

c/planarityApp/planarityTestAllGraphs.c
```diff
@@ -327,3 +337,195 @@ int outputTestAllGraphsResults(char command, char modifier, testAllStatsP stats,
...
+    // the end of the input, then the number of graphs successfully read is
+    // lineNum - 1.
+    stats->numGraphsTested = lineNum - 1;
+
+    // Start a timer, and we will iterate the work many times to
+    // get a more reliable reading.
+
+    platform_GetTime(start);
+
+    for (int i = 0; i < 1000; ++i)
```
wbkboyer, line 476: This is a nice way to factor out the cost of IO when testing!! It'd be super neat to add a parameter for number of times to run through all graphs with N=8 (rather than hard-coded to 1000), but that would require this whole machinery to be refactored to make that available at runtime 😭

c/graphLib/graph.h
wbkboyer: @john-boyer-phd replied in private correspondence:

> This is for the future, not the `5.0.0.0` release. We need it here now because it means there will be no changes to the `graphStruct` in the future when it is used [i.e. no changes to API requiring `graphLib` version bump]. It will help solve some problems related to moving stuff out of `graphPrivateStruct`, and will probably make copy extensions even faster as a nice side effect.

c/graphLib/planarityRelated/graphDrawPlanar.h
wbkboyer: From @john-boyer-phd 

> We can’t turn these into macros because then we’d have to make public the data structures they interrogate.

(Which absolutely makes sense due to inline replacement, no doyyyyyyy)

c/graphLib/planarityRelated/graphDrawPlanar.c
wbkboyer: From @john-boyer-phd 

> Where I landed there was that the values are not vertex/edge array index values but rather spatial values, so best to return a specific out of range value but not `NIL`  (which is for vertices and edges).

c/planarityApp/planarityTestAllGraphs.c
wbkboyer: From @john-boyer-phd :

> What would be ideal is `while (TRUE)` and then no error check after. My secondary hope in writing this is that it would be generalizable, so it would be a `while (TRUE)` but then grow the array by doubling until memory failure, or maybe I still have to hard code a big number. Either way, right now the `origGraphRead` array is hard coded to a size, so we can’t keep reading until file end and check for overrun after because it would `segfault` before the [`g6_EndReached()`] check (if given a file that was named `n8.mALL.g6` but really had more graphs). 

---

If we wanted to make this more flexible I think we'd have to change the function signature to include number of times to run the test and the expected number of graphs in the file, then to error out if the number of lines/graphs in the file doesn't match the expected number of graphs in the file. However, this code was added "to test speed on graphs of all shapes as [he] made all those private data changes [associated with this PR]. And when [he] change[s] it again, it’ll be useful to uncomment that `#define` to see if the new full extension way of doing `Planarity` and `DFSUtils` has a big performance hit."

wbkboyer (changes requested):
Looks really clean! All the questions I had were resolved, although the one issue you brought up is that `gp_DupGraph()` no longer duplicates extensions.

wbkboyer (approved):
Did some additional testing using ASAN (MacOS clang, Windows MSVC cl, Debian gcc) + LSAN (Debian gcc) using the `testAllGraphsN8()` function (compiled with `-fsanitize=address` and `-DTESTALLGRAPHS_MEMORY_TIMING_TEST`, `/f`/`/D` for Windows) and detected no errors! 🎉

PR #264, opened 2026-07-06
Remediate DFSUtils macro accessors

### LINKED ISSUE #261: Remediation for three DFSUtils macro methods
## Describe the bug

While finalizing the planarity Python package for 1.0.0 release, we noticed three DFSUtils macro methods that require remediation.

The methods `gp_IsDFSTreeRoot()`, `gp_IsNotDFSTreeRoot()`, and `gp_GetVertexFromBicompRoot()` in `graphDFSUtils.h` are successfully used in a number of places within the graphLib to implement key algorithms. However, from within the graphLib, source files also include `graphDFSUtils.private.h` to gain access to non-public speedy (inline) macros for several other methods.

The problem is that the macro implementations of these three methods are invoking a package-private macro called `gp_GetVertexParent()` rather than the public API function `gp_GetParent()`. As a result, application programs that including the public APIs only via including `graphLib.h` cannot use these three methods because nothing in the public API defines `gp_GetVertexParent()`. Such programs would not link.   

## Required solution

The solution is almost as simple as replacing `gp_GetVertexParent()` with `gp_GetParent()` in the three macro implementations.

However, doing so would introduce function calls into some places in planarity-related algorithm implementations that currently use speedier macros.  The first rule of Edge Addition Planarity is to not decrease the implementation speed of the core planarity-related algorithms, so here's what is needed instead.

1. Place copies of `gp_IsDFSTreeRoot()`, `gp_IsNotDFSTreeRoot()`, and `gp_GetVertexFromBicompRoot()` into `graphDFSUtils.private.h` and precede each with an underscore before the `gp` (which makes them package private versions).
2. Comment out the occurrences of `gp_IsDFSTreeRoot()`, `gp_IsNotDFSTreeRoot()`, and `gp_GetVertexFromBicompRoot()` in the public `graphDFSUtils.h` file and then work to get the project to compile by replacing all occurrences of invocations of the three methods with their underscore veresion.
3. Uncomment the three methods in the public `graphDFSUtils.h` and replace `gp_GetVertexParent()` with `gp_GetParent()` in each.

## Testing

In `planaritySpecificGraph.c` you can temporarily inject lines of code that are called just after `gp_Embed()` and that just print the results of invoking `gp_IsDFSTreeRoot()` on the first vertex, `gp_IsNotDFSTreeRoot()` on the second vertex, and `gp_GetVertexFromBicompRoot()` on vertex N+1. If the program compiles and links, then you are successful. Cut the temporary injection of the three calls and paste them into your PR submission.

WilliamK112:
## Summary

Fixes #261.

This updates the public DFSUtils macros to use the public `gp_GetParent()` accessor instead of the package-private `gp_GetVertexParent()` macro. For graphLib internals, this adds private underscore-prefixed macro copies and updates internal call sites to use those private versions.

The private `_gp_GetVertexFromBicompRoot()` macro computes the DFS child location directly, so `graphDFSUtils.private.h` remains usable without depending on the public DFSUtils header macro.

## Tests

- `git diff --check`
- Public-header syntax smoke check using `c/graphLib/graphLib.h` with `-Werror=implicit-function-declaration`
- Private-header syntax smoke check using `c/graphLib/graphDFSUtils.private.h` with `-Werror=implicit-function-declaration`
- `/tmp/eaps-planarity-check/planarity.exe -test ./c/samples/`

## Safety

This keeps the existing internal behavior while avoiding private accessor leakage through the installed public header. The change is limited to DFSUtils macro boundaries and internal graphLib call sites.

### DIFF
--- a/c/graphLib/graphDFSUtils.private.h
+++ b/c/graphLib/graphDFSUtils.private.h
@@ -52,6 +52,10 @@ extern "C"
 #define gp_GetVertexParent(theGraph, v) (theGraphDVI(theGraph)[v].parent)
 #define gp_SetVertexParent(theGraph, v, theParent) (theGraphDVI(theGraph)[v].parent = theParent)
 
+#define _gp_IsDFSTreeRoot(theGraph, v) gp_IsNotVertex(theGraph, gp_GetVertexParent(theGraph, v))
+#define _gp_IsNotDFSTreeRoot(theGraph, v) gp_IsVertex(theGraph, gp_GetVertexParent(theGraph, v))
+#define _gp_GetVertexFromBicompRoot(theGraph, R) gp_GetVertexParent(theGraph, (R) - gp_GetN(theGraph))
+
 #define gp_GetVertexLeastAncestor(theGraph, v) (theGraphDVI(theGraph)[v].leastAncestor)
 #define gp_SetVertexLeastAncestor(theGraph, v, theLeastAncestor) (theGraphDVI(theGraph)[v].leastAncestor = theLeastAncestor)
 
--- a/c/graphLib/graphDFSUtils.private.h
+++ b/c/graphLib/graphDFSUtils.private.h
@@ -8,6 +8,7 @@ See the LICENSE.TXT file for licensing information.
 #define GRAPHDFSUTILS_PRIVATE_H
 
 #include "graph.private.h"
+#include "graphDFSUtils.h"
 
 #ifdef __cplusplus
 extern "C"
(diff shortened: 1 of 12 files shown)


c/graphLib/graphDFSUtils.private.h
```diff
@@ -52,6 +52,10 @@ extern "C"
 #define gp_GetVertexParent(theGraph, v) (theGraphDVI(theGraph)[v].parent)
 #define gp_SetVertexParent(theGraph, v, theParent) (theGraphDVI(theGraph)[v].parent = theParent)
 
+#define _gp_IsDFSTreeRoot(theGraph, v) gp_IsNotVertex(theGraph, gp_GetVertexParent(theGraph, v))
+#define _gp_IsNotDFSTreeRoot(theGraph, v) gp_IsVertex(theGraph, gp_GetVertexParent(theGraph, v))
+#define _gp_GetVertexFromBicompRoot(theGraph, R) gp_GetVertexParent(theGraph, (R) - gp_GetN(theGraph))
```
john-boyer-phd: One small change request here. Please can you change this to call `gp_GetDFSChildFromBicompRoot()` here (which means also adding the public DFSUtils header include).

It's no worse because the package-private is allowed to consume the public level; it's just the public level that shouldn't access the package-private stuff (which you are fixing, with thanks).

The reasons it is better, though, are explanatory power and future-proofing. The method name says _what_ the `R - N ` is doing. Moreover, it is possible that one day the virtual vertex allocation method will be generalized, replacing the current static allocation and assignment. If/when that happens, the mechanism by which we get a DFS child from a bicomp root might change from the current `R - N` method. Calling `gp_GetDFSChildFromBicompRoot()` means that `_gp_GetVertexFromBicompRoot()` is insulated from that possible future change.

john-boyer-phd (changes requested):
Excellent work, one small change request please?

john-boyer-phd (approved):
Excellent, thank you once again William ^_^

PR #265, opened 2026-07-06
Allow extension immediately after graph creation

### LINKED ISSUE #236: Enable calling `gp_ExtendWith...()` functions right after `gp_New()`
In each of the 7 `gp_ExtendWith_Subclass()` functions, change the first line of code (after variable declarations) from:

`    if (theGraph == NULL || gp_GetN(theGraph) <= 0)
`
to:

`    if (theGraph == NULL)
`

Also, in the file `planarityUtils.c`, change the first lines of the function `ExtendGraph()` from

```
    if (theGraph == NULL || theGraph->N <= 0)
    {
        gp_ErrorMessage("Unable to extend graph with algorithm extension due "
                        "to NULL or uninitialized graph.");
```

to

```
    if (theGraph == NULL)
    {
        gp_ErrorMessage("Unable to extend graph with algorithm extension due to NULL graph.");
```

The reason for these changes is that the subsequent code of all graph subclass extensions is already set up to handle the case in which the number of vertices _N_ has not yet been set by `gp_EnsureVertexCapacity()`, so there is no need to cause the failure of those  `gp_ExtendWith_Subclass()` functions.

In fact, although `gp_ExtendWith_Subclass()` functions can be called and should be callable after `gp_EnsureVertexCapacity()` (because the graph library is supporting run-time subclassing), it is conceptually useful to place the `gp_ExtendWith_Subclass()` function invocation directly after `gp_New()` (before `gp_EnsureVertexCapacity()`) because the pair of invocations of `gp_New()` and `gp_ExtendWith_Subclass()` are then analogous to declaring a new instance of a subclass in a single line of code as would occur in C++, Java, or other languages with compile-time subclassing.

To ensure that this new order of invocations is always tested going forward, make the following two changes to the `testAllGraphs()` function. First change:

```
    if ((graphForEmbedding = gp_New()) == NULL ||
        gp_EnsureVertexCapacity(graphForEmbedding, order) != OK)

```
to:

```
    if ((graphForEmbedding = gp_New()) == NULL ||
        ExtendGraph(graphForEmbedding, command) != OK)
```

Second, change the code that, before the immediately preceding change, did say :

```
    if (ExtendGraph(graphForEmbedding, command) != OK)
    {
        gp_ErrorMessage("Unable to extend graph for embedding operation.");
```

to:

```
    if (gp_EnsureVertexCapacity(graphForEmbedding, order) != OK)
    {
        gp_ErrorMessage("Unable to expand graph storage for expected number of vertices.");
```

 Finally, note that there are other locations that could be changed to match the above. However, the order of invocations in `SpecificGraph()` and in `MakeGraph()` should be left as-is so that we also continue to also test the alternative order with specific graphs (also invoked during `-test`) and during random graphs testing.

WilliamK112:
## Summary

Fixes #236.

This allows the graph extension functions to be called immediately after `gp_New()` by removing the `N > 0` precondition from the seven `gp_ExtendWith...()` entry points and from the planarity app's `ExtendGraph()` helper.

The `testAllGraphs()` path now extends `graphForEmbedding` before calling `gp_EnsureVertexCapacity()`, so `planarity -test` covers the new supported call order while other paths continue to exercise the existing order.

## Tests

- `git diff --check`
- `/tmp/eaps-planarity-check-236/planarity.exe -test ./c/samples/`
- API smoke test calling each `gp_ExtendWith...()` immediately after `gp_New()` and before capacity expansion

## Safety

The extension initialization paths already defer vertex-capacity-dependent allocation to later operations. This change only removes the early `N <= 0` rejection and adds test coverage for the intended order.

### DIFF
--- a/c/graphLib/graphDFSUtils.c
+++ b/c/graphLib/graphDFSUtils.c
@@ -40,7 +40,7 @@ extern void _ClearVertexVisitedFlags(graphP theGraph, int);
 
 int gp_ExtendWith_DFSUtils(graphP theGraph)
 {
-    if (theGraph == NULL || gp_GetN(theGraph) <= 0)
+    if (theGraph == NULL)
         return NOTOK;
 
     // if the Graph has already been extended with DFS Utils,
--- a/c/graphLib/homeomorphSearch/graphK23Search_Extensions.c
+++ b/c/graphLib/homeomorphSearch/graphK23Search_Extensions.c
@@ -51,7 +51,7 @@ int gp_ExtendWith_K23Search(graphP theGraph)
 {
     K23SearchContext *context = NULL;
 
-    if (theGraph == NULL || gp_GetN(theGraph) <= 0)
+    if (theGraph == NULL)
         return NOTOK;
 
     // If the K2,3 search feature has already been attached to the graph
--- a/c/graphLib/homeomorphSearch/graphK33Search_Extensions.c
+++ b/c/graphLib/homeomorphSearch/graphK33Search_Extensions.c
@@ -72,7 +72,7 @@ int gp_ExtendWith_K33Search(graphP theGraph)
 {
     K33SearchContext *context = NULL;
 
-    if (theGraph == NULL || gp_GetN(theGraph) <= 0)
+    if (theGraph == NULL)
         return NOTOK;
 
     // If the K3,3 search feature has already been attached to the graph,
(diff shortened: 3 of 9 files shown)


john-boyer-phd (approved):
Very helpful to have this cleaned up, thank you William.

PR #266, opened 2026-07-06
Add dynamic edge insertion

### LINKED ISSUE #242: Add a function `gp_DynamicInsertEdge()`
Currently, the function `gp_AddEdge()` adds an edge unless the edge capacity limit has been reached, and the function `gp_DynamicAddEdge()` expands the edge capacity if necessary and then adds the edge.

The goal of this issue is to add the analogous function `gp_DynamicInsertEdge()` corresponding to the current function `gp_InsertEdge()`. If the graph's number of edges is less than the edge capacity limit, then `gp_DynamicInsertEdge()` should just perform a `gp_InsertEdge()`. Otherwise, `gp_DynamicInsertEdge()` should first expand the edge capacity in exactly the same way that it is expanded in `gp_DynamicAddEdge()` and then it should perform the `gp_InsertEdge()`.

In either case, the function `gp_DynamicInsertEdge()` must insert the edge by calling `gp_InsertEdge()`.

As an easy way to test, go to `graph.c` and amend `gp_CreateRandomGraph()` as follows:
- in the `_GetRandomNumber()` call, multiply `theGraph->edgeCapacity` by 5 to ensure that dynamic growth will occur.
- replace the last `gp_AddEdge()` call with `gp_DynamicInsertEdge()` to add the edges instead.

Then, run the planarity application in menu mode and run planarity on 100,000 graphs of 50 vertices each. Once this produces no errors, you are ready for PR. Undo your changes to `gp_CreateRandomGraph()`, but copy-paste your `gp_DynamicInsertEdge()` call into the testing section of your PR submission.

WilliamK112:
## Summary

Fixes #242.

This adds `gp_DynamicInsertEdge()` as the dynamic-capacity counterpart to `gp_InsertEdge()`. The new function first calls `gp_InsertEdge()` directly, and only expands edge capacity when that call returns `AT_EDGE_CAPACITY_LIMIT`; after expanding, it retries insertion through `gp_InsertEdge()`.

The public graph header now declares the new API and the capacity-limit comment mentions both dynamic edge helpers.

## Tests

- `git diff --check`
- API smoke test that sets edge capacity to 1, confirms `gp_InsertEdge()` reaches `AT_EDGE_CAPACITY_LIMIT`, then verifies `gp_DynamicInsertEdge()` expands capacity and inserts successfully
- `/tmp/eaps-planarity-check-242/planarity.exe -test ./c/samples/`

## Safety

The expansion logic matches `gp_DynamicAddEdge()`, and the insertion path always goes through `gp_InsertEdge()` as requested.

### DIFF
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -2041,6 +2041,53 @@ int gp_DynamicAddEdge(graphP theGraph, int u, int ulink, int v, int vlink)
 }
 
 /********************************************************************
+ gp_DynamicInsertEdge()
+ Refer to documentation for gp_InsertEdge() for parameter description.
+
+ Calls gp_InsertEdge(); if AT_EDGE_CAPACITY_LIMIT, doubles the edge
+ capacity using gp_EnsureEdgeCapacity(), then retries gp_InsertEdge().
+
+ Returns OK on success, NOTOK on failure.
+ ********************************************************************/
+int gp_DynamicInsertEdge(graphP theGraph, int u, int e_u, int e_ulink,
+                         int v, int e_v, int e_vlink)
+{
+    int Result = OK;
+
+    Result = gp_InsertEdge(theGraph, u, e_u, e_ulink, v, e_v, e_vlink);
+
+    if (Result == AT_EDGE_CAPACITY_LIMIT)
+    {
+        // The candidate edge capacity is double the current capacity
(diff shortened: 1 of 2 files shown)


john-boyer-phd (approved):
Nice attention to detail catching the need to amend the AT_EDGE_CAPACITY_LIMIT  comment in graph.h.

PR #267, opened 2026-07-06
Add hide/restore regression test

### LINKED ISSUE #249: Add a Regression Test for Hide/Restore Vertices/Edges APIs
The goal of this issue is to add to what happens when running `planarity -test` (which is run very frequently). The new code will directly run the methods `gp_HideVertex()` and `gp_RestoreVertices()`, which will implicitly also run  `gp_RestoreVertex()`, `gp_HideEdge()`, and `gp_RestoreEdge()`. The following steps are needed:

- [ ] In `planarityCommandLine.c`, add another `else if` clause below `runTestAllGraphsTests()` to invoke a new method called `runHideRestoreTests()`. Start by adding an empty `runHideRestoreTests()` method that just returns `OK`.

- [ ] Begin implementing `int runHideRestoreTests(void)` by adding lines to say what it's starting to do and give an execution duration as part of saying that it's done. See `TestAllGraphs()` in `planarityTestAllGraphs.c` for code to copy.

- [ ] Then, in `runHideRestoreTests()`, use a G6 read iterator to read each graph from the `n8.mAll.g6` file in the samples directory. See `testAllGraphs()` for sample G6 iterator code. For each graph, invoke a new method with signature `int runHideRestoreTest(graphP theGraph)`. Just have `runHideRestoreTest()` return `OK` to start. Please remember to free the graph and the G6 read iterator when done. If `runHideRestoreTest()` returns `NOTOK`, then discontinue the iteration through the G6 file and return NOTOK.

- [ ] Within `runHideRestoreTest()`, first use `gp_WriteToString()` to get a string containing the adjacency list representation of the received graph before manipulation. Then, use a vertex iteration loop to invoke `gp_HideVertex()` for each vertex. Test for a result `!= OK` on each `gp_HideVertex()` and return `NOTOK` if this occurs (remember to free the string). After the loop, invoke `gp_RestoreVertices()` and return `NOTOK` if the function returns a result `!= OK` (and free the string). Last, use `gp_WriteToString()` to get a second string containing the adjacency list representation of the graph after manipulation. Store the result of `strcmp` on the two strings, then free both strings. Return OK if the `strcmp` result was 0, and return `NOTOK` otherwise. 

- [ ] Ensure each `NOTOK` return in `runHideRestoreTest()` has a `gp_ErrorMessage()` to indicate what went wrong.

- [ ] In `_HideVertex()` in `graph.c`, put a size versus capacity check before the `sp_Push()` in the loop or the block of 7 `sp_Push() calls after the loop. Use `sp_GetCurrentSize()` and `sp_GetCapacity()`, and return `NOTOK` if the stack size would become greater than the capacity after the push or block of push calls (the size must be less than the capacity before a push).  The `gp_ErrorMessage()` should say "_HideVertex() is attempting to push to a full stack."

- [ ] In `_RestoreVertex()` in `graph.c`, before each `sp_Pop()` or block of `sp_Pop()` calls, perform a check to ensure that the stack is not empty before the call. For the single `sp_Pop()` call, a stack underflow can be tested and avoided with `sp_IsEmpty()`. For the block of `sp_Pop()` calls, use `sp_GetCurrentSize()`. Return `NOTOK` if the test reveals that a stack underflow would occur if the `sp_Pop()` or block of them were called. The `gp_ErrorMessage()` should say "_RestoreVertex() is attempting to pop from an empty stack."

- [ ] This new test can be run in debug mode by going into `launch.json` and uncommenting the `args` that run `planarity -test <samplesDir>`.

If `planarity -test` works, then you are ready to submit a PR. If you are getting errors that are in `gp_HideVertex()` and `gp_RestoreVertices()`, or the underlying methods they call, then please place a comment in this issue to give your fork and branch location and to request assistance with diagnosing the problem (because the issue complexity will be increased to above easy).

WilliamK112:
## Summary

Fixes #249.

This adds a hide/restore regression pass to `planarity -test`. The new test reads each graph from `n8.mALL.g6`, writes the graph to an adjacency-list string, hides every vertex, restores all vertices, writes the graph again, and verifies the before/after strings match.

It also adds the requested stack safety checks around `_HideVertex()` pushes and `_RestoreVertex()` pops so malformed or overfull stack states produce a clear `gp_ErrorMessage()` instead of silently relying on unchecked stack operations.

## Tests

- `git diff --check`
- `%TEMP%\eaps-planarity-check-249\planarity.exe -test ./c/samples/`

## Safety

The new test runs only through the existing `planarity -test` regression path. The graph library changes are guard checks only; they do not change hide/restore behavior when stack capacity and contents are valid.

### DIFF
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -2443,12 +2443,24 @@ int _HideVertex(graphP theGraph, int vertex)
     // Cycle through all the edges, pushing and hiding each
     while (gp_IsEdge(theGraph, e))
     {
+        if (sp_GetCurrentSize(theGraph->theStack) >= sp_GetCapacity(theGraph->theStack))
+        {
+            gp_ErrorMessage("_HideVertex() is attempting to push to a full stack.");
+            return NOTOK;
+        }
+
         sp_Push(theGraph->theStack, e);
         gp_HideEdge(theGraph, e);
         e = gp_GetNextEdge(theGraph, e);
     }
 
     // Push the additional integers needed by gp_RestoreVertex()
+    if (sp_GetCurrentSize(theGraph->theStack) + 7 > sp_GetCapacity(theGraph->theStack))
+    {
+        gp_ErrorMessage("_HideVertex() is attempting to push to a full stack.");
+        return NOTOK;
+    }
+
     sp_Push(theGraph->theStack, hiddenEdgeStackBottom);
     sp_Push(theGraph->theStack, NIL);
     sp_Push(theGraph->theStack, NIL);
@@ -2737,7 +2749,10 @@ int _RestoreVertex(graphP theGraph)
     int u, v, e_u_succ, e_u_pred, e_v_first, e_v_last, HESB, e;
 
     if (sp_GetCurrentSize(theGraph->theStack) < 7)
+    {
+        gp_ErrorMessage("_RestoreVertex() is attempting to pop from an empty stack.");
         return NOTOK;
+    }
 
     sp_Pop(theGraph->theStack, v);
     sp_Pop(theGraph->theStack, u);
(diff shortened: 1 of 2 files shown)


john-boyer-phd (approved):
This is great, William, thank you!

PR #268, opened 2026-07-07
Add identify/contract regression test

### LINKED ISSUE #250: Add a Regression Test for Vertex Identification and Edge Contraction
The goal of this issue is to add to what happens when running `planarity -test` (which is run very frequently). The new code will directly run the methods `gp_IdentifyVertices()` and `gp_RestoreVertices()`, which will implicitly also run  `gp_ContractEdge()`. The following steps are needed:

- [ ] In `planarityCommandLine.c`, add another `else if` clause below `runTestAllGraphsTests()` to invoke a new method called `runIdentifyContractTests()`. Start by adding an empty `runIdentifyContractTests()` method that just returns `OK`.

- [ ] Begin implementing `int runIdentifyContractTests(void)` by adding lines to say what it's starting to do and give an execution duration as part of saying that it's done. See `TestAllGraphs()` in `planarityTestAllGraphs.c` for code to copy.

- [ ] Then, in `runIdentifyContractTests()`, use a G6 read iterator to read each graph from the `n8.mAll.g6` file in the samples directory. See `testAllGraphs()` for sample G6 iterator code. For each graph, invoke a new method with signature `int runIdentifyContractTest(graphP theGraph)`. Just have `runIdentifyContractTest()` return `OK` to start. Please remember to free the graph and the G6 read iterator when done. If `runIdentifyContractTest()` returns `NOTOK`, then discontinue the iteration through the G6 file and return NOTOK.

- [ ] Within `runIdentifyContractTest()`, first use `gp_WriteToString()` to get a string containing the adjacency list representation of the received graph before manipulation. Then, systematically use `gp_IdentifyVertices()` to reduce the graph to a single vertex as follows: identify vertex 1 with 2, 3 with 4, 5 with 6 and 7 with 8; then identify vertex 1 with 3 and 5 with 7; then identify vertex 1 with 5. Test for a result `!= OK` on each `gp_IdentifyVertices()` and return `NOTOK` if this occurs (remember to free the string). After the loop, invoke `gp_RestoreVertices()` and return `NOTOK` if the function returns a result `!= OK` (and free the string). Last, use `gp_WriteToString()` to get a second string containing the adjacency list representation of the graph after manipulation. Store the result of `strcmp` on the two strings, then free both strings. Return OK if the `strcmp` result was 0, and return `NOTOK` otherwise. 

- [ ] Ensure each `NOTOK` return in `runIdentifyContractTest()` has a `gp_ErrorMessage()` to indicate what went wrong.

- [ ] In `_IdentifyVertices()` and `_ContractEdge()` in `graph.c`, put a size versus capacity check before each `sp_Push()` or block of `sp_Push()` calls using `sp_GetCurrentSize()` and `sp_GetCapacity()`. Return `NOTOK` if the stack size would become greater than the capacity after the push or block of push calls (the size must be less than the capacity before a push).  The `gp_ErrorMessage()` should say "_IdentifyVertices() is attempting to push to a full stack." or "_ContractEdge() is attempting to push to a full stack."

- [ ] This new test can be run in debug mode by going into `launch.json` and uncommenting the `args` that run `planarity -test <samplesDir>`.

If `planarity -test` works, then you are ready to submit a PR. If you are getting errors that are in `gp_IdentifyVertices()`,  `gp_ContractEdge()`, or `gp_RestoreVertices()`, then please place a comment in this issue to give your fork and branch location and to request assistance with diagnosing the problem (because the issue complexity will be increased to above easy).

WilliamK112:
## Summary
- Add an identify/contract regression pass to `planarity -test` over `n8.mALL.g6`
- Verify `gp_IdentifyVertices()` / `gp_RestoreVertices()` preserve the adjacency-list representation after reducing each graph to one vertex
- Add stack-capacity checks before identify/contract stack pushes
- Fix restored vertex adjacency lists so the restored last edge has `next` reset to `NIL`

Fixes #250

## Tests
- `git diff --check`
- `gcc -std=c99 -I c -I c\graphLib -I c\graphLib\lowLevelUtils -I c\graphLib\io -I c\graphLib\extensionSystem -I c\graphLib\planarityRelated -I c\graphLib\homeomorphSearch -I c\planarityApp <all c/graphLib and c/planarityApp .c files> -o %TEMP%\eaps-planarity-check-250\planarity.exe`
- `%TEMP%\eaps-planarity-check-250\planarity.exe -test ./c/samples/`
- `clang -std=c99 -I c -I c\graphLib -I c\graphLib\lowLevelUtils -I c\graphLib\io -I c\graphLib\extensionSystem -I c\graphLib\planarityRelated -I c\graphLib\homeomorphSearch -I c\planarityApp <all c/graphLib and c/planarityApp .c files> -o %TEMP%\eaps-planarity-check-250-clang\planarity.exe`
- `%TEMP%\eaps-planarity-check-250-clang\planarity.exe -test ./c/samples/`

Note: I could not run the repository's full autotools CI flow locally because this Windows/MSYS environment does not have `autoreconf` installed.

### DIFF
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -2505,6 +2505,13 @@ int _ContractEdge(graphP theGraph, int e)
     v = gp_GetNeighbor(theGraph, e);
 
     eBefore = gp_GetNextEdge(theGraph, e);
+
+    if (sp_GetCurrentSize(theGraph->theStack) >= sp_GetCapacity(theGraph->theStack))
+    {
+        gp_ErrorMessage("_ContractEdge() is attempting to push to a full stack.");
+        return NOTOK;
+    }
+
     sp_Push(theGraph->theStack, e);
     gp_HideEdge(theGraph, e);
 
@@ -2571,6 +2578,11 @@ int _IdentifyVertices(graphP theGraph, int u, int v, int eBefore)
     if (gp_IsEdge(theGraph, e))
     {
         int result = gp_ContractEdge(theGraph, e);
+        int hiddenEdgesStackBottomIndex;
+        int hiddenEdgesStackBottomValue;
+
+        if (result != OK)
+            return result;
 
         // The edge contraction operation pushes one hidden edge then
         // recursively calls this method. This method then pushes K
@@ -2583,8 +2595,8 @@ int _IdentifyVertices(graphP theGraph, int u, int v, int eBefore)
         // six more integers to indicate edges that were moved from
         // v to u, so the "hidden edges stackBottom" is in the next
         // position down.
-        int hiddenEdgesStackBottomIndex = sp_GetCurrentSize(theGraph->theStack) - 7;
-        int hiddenEdgesStackBottomValue = sp_Get(theGraph->theStack, hiddenEdgesStackBottomIndex);
+        hiddenEdgesStackBottomIndex = sp_GetCurrentSize(theGraph->theStack) - 7;
+        hiddenEdgesStackBottomValue = sp_Get(theGraph->theStack, hiddenEdgesStackBottomIndex);
 
         sp_Set(theGraph->theStack, hiddenEdgesStackBottomIndex, hiddenEdgesStackBottomValue - 1);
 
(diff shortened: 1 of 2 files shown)


john-boyer-phd (approved):
This PR makes me _very_ happy, William. Some years ago, I had graph coloring code that was getting a very occasional fault on a Fedora Linux VM, but I had to remove the code during prep for inclusion in Debian science packages before I could look into it. I had hoped these tests would reveal the issue, so thank you for looking for, finding, and fixing the issue. Looks great!

PR #271, opened 2026-07-07
fix: promote visitedInfo to visitedIndex in graphIO.c

### LINKED ISSUE #270: Eliminate use of `visitedInfo`/`visitedIndex` from `_ReadAdjList()` in `graphIO.c`
If #269 has been done, then this issue is about use of get/set `visitedIndex` use in `_ReadAdjList()`. Otherwise, it's about get/set `visitedInfo`.

The `visitedInfo` exists in the Planarity Graph logical subclass, and is being moved to a baser class, the DFSUtils Graph. However, graph I/O is logically associated with the base Graph class. So, it is a matter of continued tidying up within the package-private code space to stop using `visitedInfo`/`visitedIndex` in `_ReadAdjList()`. Fortunately, though not trivial, it is very easy to switch to using the vertex `index` in the base Graph `VertexRec` instead, as follows:

 - [ ]  In the vertex iteration loop just after `gp_EnsureVertexCapacity()`, change from calling `gp_SetVertexVisitedInfo()` to `gp_SetIndex()` so the vertex index values are guaranteed to be `NIL` before the main reading loop.
 - [ ] Inside the main reading `for` loop, remove the `gp_SetIndex()` call, and change the comparison that calls `gp_GetIndex()` to just compare `v` with `indexValue` instead.
 - [ ] In the main reader `for` loop, change calls to get/set `visitedInfo`/`visitedIndex` to instead call `gp_GetIndex()` and `gp_SetIndex()`.
 - [ ] Also in the `while` loop after the main reader `for` loop, change the call to set `visitedInfo`/`visitedIndex` to instead call `gp_SetIndex()`.
 - [ ] At the end of `_ReadAdjList()`, just before `return OK`, add a vertex iteration `for` loop like the one you amended in the first step above to call `gp_SetIndex(theGraph, v, v)` (because the exit condition of this method is that the `index` fields equal the vertex array index location until a depth-first search is called).
 - [ ] Check that `planarity -test` succeeds (uncomment the command line args in `launch.json`).

Yashas77:
Resolves #270

## Type of change
- Bug fix 

### Updated `graphIO.c`
Knocked out the refactoring requested in Issue #270:
- Swapped `gp_SetVertexVisitedInfo` to `gp_SetIndex` before the main reading loop.
- Pulled out `gp_SetIndex` inside the main loop and updated the `indexValue` comparison.
- Swapped all get/set `visitedInfo` calls to their `Index` equivalents within the `while` loops.
- Added the final vertex iteration loop right before the return statement.

## Testing
- Walked through the logic line-by-line to match the exact checklist provided in the issue.

### DIFF
--- a/c/graphLib/io/graphIO.c
+++ b/c/graphLib/io/graphIO.c
@@ -161,7 +161,7 @@ int _ReadAdjList(graphP theGraph, strOrFileP inputContainer)
     // Clear the visited members of the vertices so they can be used
     // during the adjacency list read operation
     for (v = gp_LowerBoundVertices(theGraph); v < gp_UpperBoundVertices(theGraph); ++v)
-        gp_SetVertexVisitedInfo(theGraph, v, NIL);
+        gp_SetIndex(theGraph, v, NIL);
 
     // Do the adjacency list read operation for each vertex in order
     for (v = gp_LowerBoundVertices(theGraph); v < gp_UpperBoundVertices(theGraph); ++v)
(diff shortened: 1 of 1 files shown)


john-boyer-phd (approved):
This looks great, thank you Yashas! (That failed GitHub Action was to do with Microsoft and not your PR code, and it righted itself when Wanda ran it again).

Issue #272, opened 2026-07-07
Remediate digraph adjacency list reading with 0-based array compile mode


john-boyer-phd:
Although the releases of the Edge Addition Planarity Suite (EAPS) are based on the `USE_1BASEDARRAYS` compile mode, we still have `USE_0BASEDARRAYS` as a legacy compile mode because it can, in theory at least, help find problems because it is easier to surface memory overruns if they are actually happening.

The `planarity -test` regression test for digraphs is not working with `Petersen.digraph.txt` in `USE_0BASEDARRAYS` compile mode. To resolve this issue, the following steps are needed:

 - [ ] Uncomment `USE_0BASEDARRAYS` in `appconst.h`.
 - [ ] Modify your local copy of `launch.json` (not the one in devEnvSetup) to uncomment the command-line args for running `-test`.
 - [ ] Temporarily comment out other regression tests so you can focus on the `Petersen.digraph.txt` test.
 - [ ] Compile and run in debug mode.
 - [ ] Step through `_ReadAdjList()` to figure out what it is doing differently than when run in `USE_1BASEDARRAYS`
 - [ ] Determine and propose a fix using a PR

Issue #245, opened 2026-06-16
Finish articulating extension for Planarity Graph Subclass


john-boyer-phd:
The Planarity Graph subclass still relies on non-public code that is run by the base Graph class. The goal of this issue is to create an explicit graphExtension, as is done for planar graph drawing, and then migrate the planarity-specific private data structures and logic to the extension. 

This issue relies on (is blocked by) completion of #239 because many macros that currently refer to the members  of the graphPrivateDataStruct must instead refer into the Planarity graphExtensionP and have a constant-time way to find the extension.

- [ ] In the following files, replace `graph` with `graphPlanarity_` so it is clearer that these files are part of the Planarity extension implementation: `graphEmbed.c`, `graphIsolator.c`, and `graphNonplanar.c`. This task includes making the proper changes in `makefile.am`.

- [ ] Refactor `graphTests.c` to extract into a separate file named `graphPlanarity_Tests.c` those methods that are planarity-specific, such as those for checking embedding integrity. Generic methods like `_TestSubgraph()` should stay in `graphTests.c` but methods that check facial integrity or check the structure of a _K<sub>3,3</sub>_  or _K<sub>5</sub>_ should go to `graphPlanarity_Tests.c`. This includes adding `graphPlanarity_Tests.c` to `makefile.am`.

- [ ] In `graphPlanarity.private.h`, add a declaration for the `PLANARITY_ID` and `PlanarityContext` structure and pointer. Transfer from the graphPrivateDataStruct the following: `PVI`, `sortedDFSChildLists`, `extFace`, and `IC`.

- [ ] In `graphPlanarity_Extensions.c`, add the `PLANARITY_ID` declaration and amend the implementation of `gp_ExtendWith_Planarity()` to be analogous with `gp_ExtendWith_DrawPlanar()`, such as invoking `gp_AddExtension()` and clearing, creating and initializing structures. 

- [ ] Implementations for the methods for clearing, creating, and initializing structures must be created by transferring the code out of the base Graph implementations that currently perform those operations.

[...]

PR #277, opened 2026-07-13
Implement extension lookup table

### LINKED ISSUE #239: Implement the `extensionLookupTable`
There is an initial implementation of this capability in `gp_CopyExtension()` in `graphExtensions.c`, though it is named `srcExtensionIDMap`. The goal of this issue is to promote/replace the `srcExtensionIDMap` to/with a generalized `extensionLookupTable` feature. The following details how to do this:

1. Move the declaration of `MAXNUMSUPPORTEDEXTENSIONS` from `graphExtensions.c` to `graph.private.h`, just after the `#endif` for the `extern "C"`.
2. In `graph.c`, add to `gp_New()` an error-checked call to `calloc()` to allocate a `MAXNUMSUPPORTEDEXTENSIONS+1`  pointers of type `graphExtensionP`. Amend the `if` condition, the `if` code, and `else` code appropriately. Put the code right after the code that handles `theGraphPrivateData` in each case.
3. Deallocate the `extensionLookupTable` in `gp_Free()`. Put the code just after the code that frees the `privateData` structure.

Once the changes above are made to allocate and free the extension lookup table, it is necessary to amend code in `graphExtensions.c` to use the extension lookup table, as follows:

1. Amend `gp_AddExtension()` to set `extensionLookupTable[moduleID]` equal to the `newExtension`. Make the assignment just before the return.
2. In `gp_FindExtension()`, replace the linked list search with `*pContext = theGraph->extensionLookupTable[moduleID]`
3. In `gp_RemoveExtension()`, just before `FreeExtension(curr)`, assign `NULL` to `theGraph->extensionLookupTable[curr->moduleID]`
4. In `gp_DupExtension()`, just after the block of code that assigns values to the data members of the `newNext` extension, assign the `newNext` pointer to `dstGraph->extensionLookupTable[newNext->moduleID]`. It'd be great if you could add a comment before to say "// Add the new extension to the dstGraph extension lookup table" and then add a comment just afterward to say "// Add the new extension to the dsstGraph extensions linked list"
5.  In `gp_CopyExtension()`, remove the declaration, initialization, and usage of `srcExtensionIDMap` and instead just use the source graph's `extensionLookupTable` to get the `srcExtension`.
6. In `gp_FreeExtensions()`, assign `NULL` to all the pointers in the `extensionLookupTable` using `memset()` to 0; remember that the array length is `MAXNUMSUPPORTEDEXTENSIONS+1`.

Once these are done, you can test whether this works by compiling/running in Debug mode, and then choose menu option 3 for K3,3 search, then do 100,000 graphs with 20 vertices. If it doesn't work, double-check all your code, and if you're sure that all the instructions have been followed, please place a comment on this issue that includes a link to your branch on your fork.

If all is working well, then submit a PR and ask for a review.

WilliamK112:
﻿## Summary
- allocate and free `extensionLookupTable` with each graph instance
- maintain the lookup table when extensions are added, removed, duplicated, copied, and freed
- replace the temporary `srcExtensionIDMap` copy helper with the source graph lookup table

Closes #239.

## Testing
- `git diff --check`
- `gcc -O2 -Wall -Wextra -Werror -Wno-unused-parameter -o ./planarity.exe <all planarityApp + graphLib C sources>`
- `./planarity.exe -r -3 100000 20`
- `gcc -O2 -Wall -Wextra -Werror -Wno-unused-parameter -DUSE_0BASEDARRAYS -o ./planarity-0based.exe <all planarityApp + graphLib C sources>`
- `./planarity-0based.exe -r -3 100000 20`
- default and `USE_0BASEDARRAYS` stdin-fed graphLib smoke harnesses for `gp_CopyGraph()` and `gp_DupGraph()` with K3,3/K4 extensions attached in different source/destination orders

Note: a full `USE_0BASEDARRAYS` `./planarity-0based.exe -test ./c/samples/` run still reaches the known Petersen digraph failure covered by #272/#275; the focused extension stress checks above pass.

### DIFF
--- a/c/graphLib/extensionSystem/graphExtensions.c
+++ b/c/graphLib/extensionSystem/graphExtensions.c
@@ -8,6 +8,7 @@ See the LICENSE.TXT file for licensing information.
 #include <string.h>
 
 #include "../lowLevelUtils/appconst.h"
+#include "../graph.private.h"
 
 #include "graphExtensions.private.h"
 #include "graphExtensions.h"
@@ -182,7 +183,12 @@ int gp_AddExtension(graphP theGraph,
 
     if (theGraph == NULL || pModuleID == NULL ||
         context == NULL || copyData == NULL || freeContext == NULL ||
-        functions == NULL)
+        functions == NULL || theGraph->extensionLookupTable == NULL)
+    {
+        return NOTOK;
+    }
+
+    if (*pModuleID < 0 || *pModuleID > MAXNUMSUPPORTEDEXTENSIONS)
     {
         return NOTOK;
     }
@@ -196,6 +202,10 @@ int gp_AddExtension(graphP theGraph,
     // Assign a unique ID to the extension if it does not already have one
     if (*pModuleID == 0)
     {
+        if (moduleIDGenerator >= MAXNUMSUPPORTEDEXTENSIONS)
+        {
+            return NOTOK;
+        }
         *pModuleID = ++moduleIDGenerator;
     }
 
@@ -218,6 +228,7 @@ int gp_AddExtension(graphP theGraph,
     // Make the new linkages
     newExtension->next = (struct graphExtensionStruct *)theGraph->extensions;
     theGraph->extensions = newExtension;
+    theGraph->extensionLookupTable[newExtension->moduleID] = newExtension;
 
     // The new extension was successfully added
     return OK;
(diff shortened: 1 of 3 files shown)


john-boyer-phd (approved):
All good, William. This helps extra because it unblocks #245, which I've been looking forward to for a long time. (It was technically feasible to do #245 without this, but not really because we don't want planarity to ever get slower, and having the lookup table means that accessing planarity-specific data structures in an extension should not really be any slower than the current method based on the graph's "private data structure" data members).

PR #280, opened 2026-07-13
Add planar embedding face APIs

### LINKED ISSUE #244: Add methods for counting and creating a list of faces of a planar embedding
It would be useful to have public methods that report the number of faces in a planar embedding and that create a list of the faces in a planar embedding.

Place the method signatures in `graphPlanarity.h` and place the method implementations in a new file called `graphPlanarity_Faces.c`. Copy/paste the project copyright message from any other source file, and ensure that `graphPlanarity_Faces.c` is added to the `makefile.am` in the `libplanarity_la_SOURCES` just below `graphPlanarity_Extensions.c`. 

Informative code appears in `_CheckEmbeddingFacialIntegrity()` in `graphTests.c`.

First, add the signature and implementation of `int gp_CountEmbeddingFaces(graphP theGraph)`. The code can be directly pulled from the beginning of `_CheckEmbeddingFacialIntegrity()`, up to the end of the `while` loop. The method should return -1 on an error, especially the three locations that currently `return NOTOK`. Also, return -1 (an error) if the face count calculated by the `while` loop does not exactly equal `gp_GetM(theGraph) - gp_GetN(theGraph) + 2`. Otherwise, return the face count. Note that the _expected_ face count is clearly a constant-time calculation, but this method indicates how many faces that theGraph does have. Future versions of this method may return a result other than `gp_GetM(theGraph) - gp_GetN(theGraph) + 2` if, for example, the graph is embedded on a surface having greater topological complexity than a plane. 

Second, refactor `_CheckEmbeddingFacialIntegrity()` to call `gp_CountEmbeddingFaces()`, rather than continuing to use the loops just extracted from it. 

The new method `gp_CountEmbeddingFaces()` can then be easily tested by by just using menu mode to run planar embedding on 100,000 randomly generated 20-vertex graphs.

Third, add the signature and implementation of `int gp_CreateEmbeddingFaceList(graphP theGraph, char **pFaceList)`. The method should return `OK` on success, or `NOTOK` on failure, such as a `NULL` input graph or `pFaceList` pointer, a non-`NULL` `*pFaceList` pointer, if `GRAPHFLAGS_DFSNUMBERED` is not set, or if an error result (-1) is obtained from `gp_CountEmbeddingFaces()`.

On success, this method generates a caller-owned character string that is ouput to `*pFaceList`. The caller is responsible to call free on the string when it isn't needed anymore. The low-level utility object `strBuf` should be used to facilitate easily building the desired string. It would be useful to add a helper method to `strBuf` called `int sb_ConcatInt(strBufP, int)` that does the following:
- Declares `char numberStr[MAXCHARSFOR32BITINT + 1];`
- Uses `sprintf()` to get a string version of the integer parameter;
- Invokes `sb_ConcatString()`

Using a `strBuf`, build up the output string in the format of a YAML file. YAML has been selected because the format is sufficiently flexible to represent one face list per connected component and because there are YAML readers available in multiple languages including Python, C, and Java. The output string format should look like this:

```
Component_1: |
    1, 2, 3, 4, 5, 1
    6, 7, 8, 6
    9, 10, 11, 12, 9
...
Component_n: |
    21, 22, 23, 24, 25, 21
    26, 27, 28, 26
    29, 30, 31, 32, 29
```

The number of lines of numbers in a component corresponds to the number of faces in that component. The numbers in each line's comma-separated list are the indexes of the vertices visited while traversing the face.

The code for walking the faces can be similar to the `for` and `while` loops in `gp_CountEmbeddingFaces()`. However, it is necessary to output component-level separators, so the `while` loop can only process the edges from one component at a time. The `for` and `while` loops in `gp_DepthFirstSearch()` can be used to detect each connected component (the `for` loop) and then push the edges in the connected component (like the `while` loop in `gp_DepthFirstSearch()`). Once the edges of a connected component have been pushed, then a second `while` loop like the `while` loop in `gp_CountEmbeddingFaces()` can be used to walk the faces of the connected component.  

To test, create an adjacency list representation of a custom graph consisting of the following:
- Create hexagon on 6 vertices and a triangle on 3 more vertices.
- Place another "separator" vertex between the hexagon and triangle, and connect the vertex with an edge to the hexagon and an edge to the triangle.
- Duplicate the above construct two times so that there are three separate connected components.

Perform `gp_Embed()` then `gp_CreateEmbeddingFaceList()`. Observe whether you get the three components each having three faces, one with 6 edges (7 numbers), one with 3 edges (4 numbers), and one that appears to have 13 edges, but with some duplication of edges containing the "separator" vertex.  

Add your sample input file and expected output file to the samples directory and its `makefile.am`, and add this test to the list of tests performed by `planarity -test`, i.e., add this test with the method name `runFaceListTest()` and invoke it in a new `else` clause within `runQuickRegressionTests()`, after the `else` clause that invokes `runTestAllGraphsTests()`. Perform `planarity -test _samplesDir_` to validate the implementation.

WilliamK112:
## Summary
- add `gp_CountEmbeddingFaces()` and `gp_CreateEmbeddingFaceList()` for planar embeddings
- reuse the new face counter from the existing embedding facial integrity check
- add 1-based and 0-based regression samples for a disconnected planar embedding face list

## Notes
- The face count validates the extended Euler formula for disconnected graphs, preserving the existing facial integrity behavior.

Fixes #244

## Testing
- `git diff --check`
- `gcc -O2 ... -o .\planarity.exe`
- `.\planarity.exe -test .\c\samples\`
- `gcc -O2 -DUSE_0BASEDARRAYS ... -o .\planarity-0based.exe`
- `.\planarity-0based.exe -test .\c\samples\` reaches and passes `Face List Test`, then hits the known `Petersen.digraph.txt` 0-based sample reader failure tracked by #272/#275
- `.\planarity.exe -r -p 100000 20`
- `.\planarity.exe -r -3 100000 20`
- `.\planarity.exe -r -4 100000 20`
- `.\planarity-0based.exe -r -p 100000 20`
- `.\planarity-0based.exe -r -3 100000 20`
- `.\planarity-0based.exe -r -4 100000 20`


john-boyer-phd (commented):
This is wonderful work, William, really!

I'd like to ask if you would be able to make some small changes.

First, it looks like some of the other changes from the other PRs have created merge conflicts now that they have been merged into master. Please can you update your own fork's master, then apply that to your branch for this PR. This should reproduce the merge conflicts for you and enable you to resolve them, and then do a commit to your branch with the merge conflict fixes. That should apply automatically to the PR, clearing up the merge conflict status of the PR.

Second, in `runFaceListTest()`, I'd like to ask you to call `gp_SortVertices()` after `gp_TestEmbedResultIntegrity()` (please see Line 188 of `planaritySpecificGraph.c` for an example). The reason for this change is that the face list outputs will then have the same vertex numbering as the input graphs. The code change will cause the need for you to update the output sample files. For example, in `faceListComponents.out.txt`, the last line of `Component_1` should become `7, 8, 9, 7`.

Third, I originally didn't ask for this, but seeing your work now, there is clearly a nice opportunity to also include textual drawings of the two faceListComponents files so that someone can look at the drawings and your face list outputs at the same time and easily see the correspondence. You'd have to generate and add the two drawings to the samples, such as `faceListComponents.Drawing.txt` and `faceListComponents.0-based.Drawing.txt`. Then, you'd have to change to extending with DrawPlanar, then test that drawings generated by `gp_DrawPlanar_RenderToString()` match the samples.

I'm very much looking forward to getting this great work into the master branch, William, thank you!

WilliamK112:
Thanks, John. I merged the latest master and resolved the conflicts. runFaceListTest() now extends with DrawPlanar, calls gp_SortVertices() after the integrity check, verifies the updated face-list samples, and compares gp_DrawPlanar_RenderToString() against new 1-based and 0-based drawing samples. Both the default and USE_0BASEDARRAYS builds and complete test suites pass locally. The update is pushed in commit 304573a.

john-boyer-phd (approved):
Excellent work, William. Please have a look at the commits I added to this PR for future work.

PR #281, opened 2026-07-14
Support incremental embedding initialization

### LINKED ISSUE #240: Refactor `_EmbeddingInitialize()` to respond to already-computed DFS and Lowpoint
In preparation for new graph reading methods that may provide either depth-first search (DFS) or both DFS and leastAncestor/lowpoint calculations, it is necessary to refactor how the default `_EmbeddingInitialize()` method works so that it accepts the information of steps already done and only does the steps that have not been done.

To start, we need to add knowledge of whether lowpoints have been computed:

1. In `graphDFSUtils.h`, add a new graph flags declaration: `GRAPHFLAGS_LOWPOINTSCOMPUTED` equal to 2048 just below the declaration for `GRAPHFLAGS_SORTEDBYDFI`.
2. In `graphDFSUtils.c`, at the end of `gp_ComputeLowpoints()`, add a line that sets the `graphFlags` bit flag for `GRAPHFLAGS_LOWPOINTSCOMPUTED`, just as you see for `GRAPHFLAGS_DFSNUMBERED` at the end of `gp_DepthFirstSearch()`.
3. In `graphEmbed.c`, at the end of `_EmbeddingInitialize()`, just before the return and `gp_LogLine()` above it, add the same assignment to set the  `graphFlags` bit flag for `GRAPHFLAGS_LOWPOINTSCOMPUTED`.

Next, we need to refactor `_EmbeddingInitialize()` into two subfunctions:

1. The current implementation should go into a function called `_EmbeddingInitialize_Full()` that should be invoked by `_EmbeddingInitialize()` if none of `GRAPHFLAGS_DFSNUMBERED`, `GRAPHFLAGS_SORTEDBYDFI`, nor `GRAPHFLAGS_LOWPOINTSCOMPUTED` are set. 
2. Call a new function `_EmbeddingInitialize_Incremental()` if any of `GRAPHFLAGS_DFSNUMBERED`, `GRAPHFLAGS_SORTEDBYDFI`, nor `GRAPHFLAGS_LOWPOINTSCOMPUTED` are set.

The new subfunction `_EmbeddingInitialize_Incremental()` can immediately call `gp_ErrorMessage()` and return `NOTOK` if the bit flag settings don't make sense. The following are the sensible settings:

1. `GRAPHFLAGS_DFSNUMBERED` is set, but `GRAPHFLAGS_SORTEDBYDFI` and `GRAPHFLAGS_LOWPOINTSCOMPUTED` are not.
2. `GRAPHFLAGS_DFSNUMBERED` and `GRAPHFLAGS_SORTEDBYDFI` are set but  `GRAPHFLAGS_LOWPOINTSCOMPUTED` is not,
3. None of or all three of `GRAPHFLAGS_DFSNUMBERED`, `GRAPHFLAGS_SORTEDBYDFI`, and `GRAPHFLAGS_LOWPOINTSCOMPUTED` are set.

For example, it doesn't make sense for either or both of `GRAPHFLAGS_SORTEDBYDFI` and `GRAPHFLAGS_LOWPOINTSCOMPUTED` to be set if `GRAPHFLAGS_DFSNUMBERED` is not. Similarly, it is an error if `GRAPHFLAGS_LOWPOINTSCOMPUTED` is set if either or both of `GRAPHFLAGS_DFSNUMBERED` and `GRAPHFLAGS_SORTEDBYDFI` are not set.

It would also be useful, for internal testing purposes, to be able to comment out the invocation of `_EmbeddingInitialize_Full()` and always call `_EmbeddingInitialize_Incremental()`, even if we don't leave it that way for the release. Therefore, the following implementation for `_EmbeddingInitialize_Incremental()` is needed:

1. If the `GRAPHFLAGS_DFSNUMBERED` flag is not set, then invoke `gp_DepthFirstSearch()`.
2. If the `GRAPHFLAGS_SORTEDBYDFI` flag is not set, then invoke `gp_SortVertices()`.
3. If the `GRAPHFLAGS_LOWPOINTSCOMPUTED` flag is not set, then invoke `gp_ComputeLowpoints()`.
4. Perform the `_ClearVertexVisitedFlags()`
5. Perform the first `for` loop from `_EmbeddingInitialize_Full()`, except take a copy of it into `_EmbeddingInitialize_Incremental()` and refactor the copy to remove the code that invokes `gp_SetIndex()`, `gp_SetVertexParent()`, `gp_SetEdgeType()` and `gp_SetVertexLeastAncestor()`. We need the loop to do only items 3, 8 and 4 from the `_EmbeddingInitialize_Full()` list.
6. Perform the last `for` loop in `_EmbeddingInitialize_Full()`, except take a copy of it into `_EmbeddingInitialize_Incremental()` and refactor the copy to remove code that invokes `gp_SetVertexLowpoint()` and its immediately preceding `while` loop and `leastValue` assignment. We need this loop to do the rest of steps 7 and 8 except that the `lowpoint` values have already been computed.

Finally, we need a way to test the new behavior, which will be done at the planarity application application level by amending the `RandomGraphs()` method in `planarityRandomGraphs.c`. Specifically, just before the invocation of `gp_Embed()`, add code that calls `gp_DepthFirstSearch()`, `gp_SortVertices()`, and `gp_ComputeLowpoints()`. Include error testing and `gp_ErrorMessage()` reporting in case those methods return `NOTOK`. Then, you will be able to run the following tests:

1. Use the debugger to execute the code as-is. In the menu mode, press 'P' for planarity and run 100,000 graphs at 20 vertices per graph to see if the implementation works when all three of `GRAPHFLAGS_DFSNUMBERED`, `GRAPHFLAGS_SORTEDBYDFI`, and `GRAPHFLAGS_LOWPOINTSCOMPUTED` are set.
2. Comment out the block of code that calls `gp_ComputeLowpoints()` and run again in the debugger, selecting 'P' for planarity and running 100,000 graphs with 20 vertices per graph because this will test the case of having `GRAPHFLAGS_DFSNUMBERED` and `GRAPHFLAGS_SORTEDBYDFI` set and `GRAPHFLAGS_LOWPOINTSCOMPUTED` not set.
3. Now also comment out the block that invokes `gp_SortVertices()` and run the same random graphs test as in 1 and 2, which will test the incremental function when only `GRAPHFLAGS_DFSNUMBERED` is set.
4. Now comment out the `gp_DepthFirstSearch()` code block and also comment out the invocation of `_EmbeddingInitialize_Full()` so that `_EmbeddingInitialize()` only invokes `_EmbeddingInitialize_Incremental()`. This will help us to know if `_EmbeddingInitialize_Incremental()` can actually do the full job.

If the four tests above all pass with no errors, then uncomment everything that was commented out during the four tests we want `_EmbeddingInitialize()` to call `_EmbeddingInitialize_Full()` when all bits are cleared, and we want `RandomGraphs()` to invoke `gp_DepthFirstSearch()`, `gp_SortVertices()`, and `gp_ComputeLowpoints()`. Then, you are ready to submit a PR.

WilliamK112:
## Summary
- add `GRAPHFLAGS_LOWPOINTSCOMPUTED` and maintain it across lowpoint computation, vertex sorting, graph reset, and adjacency-list copy
- split embedding initialization into full and incremental paths so callers can precompute DFS, sort, and lowpoints before `gp_Embed()`
- precompute DFS, sort, and lowpoints in random graph runs before embedding to exercise the incremental initialization path

Fixes #240

## Testing
- `git diff --check`
- `./planarity.exe -r -p 100000 20`
- `./planarity.exe -r -o 100000 20`
- `./planarity.exe -r -2 100000 20`
- `./planarity.exe -r -3 100000 20`
- `./planarity.exe -r -4 100000 20`
- rebuilt `planarity-0based.exe` with `gcc -O2 -DUSE_0BASEDARRAYS`
- `./planarity-0based.exe -r -p 100000 20`
- `./planarity-0based.exe -r -3 100000 20`
- `./planarity-0based.exe -r -4 100000 20`
- temporary harness for invalid embedding flag combinations
- temporary harness for embedding initialization modes: none, DFS-only, DFS+sort, DFS+sort+lowpoints over 10,000 random graphs, default and `USE_0BASEDARRAYS`

### DIFF
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -344,6 +344,7 @@ void _ResetGraphStorage(graphP theGraph)
 
     theGraph->graphFlags &= ~GRAPHFLAGS_DFSNUMBERED;
     theGraph->graphFlags &= ~GRAPHFLAGS_SORTEDBYDFI;
+    theGraph->graphFlags &= ~GRAPHFLAGS_LOWPOINTSCOMPUTED;
     theGraph->graphFlags &= ~GRAPHFLAGS_DIRECTEDEDGEDETECTED;
     _InitVertices(theGraph);
     _InitEdges(theGraph);
@@ -970,6 +971,7 @@ int gp_CopyAdjacencyLists(graphP dstGraph, graphP srcGraph)
 
     dstGraph->graphFlags &= ~GRAPHFLAGS_DFSNUMBERED;
     dstGraph->graphFlags &= ~GRAPHFLAGS_SORTEDBYDFI;
+    dstGraph->graphFlags &= ~GRAPHFLAGS_LOWPOINTSCOMPUTED;
     dstGraph->graphFlags &= ~GRAPHFLAGS_DIRECTEDEDGEDETECTED;
     if (gp_GetGraphFlags(srcGraph) & GRAPHFLAGS_DIRECTEDEDGEDETECTED)
         dstGraph->graphFlags |= GRAPHFLAGS_DIRECTEDEDGEDETECTED;
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -53,7 +53,8 @@ void _ClearEdgeVisitedFlags(graphP theGraph);
 int _ClearAllVisitedFlagsInBicomp(graphP theGraph, int BicompRoot);
 int _ClearAllVisitedFlagsInOtherBicomps(graphP theGraph, int BicompRoot);
 void _ClearEdgeVisitedFlagsInUnembeddedEdges(graphP theGraph);
-int _FillVertexVisitedIndexInBicomp(graphP theGraph, int BicompRoot, int FillValue);
+int _FillVertexVisitedIndexes(graphP theGraph, int FillValue);
+int _FillVertexVisitedIndexesInBicomp(graphP theGraph, int BicompRoot, int FillValue);
 int _ClearObstructionMarksInBicomp(graphP theGraph, int BicompRoot);
 
 int _gp_FindEdge(graphP theGraph, int u, int v);
@@ -775,7 +776,27 @@ int _SetAllVisitedFlagsOnPath(graphP theGraph, int u, int v, int w, int x)
 }
 
 /********************************************************************
- _FillVertexVisitedIndexInBicomp()
+ _FillVertexVisitedIndexes()
+
+ Places the FillValue into the visitedIndex of all non-virtual vertices
+ in the graph.
+
+ Returns OK on success, NOTOK on failure.
+ ********************************************************************/
+
+int _FillVertexVisitedIndexes(graphP theGraph, int FillValue)
+{
+    if (theGraph == NULL)
+        return NOTOK;
+
+    for (int v = gp_LowerBoundVertices(theGraph); v < gp_UpperBoundVertices(theGraph); ++v)
+        gp_SetVertexVisitedIndex(theGraph, v, FillValue);
+
+    return OK;
+}
+
+/********************************************************************
+ _FillVertexVisitedIndexesInBicomp()
 
  Places the FillValue into the visitedIndex of the non-virtual vertices
  in the bicomp rooted by BicompRoot.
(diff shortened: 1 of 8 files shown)


john-boyer-phd (approved):
Thank you for helping me pull back apart that initialization into something that is much easier for people to follow, compared to the full version in which steps were pushed together to optimize relative to the original reference implementation. Combining this with your PR that adds regression testing that invokes RandomGraphs(), we also now will have regression testing for the three big methods of DFSUtils.

PR #282, opened 2026-07-14
Name planarity app choice strings

### LINKED ISSUE #190: Update comparisons to string literals in code to call `#define` macro that evaluates to string literal
During the course of #186 , I changed how we were handling user input (i.e. calling `GetLineFromStdin()` and then `sscanf()` for the expected input). In particular, in `c/planarityApp/planarityUtils.c`'s `Reconfigure()` and `c/planarityApp/planarityRandomGraphs.c`'s `RandomGraph()`, I use `YESNOCHOICECHARS` (declared in `c/planarityApp/planarity.h`) for user input validation of configuration choices.

There are other opportunities to create macros which evaluate to string literals to ensure we're not manually duplicating the same string literal throughout the project:
* In `c/planarityApp/planarityUtils.c`'s `Reconfigure()`, when the user has indicated that they wish to generate random graphs, they are asked whether they wish to save only the last 10 graphs as adjacency list or to save all graphs in a `.g6` output file, comparing the output to `aAgG`; instead, we could `#define RANDOMGRAPHSOUTPUTCHOICECHARS "aAgG"`
    * __**N.B.**__ `c/planarityApp/planarityMenu.c`'s `TransformGraphMenu()` is the only place that uses `GetSupportedOutputFormats()`, which returns the string literal `"gam"`... Should we instead introduce `#define TRANSFORMGRAPHOUTPUTCHOICES "gGaAmM"`, and update `Reconfigure()` and `RandomGraphs()` to also allow the user to specify that they wish to save the last 10 graphs as adjacency matrix? This would require us to update the if statement around [line 180](https://github.com/graph-algorithms/edge-addition-planarity-suite/blob/master/c/planarityApp/planarityRandomGraphs.c#L180) to `WRITE_ADJLIST` *or* `WRITE_ADJMAT` depending on the choice in the `Reconfigure()` menu)? Then, both `TransformGraphMenu()` and `Reconfigure()`/`RandomGraphs()` would use the same `TRANSFORMGRAPHOUTPUTCHOICES` macro!
* `planarityUtils.c` has `GetAlgorithmChoices()` return string literal `pdo234`; instead, we could `#define GRAPHALGORITHMCHOICES "pdo234"`

WilliamK112:
## Summary
- add named macros for planarity app mode, random graph output, algorithm, and transform output choices
- use those macros in `Reconfigure()` validation and existing helper return values
- preserve existing accepted inputs and avoid behavior changes

Fixes #190

## Testing
- `git diff --check`
- temporary harness verifying the changed choice helpers return the new macros

### DIFF
--- a/c/planarityApp/planarity.h
+++ b/c/planarityApp/planarity.h
@@ -21,7 +21,11 @@ extern "C"
 #define SUFFIXMAXLENGTH 32
 #define COMMANDSTRINGMAXLENGTH 2
 
+#define MODECHOICECHARS "rsmn"
 #define YESNOCHOICECHARS "yYnN"
+#define RANDOMGRAPHSOUTPUTCHOICECHARS "aAgG"
+#define GRAPHALGORITHMCHOICES "pdo234"
+#define TRANSFORMGRAPHOUTPUTFORMATCHOICES "gam"
 
     char const *GetProjectTitle(void);
     char const *GetAlgorithmFlags(void);
--- a/c/planarityApp/planarityUtils.c
+++ b/c/planarityApp/planarityUtils.c
@@ -45,7 +45,7 @@ int Reconfigure(void)
 
         if (strlen(lineBuff) != 1 ||
             sscanf(lineBuff, " %c", &Mode) != 1 ||
-            !strchr("rsmn", tolower(Mode)))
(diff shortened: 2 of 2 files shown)


john-boyer-phd (approved):
All good.

Issue #251, opened 2026-06-19
Overload for `gp_DeleteEdge()`


john-boyer-phd:
Currently, the method `gp_DeleteEdge()` is not overloadable by Graph subclasses, so extensions that manage extra edge data must implement and exclusively use their own `_Feature_DeleteEdge()` methods that clear an edge's data and then call `gp_DeleteEdge()`. This works, but a better implementation is to have an overloadable `gp_DeleteEdge()`, as follows:

- [x] Add `fpDeleteEdge` to the function table, just above `fpHideEdge`. 
- [x] Rename the current `gp_DeleteEdge( )` in `graph.c` to be  `_DeleteEdge()`, and then assign `_DeleteEdge()` to the `fpDeleteEdge` pointer in `InitFunctionTable()`, just above the assignment to `fpHideEdge`.
- [x] Write a new `gp_DeleteEdge()`, move the parameter checks from `_DeleteEdge()` to the new `gp_DeleteEdge()` and then make `gp_DeleteEdge()` call the `fpDeleteEdge` function. See `gp_ContractEdge()` and `_ContractEdge()` for reference on how to do this.
- [x] At this point, you should be able to compile and then run planarity on 100,000 20-vertex graphs successfully.

- [x] Refactor the code in the *K<sub>3,3</sub>* search to assign its `_K33Search_DeleteEdge()` as an overload of `fpDeleteEdge` in `gp_ExtendWith_K33Search()`, to have `_K33Search_DeleteEdge()` properly call the base class `fpDeleteEdge` using the function pointer in the `K33Search` extension context. See the `else` clause in `_K33Search_EmbedPostprocess()` for how to do this.
- [x] Replace all direct invocations of `_K33Search_DeleteEdge()` in `graphK33Search.c` with invocations of `gp_DeleteEdge()`.
- [x] Move `_K33Search_DeleteEdge()` from `graphK33Search.c` to `graphK33Search_Extensions.c`, just below `_K33Search_EmbedPostprocess` (and add the forward declaration at the top, just below the one for `_K33Search_EmbedPostprocess`)
- [x] Compile and run the planarity application, then use command 3 to test  *K<sub>3,3</sub>* search on 100,000 20-vertex graphs.

- [x] Similarly refactor the code in the *K<sub>4</sub>* search related to `_K4_DeleteEdge()` and then test *K<sub>4</sub>* search on 100,000 20-vertex graphs.

[...]

PR #283, opened 2026-07-14
Name low-level utility header parameters

WilliamK112:
## Summary
- add parameter names to `strbuf.h` declarations to match the implementation definitions
- add parameter names to `stack.h` declarations to match the implementation definitions
- add parameter names to private logging string helper declarations

Refs #188

This intentionally avoids `graphFunctionTable.h` because it overlaps an open PR/claim for #251 and its function-pointer fields are a separate conflict surface.

## Testing
- `git diff --check`

### DIFF
--- a/c/graphLib/io/strbuf.h
+++ b/c/graphLib/io/strbuf.h
@@ -23,12 +23,12 @@ extern "C"
 
         typedef strBuf *strBufP;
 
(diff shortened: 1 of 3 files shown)


john-boyer-phd (approved):
Welcomed improvements.

PR #284, opened 2026-07-14
Name extension callback parameters

WilliamK112:
## Summary
- add parameter names to extension callback function pointer declarations in `graphExtensions.h`
- mirror those names in the private extension struct callback fields

## Validation
- `git diff --check`
- built `planarity.exe` with GCC from the current C sources

Refs #188


john-boyer-phd (approved):
The names are really helpful since the params must be void pointers.

Issue #178, opened 2025-09-24
Create GitHub Action or update existing GitHub Action(s) to test `planarityApp` endpoints for memory leaks and undefined behaviour


wbkboyer:
In the course of #171 , I noted that the current [`sanitisers.yml`](https://github.com/graph-algorithms/edge-addition-planarity-suite/blob/master/.github/workflows/sanitisers.yml) Github Actions workflow file only runs `-fsanitize=address` and `-fsanitize=undefined` for the `-test` flag (i.e. `runQuickRegressionTests()` in `c/planarityApp/planarityCommandLine.c`, which calls `runSpecificGraphTests()`). This only calls `runSpecificGraphTest()` and `runGraphTransformationTest()`, which although helpful, doesn't provide comprehensive coverage of all `planarityApp` command-line endpoints.

There are a few options:
1. Update the existing `sanitisers.yml` `sanitisers` and `valgrind` `jobs`' `steps` with the more comprehensive tests performed in the sub-tasks of #171 
2. Create a new workflow file that has a `jobs` section for each `planarityApp` command-line endpoint
3. Update `test-samples.sh.in` to include additional test steps (so that no changes need to be made to `sanitisers.yml`)
    * _**N.B.**_ this would *also* impact what is run during the `make distcheck` step of [Making the Distribution](https://github.com/graph-algorithms/edge-addition-planarity-suite?tab=readme-ov-file#making-the-distribution)

wbkboyer:
@john-boyer-phd Should I mark this one as closed, due to @WilliamK112 and my work on #288 as well as the commits [`bc6578c4`](https://github.com/graph-algorithms/edge-addition-planarity-suite/commit/bc6578c40be1b166c0a3caaf88faeca8af08d8ca) (create artifact from `configure.log` on job) and [`43592d0`](https://github.com/graph-algorithms/edge-addition-planarity-suite/commit/43592d0a8e423a32ec6fa752a80a2c1ff3e19e4c) (I figured out that `ubuntu-latest` corresponded to ubuntu 24.04 which only had up to `clang` 18, which didn't have the `-fno-sanitize-merge` flag, hence the [failed job](https://github.com/graph-algorithms/edge-addition-planarity-suite/actions/runs/31330874073/job/93288796712) after merging 288. This meant we explicitly had to use Ubuntu 26.04 image)?

WilliamK112:
[...]

PR #286, opened 2026-07-14
Smoke-test random graph endpoints

WilliamK112:
## Summary
- extend `test-samples.sh.in` to smoke-test the `-r`, `-rm`, and `-rn` random graph CLI endpoints after the existing `-test` run
- verify each endpoint writes nonempty output files
- handle `-rn`'s expected `NONEMBEDDABLE` command-line status while still failing on unexpected statuses

## Testing
- `bash -n test-samples.sh.in`
- `git diff --check`
- direct UCRT GCC build of `planarity.exe`
- generated a temporary configured copy of `test-samples.sh.in` and ran it with MSYS bash
- repeated the new `-r`/`-rm`/`-rn` endpoint smoke checks 5 times

Refs #178

### DIFF
--- a/.github/workflows/compile-warnings.yml
+++ b/.github/workflows/compile-warnings.yml
@@ -1,10 +1,10 @@
 name: Compiler warnings
 on:
   workflow_dispatch:
-#  pull_request:
-#    branches: [ '**' ]
-#    paths-ignore:
-#      - '**/*.md'
+  pull_request:
+    branches: [ '**' ]
+    paths-ignore:
+      - '**/*.md'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/sanitisers.yml
+++ b/.github/workflows/sanitisers.yml
@@ -1,10 +1,10 @@
 name: Sanitisers
 on:
   workflow_dispatch:
-#  pull_request:
-#    branches: [ '**' ]
-#    paths-ignore:
-#      - '**/*.md'
+  pull_request:
+    branches: [ '**' ]
+    paths-ignore:
+      - '**/*.md'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -1,6 +1,6 @@
 name: Spelling
-# on: [pull_request, workflow_dispatch]
-on: [workflow_dispatch]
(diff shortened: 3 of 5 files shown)


john-boyer-phd (changes requested):
I really like the sentiment of this PR, William, which is to test the random graph generator APIs via invoking the endpoints that call them.

We would, however, need a different implementation (which you have already shown that you know how to do easily). The issue is that this project has been adopted as a package into many versions of Linux, and in those cases we don't have write access within the planarity directory, so the script would fail. (We might also get a security ding for using $$.)

Due to the above issue, `planarity -test` never writes to disk (because we also can't write to the samples directory). 

Although the planarity executable can be useful to end-users, it's primary purpose is really to demo and test the capabilities offered by the graphLib API. So, it wouldn't be as important to test the CLI code leading up to the API calls as it would to test the APIs themselves. To that end, a new subfunction `runRandomGraphTests()` could be created to test `gp_CreateRandomGraph()` and especially the new `gp_CreateRandomGraphEx()` you've written. 

I was inspired by this work you have done, so I made some edits to `RandomGraphs()` to make it possible to invoke it from a `runRandomGraphTests()` and have it behave entirely silently (forceQuiet) and test both `gp_CreateRandomGraph()` and especially the new `gp_CreateRandomGraphEx()` (useExGenerator == FALSE or TRUE). Doing 1000 20-vertex graphs for each call ought to suffice. 

It turns out that I have now already forced `gp_CreateRandomGraphEx()` to run on lots of graphs of various sizes by just setting `useExGenerator = TRUE` before the `K` loop in `RandomGraphs()`, so your new implementation has been heavily stress-tested now, including on graphs with fewer than 3N-6 edges (so a better stress test than using `-rm` because the optional edges feature is activated).

If you were to roll back your change to test-samples.sh.in and then do the above instead, an additional advantage would be that the `planarity -test` would include testing the random graph API endpoints, even when running with the sanitizers.

WilliamK112:
Thanks, John. I removed the filesystem-writing shell smoke tests and added runRandomGraphTests() to planarity -test. It runs 1,000 20-vertex graphs through gp_CreateRandomGraph() and another 1,000 through gp_CreateRandomGraphEx(), using forceQuiet in both cases. I also merged the latest master and verified the full autotools build and ./test-samples.sh locally. The update is pushed in commit 65ed7d3.

john-boyer-phd (approved):
Looks great, thank you William!

Issue #56, opened 2024-04-22
`strOrFile` container refactor - `remove()` _output_ file after closing in failure state


wbkboyer:
When outputting the results of transforming an input graph to `.g6` format or testing all graphs in a `.g6` input file (see #10 ), we use a `strOrFile` container to reduce code duplication for some of the flows that are also used when we are outputting to string. It is desirous that the output file be removed from the system so that incomplete file contents are not left after an error is encountered during these flows.

This refactor will require the `strOrFile` container to "know" the file access flag (see [`fopen`](https://en.cppreference.com/w/c/io/fopen) docs), as well as the filename (i.e. [`remove`](https://en.cppreference.com/w/c/io/remove) requires the `pathname`).

One thorn might be the question of what happens when you try to `remove()` the output file if `fclose()` has failed (e.g. if there's a disk error); from the [`remove`](https://en.cppreference.com/w/c/io/remove) docs:
> If the file is currently open by any process, the behavior of this function is implementation-defined. POSIX systems unlink the file name (directory entry), but the filesystem space used by the file is not reclaimed while it is open in any process and while other hard links to the file exist. Windows does not allow the file to be deleted in such cases.

wbkboyer:
Should combine with the work of #187

wbkboyer:
Thanks to @WilliamK112 's excellent work (see #287 and commit [85f3cfa](https://github.com/graph-algorithms/edge-addition-planarity-suite/commit/85f3cfae5111656b20e5c3ee5bd6607b88a13485) with the final result of our combined work coming in commit [8975a84](https://github.com/graph-algorithms/edge-addition-planarity-suite/commit/8975a8487aa4d5767050279824e3db6d11239d5a)), I think this issue is done and dusted!

As an aside, I'm a bit cheezed off that the merge commit didn't list both of us as contributors, hence why I'm giving the credit here! 😭

PR #287, opened 2026-07-14
Remove failed strOrFile output files

WilliamK112:
## Summary
- track ordinary file names and modes in `strOrFile` output containers
- add a failure flag so failed output containers discard partial strings and remove partial ordinary output files after closing
- mark G6 writer output as failed when initialization or write validation fails after the output container has been created

## Validation
- reproduced the current failure on `master`: directed adjacency-list to G6 failed but left a header-only `.g6` file
- verified the same directed-to-G6 transform still fails but no longer leaves the partial `.g6` file
- verified successful G6 transform still writes expected output
- verified `planarity -r -q -p 1 5` still writes a nonempty G6 output file
- verified failed directed-to-G6 `gp_WriteToString()` leaves the output string `NULL` with a temporary local C smoke test
- direct UCRT GCC build of `planarity.exe`
- `planarity.exe -test ./c/samples/`
- `git diff --check`

Refs #56


wbkboyer (approved):
@WilliamK112 Another excellent contribution! You are positively crushing it 💪🏼 I made some slight changes, additions, and removing some vestigial code, then did some testing with the induced error state to verify that we clean up the output file when we encounter an error from `RandomGraphs()` (as well as ran `TestAllGraphs()` for N=8 just to make sure there were no other issues with input or output containers). Ready to ROCK 🚀

Issue #179, opened 2025-10-02
Use `UndefinedBehaviorSanitizer` to test for undefined behaviour during execution via endpoints of `planarityApp` layer


wbkboyer:
This issue is a continuation of the work of #171 .

In the [`sanitisers.yml`](https://github.com/graph-algorithms/edge-addition-planarity-suite/blob/master/.github/workflows/sanitisers.yml) Github Actions workflow file, the `configure` step of the `autotools` pipeline is run for both `-fsanitize=address` and `-fsanitize=undefined`. The latter refers to the `UndefinedBehaviorSanitizer`:
* [`UndefinedBehaviorSanitizer` for `clang`](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)
* [`UndefinedBehaviorSanitizer` for `gcc`](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dundefined)

Both sets of documentation include a list of suboptions, but I think running with `-fsanitize=undefined` will run all `UndefinedBehaviourSanitizer` checks.

The `clang` documentation describes possible `UBSAN_OPTIONS`; of note is the section [Stack traces and report symbolization](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#id9), which indicates additional compiler flags must be set to use `UBSAN_OPTIONS=print_stacktrace=1`:
1. Compile with `-g`, `-fno-sanitize-merge` and `-fno-omit-frame-pointer` to get proper debug information in your binary.
2. Run your program with environment variable `UBSAN_OPTIONS=print_stacktrace=1`.
3. Make sure `llvm-symbolizer` binary is in PATH.

For this last point, I found in #171 that I had to explicitly specify the `ASAN_SYMBOLIZER_PATH` on MacOS as the `llvm-symbolizer` installed using the [`homebrew` formula for `llvm`](https://formulae.brew.sh/formula/llvm) *even though* `llvm-symbolizer` was on my path; for an intentional leak introduced in `planarity.c`:

* With `ASAN_OPTIONS=detect_leaks=1`
    ```
    Direct leak of 1 byte(s) in 1 object(s) allocated from:
        #0 0x000102ffb478 in malloc+0x70 (libclang_rt.asan_osx_dynamic.dylib:arm64+0x53478)
        #1 0x00010280c880 in main planarity.c:18
[...]

PR #288, opened 2026-07-14
Improve UBSan sanitizer diagnostics

WilliamK112:
## Summary
- add the UBSan-only `-fno-sanitize-merge` compile flag while preserving the existing `-g` and `-fno-omit-frame-pointer` sanitizer CFLAGS
- run the undefined-sanitizer job with `UBSAN_OPTIONS=print_stacktrace=1:log_path=ubsan.log`, so logged UBSan findings include stack traces when `llvm-symbolizer` is on `PATH`
- keep the existing sanitizer test surface unchanged

## Testing
- `git diff --check`
- parsed `.github/workflows/sanitisers.yml` with PyYAML and asserted the updated UBSan matrix values
- `clang -x c - -fsanitize=undefined -fno-sanitize-merge -fno-omit-frame-pointer -g -O1 -c` on an inline smoke program
- full local `autoreconf && configure && make && ./test-samples.sh` was not run because this MSYS install does not include `autoreconf`; the hosted sanitizer workflow remains the end-to-end validation path

Reference: https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#stack-traces-and-report-symbolization

Refs #179


wbkboyer:
This is yet another excellent contribution; the original intent of #179 was to manually test all the `planarityApp` endpoints using the undefined behaviour sanitizer, but it'll be great to add this to our raft of automated tests, especially since you're `cat`-ing the `ubsan.log` so they appear in the job logs!

--- 

Just a note about your development environment: if you follow the steps in [2. Dev Setup](https://github.com/graph-algorithms/edge-addition-planarity-suite/wiki/2.-Dev-Setup#installing-dev-dependencies), then on Windows with MSYS2, you can install `autotools` using `pacman`: `pacman -S "${MINGW_PACKAGE_PREFIX}-autotools"`.

HOWEVER, after some digging, I don't believe ASAN nor UBSAN are available on Windows via MSYS2 with `gcc`, which I discovered when I tried to execute the `configure` script by appending to the `CFLAGS` (I can't find a reference to `-fno-sanitize-merge` in `gcc` [Instrumentation Options](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fsanitize_003dundefined) docs page, so I didn't include it; might be a `clang`-specific flag only):

```
autoreconf -fi
...

./configure CFLAGS="-fsanitize=address,undefined -g -fno-omit-frame-pointer $CFLAGS"
configure: loading site script /etc/config.site
checking for a BSD-compatible install... /usr/bin/install -c
checking whether sleep supports fractional seconds... yes
checking filesystem timestamp resolution... 0.01
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking xargs -n works... yes
checking whether UID '197609' is supported by ustar format... yes
checking whether GID '197609' is supported by ustar format... yes
checking how to create a ustar tar archive... gnutar
checking for gcc... gcc
checking whether the C compiler works... no
configure: error: in '/c/Users/wbkbo/git/pr_reviews/WilliamK112_edge-addition-planarity-suite':
configure: error: C compiler cannot create executables
See 'config.log' for more details
```

And then in `config.log`:
```
configure:4098: checking whether the C compiler works
configure:4120: gcc -fsanitize=address,undefined -g -fno-omit-frame-pointer    conftest.c  >&5
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lasan: No such file or directory
C:/msys64/ucrt64/bin/../lib/gcc/x86_64-w64-mingw32/16.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lubsan: No such file or directory
collect2.exe: error: ld returned 1 exit status
```

This seems to be confirmed by [this](https://stackoverflow.com/a/70505139/5456879) stackoverflow response, as well as [this](https://github.com/msys2/MINGW-packages/issues/11899#issuecomment-2630715052) comment on an issue about UBSAN support on the `MSYS2/MINGW-packages` repo.

I ran through the following steps on my Debian Linux VM (I decided to take a shortcut and to skip making the distribution):

```
autoreconf -fi
...
./configure CFLAGS="-fsanitize=undefined -g -fno-omit-frame-pointer $CFLAGS"
...
make
...
UBSAN_OPTIONS=print_stacktrace=1:log_path=ubsan.log ./test-samples.sh
...
All tests have succeeded.
```

Then, in the `ubsan.log` file produced after artificially introducing a signed integer overflow UB:
```
$ cat ubsan.log.11605 
c/planarityApp/planarity.c:18:7: runtime error: signed integer overflow: 2147483647 + 1 cannot be represented in type 'int'
    #0 0x56000b9eb6ac in main c/planarityApp/planarity.c:18
    #1 0x7f744e035ca7  (/lib/x86_64-linux-gnu/libc.so.6+0x29ca7) (BuildId: c495b62edadd6c356265942ec1282d98058a7b41)
    #2 0x7f744e035d64 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29d64) (BuildId: c495b62edadd6c356265942ec1282d98058a7b41)
    #3 0x56000b9eb5a0 in _start (/home/wbkboyer/git/pr_reviews/WilliamK112_edge-addition-planarity-suite/.libs/planarity+0x75a0) (BuildId: 297f4ffcba16a34f08351286e508726cbc4a2db9)

```

(Note that I disabled ASAN running alongside UBSAN; see [Bug 94328 - Logging of defects to file does not work with Asan and Ubsan combined](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94328))

wbkboyer (approved):
Not sure why the jobs are still queued and not running, but I think these changes make sense 🎉

PR #294, opened 2026-08-09
Add directed graph transpose API

### LINKED ISSUE #293: Add a method `gp_TransposeDirectedGraph()`
The goal of this issue is to add a new API endpoint that can be useful in helping to compute the strongly connected components of a digraph. The method signature should be `int gp_TransposeDirectedGraph(graphP theGraph)` and its main behavior should be to simply reverse the direction of every in-use directed edge. Undirected edges should be unaffected by this method. 

The comment above the method should say that it can be used along with a directed depth-first search to help compute the strongly connected components of a digraph.

The new method should be added to `graph.c/.h`, immediately below `gp_ClearEdgeDirectionFlags()`. The method should return `NOTOK` if `theGraph` received is `NULL` and otherwise should return `OK` after completing the main behavior. Edge array iteration loop examples appear throughout the code (look for one that calls `gp_EdgeInUse()` and increments `e += 2`). As an alternative, please just see the wiki for the example.

It would be useful for the test described next to write a little temporary code (such as in `SpecificGraph()`) to `gp_Read()` the file `Petersen.digraph.txt` (from the samples directory), change edge `(5 -> 1)` to be undirected, then call `gp_TransposeDirectedGraph()` , followed by invoking `gp_Write()` to put the result into a file called `Digraph.transposeTest.txt`. Manually check if you got the output you expected. (Then you can remove the temporary code that generated `Digraph.transposeTest.txt`).

Once you have a good `Digraph.transposeTest.txt` file, please add it to the samples directory and to the list of samples files in the `makefile.am` in the samples directory.

Last but not least, please then add a new test submethod into `planarityCommandLine.c`, named `testDigraphTranspose()` and called from `runDigraphTests()`. The new `testDigraphTranspose()` method can simply read the existing sample Digraph.transposeTest.txt` file, change edge `(5 -> 1)` to be undirected, invoke `gp_TransposeDirectedGraph()` and then use the `gp_WriteToString()` and `TextFileMatchesString()` methods to ensure that the directed edges are as expected.

tomatotomata:
## What changed

- add `gp_TransposeDirectedGraph()` to reverse each in-use directed edge while leaving undirected edges unchanged
- add the requested Petersen transpose fixture and command-line regression coverage, including NULL handling
- register the new sample in `c/samples/Makefile.am`

## Validation

- built the complete C suite with MSVC `cl`
- ran `planarity.exe -test -q c\samples` successfully

Fixes #293


john-boyer-phd:
FYI @tomatotomata , the Sanitisers issue is not a problem with your submission. We just added the undefined behavior check to the `sanitisers.yml`, but it looks like `ubuntu-latest` is not quite happy with it. So, I'm going to pull that out and commit that change to your PR, so all the tests that normally pass on all other PRs can pass on yours, too :-) Then, I can proceed with reviewing your actual work, which a brief scan shows to be in very good shape.

Edit: @wbkboyer and I figured out that `-fno-sanitize-merge` was added to `clang` 20 (according to the `tags` associated with [`llvm/llvm-project` commit `5bb650`](https://github.com/llvm/llvm-project/commit/5bb650345d83669434713146aaa431c1f7ad43d6), `20.1.0-rc1`), but the `ubuntu-latest` image (Ubuntu 24.04) only includes up to `clang` 18.1.3 (see [`Ubuntu2404-Readme.md`](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md)), so we upped the job to Ubuntu-26.04 (which includes `clang` 20), and that sanitizers are all good now.

john-boyer-phd (approved):
This was a perfect PR submission, thank you!

PR #279, opened 2026-07-13
Improve gp_CreateRandomGraphEx generation

WilliamK112:
## Summary
- Replace the random-tree triangulation in `gp_CreateRandomGraphEx()` with a face-splitting planar core generator.
- Build exact-size connected planar subgraphs for requests below `3N-6`, then add random unique edges for requests above the maximal planar edge count.
- Remove the obsolete tree-walk helpers that only supported the old generator.

Fixes #235

## Testing
- `git diff --check HEAD~1..HEAD`
- `gcc -O2 -Wall -Wextra -Werror -Wno-unused-parameter -o .\planarity.exe <planarityApp + graphLib sources>`
- `gcc -O2 -DUSE_0BASEDARRAYS -Wall -Wextra -Werror -Wno-unused-parameter -o .\planarity-0based.exe <planarityApp + graphLib sources>`
- Temporary exactness harness, default and `USE_0BASEDARRAYS`: checked N={3,4,5,8,20} across targets near `N-1`, `3N-6`, `3N-5`, and complete graph size for exact `M`, unique edges, and connectivity.
- `.\planarity.exe -test .\c\samples\`
- `.\planarity.exe -rm 20 ...` created 54 edges and reported planar.
- `.\planarity.exe -rn 20 ...` created 55 edges and reported non-planar (app exits 1 for non-embeddable result).
- `.\planarity-0based.exe -rm 20 ...` created 54 edges and reported planar.
- `.\planarity-0based.exe -rn 20 ...` created 55 edges and reported non-planar (app exits 1 for non-embeddable result).
- `.\planarity-0based.exe -r -3 100000 20`
- `.\planarity-0based.exe -r -4 100000 20`

PR #297, opened 2026-08-12
Refactor add edge methods through insert edge methods

### LINKED ISSUE #295: Minor refactor of add and insert edge methods
Recently, we made `gp_DeleteEdge()` overloadable by extensions. For the sake of parallelism, we would also like to have an overloadable `gp_InsertEdge()`, in case extensions would like to perform extra behaviors based on extra edge data they may be maintaining. In preparation for this possibility, the following tasks are needed:

 - [ ] Change `gp_AddEdge()` so that it computes the parameters needed to simply call `gp_InsertEdge()`. It is reasonable to do the parameter checks in the first two `if` statements, but then just use the valid parameters to compute the additional parameter values needed to call `gp_InsertEdge()` and then return the result of calling it.
 - [ ] Similarly, change `gp_DynamicAddEdge()` to call `gp_DynamicInsertEdge()`. This will have the desirable side effect of eliminating duplicate code in `gp_DynamicAddEdge()` that grows the edge capacity.
 - [ ] Change the implementation of `gp_DynamicInsertEdge()` so that it does not invoke `gp_InsertEdge()` twice. Instead, first ensure `theGraph` is not `NULL`, remove the first `gp_InsertEdge()` call, and replace the `if (Result == AT_EDGE_CAPACITY_LIMIT)` condition with the edge capacity test (`if (gp_GetM(theGraph) >= theGraph->edgeCapacity)`). Then, the second `gp_InsertEdge()` call can just be moved to be what the method returns.

Once these changes have been made, there is also an opportunity to be more precise about when `gp_InsertEdge()` should return `AT_EDGE_CAPACITY_LIMIT` and therefore also when `gp_DynamicInsertEdge()` should grow the edge capacity. If the `edgeHoles` stack is non-empty, then the edge record to use is already available, so there is no need to return `AT_EDGE_CAPACITY_LIMIT` in `gp_InsertEdge()`, so the current conditional test of edge capacity should be moved to an `else if` clause that attaches after the `if` that tests the `edgeHoles` stack and before the `else` clause that assigns a value to `vpos`. Likewise, in `gp_DynamicInsertEdge()`, the condition that gates whether to grow edge capacity should also test `sp_IsEmpty()` on the `edgeHoles` (because there is no reason to grow edge capacity if the `edgeHoles` stack already has an edge record pair that can be used for the new edge).

To test that the work above, works, it is only necessary to run `planarity -test` on the samples directory (such as by uncommenting `-test` and samples directory parameters in the `launch.json` file).

WilliamK112:
## Summary

- route `gp_AddEdge()` through `gp_InsertEdge()` and `gp_DynamicAddEdge()` through `gp_DynamicInsertEdge()`
- grow dynamic edge capacity only when the graph is full and has no reusable edge hole
- let `gp_InsertEdge()` reuse a deleted-edge hole even when the logical edge count is at capacity
- add regression coverage for capacity growth, single insertion, and full-capacity hole reuse

Fixes #295.

## Validation

- `./configure --enable-compile-warnings CC=clang CFLAGS=-Werror && make -j4`
- `./test-samples.sh`
- complete sample suite under AddressSanitizer
- complete sample suite under UndefinedBehaviorSanitizer
- add/insert regression passes with `USE_0BASEDARRAYS` (the later digraph transpose test also fails unchanged on `origin/master` in that configuration)
- `git diff --check`


c/graphLib/graph.c
```diff
@@ -1908,8 +1908,6 @@ void _DetachEdgeRecord(graphP theGraph, int e)
 
 int gp_AddEdge(graphP theGraph, int u, int ulink, int v, int vlink)
 {
-    int upos, vpos;
-
     if (theGraph == NULL ||
```
wbkboyer, line 1917: @john-boyer-phd Should `gp_AddEdge()` follow the same example as `gp_DynamicAddEdge()` and leverage the validation within `gp_InsertEdge()` (so these checks can be removed), or should `gp_DynamicAddEdge()` have these checks added so that we get a `NOTOK` coming from `gp_DynamicAddEdge()` rather than `gp_DynamicInsertEdge()`?

c/graphLib/graph.c
```diff
@@ -1920,93 +1918,40 @@ int gp_AddEdge(graphP theGraph, int u, int ulink, int v, int vlink)
...
-        Result = gp_EnsureEdgeCapacity(theGraph, newEdgeCapacity);
-
-        if (Result != OK)
-            return NOTOK;
-
-        Result = gp_AddEdge(theGraph, u, ulink, v, vlink);
-    }
-
-    return Result != OK ? NOTOK : Result;
+    return gp_DynamicInsertEdge(theGraph, u, NIL, ulink, v, NIL, vlink);
```
wbkboyer: See [this](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/297/changes#r3787635809) comment regarding whether we should add parameter validation to the top of `gp_DynamicAddEdge()`

c/planarityApp/planarityCommandLine.c
```diff
@@ -261,6 +264,91 @@ int runQuickRegressionTests(int argc, char *argv[])
     return retVal;
 }
 
+int runAddInsertEdgeTests(void)
+{
+    graphP theGraph = gp_New();
```
wbkboyer: For consistency with other functions, please have declaration of `int Result = OK` first, followed by an empty line, and then the declarations and initial assignment of `theGraph` to `NULL`, `initialEdgeCapacity` to 0, `edgeToDelete` to `NIL`, and `v` to `NIL`. Then, just before the `NULL` check of `theGraph`, add `theGraph = gp_New();` 🙏🏼

c/planarityApp/planarityCommandLine.c
```diff
@@ -261,6 +264,91 @@ int runQuickRegressionTests(int argc, char *argv[])
...
+
+    gp_Message("Starting Add/Insert Edge Tests");
+
+    if (theGraph == NULL)
+    {
+        gp_ErrorMessage("Unable to allocate graph for add/insert edge tests.");
+        return NOTOK;
+    }
+
+    if (gp_DynamicInsertEdge(theGraph, 0, NIL, 0, 0, NIL, 0) != NOTOK)
```
wbkboyer: This is a great check! However, when compiled with `-DDEBUG`, a `NOTOK` macro expansion will erroneously occur (see `appconst.h`)... Also, since `gp_DynamicInsertEdge()` (and other public API members) all do a coarse-grained initialization check of `theGraph` (i.e. ensure it's not `NULL` and that `N` is strictly greater than zero), I think this check can safely be removed, and in the future we can set up a more comprehensive and focused test suite of the public API guards!

c/planarityApp/planarityCommandLine.c
```diff
@@ -261,6 +264,91 @@ int runQuickRegressionTests(int argc, char *argv[])
...
+        gp_ErrorMessage("gp_DynamicInsertEdge() accepted an uninitialized graph.");
+        Result = NOTOK;
+    }
+
+    if (Result == OK &&
+        (gp_EnsureEdgeCapacity(theGraph, 2) != OK ||
+         gp_EnsureVertexCapacity(theGraph, 3) != OK))
+    {
+        gp_ErrorMessage("Unable to initialize graph for add/insert edge tests.");
+        gp_Free(&theGraph);
```
wbkboyer, line 286: It might be cleaner for all the subsequent `if` statements that you `gp_Free(&theGraph);` and `return NOTOK;` rather than assign `Result = NOTOK` and have to keep checking `if (Result == OK && ...)` for every step of the test. Then, the `Result` variable can be removed, and at the very end of the function, you only have to write:

```
gp_Message("Finished Add/Insert Edge Tests.\n");

gp_Free(&theGraph);

return OK;
```

Either that, or you can keep the `Result` and use the `goto` pattern you introduced in #279 :

```
    if (<some test step> != OK)
    {
        gp_ErrorMessage("<description of what step failed>");
        Result = NOTOK;
        goto gp_runAddInsertEdgeTests_Cleanup;
    }

...

gp_runAddInsertEdgeTests_Cleanup:

    if (Result == OK)
        gp_Message("Finished Add/Insert Edge Tests.\n");
    else
        gp_ErrorMessage("Add/Insert Edge Tests failed.\n");

    gp_Free(&theGraph);

    return Result;
```

c/planarityApp/planarityCommandLine.c
```diff
@@ -261,6 +264,91 @@ int runQuickRegressionTests(int argc, char *argv[])
...
+        gp_InsertEdge(theGraph, v, NIL, 1, v + 2, NIL, 1) != AT_EDGE_CAPACITY_LIMIT)
+    {
+        gp_ErrorMessage("gp_InsertEdge() did not report the edge capacity limit.");
+        Result = NOTOK;
+    }
+
+    if (Result == OK &&
+        (gp_DynamicInsertEdge(theGraph, v, NIL, 1, v + 2, NIL, 1) != OK ||
+         gp_GetEdgeCapacity(theGraph) <= initialEdgeCapacity ||
+         gp_GetM(theGraph) != 3))
```
wbkboyer: This is an excellent test of `gp_InsertEdge()` vs. `gp_DynamicInsertEdge()`!! I think it might be sufficient to check that `gp_GetM(theGraph)` is now 3, since you only deal with edge holes in the subsequent check...

c/planarityApp/planarityCommandLine.c
```diff
@@ -261,6 +264,91 @@ int runQuickRegressionTests(int argc, char *argv[])
...
+
+    edgeToDelete = Result == OK ? gp_FindEdge(theGraph, v, v + 1) : NIL;
+    if (Result == OK &&
+        (edgeToDelete == NIL || gp_DeleteEdge(theGraph, edgeToDelete) != OK))
+    {
+        gp_ErrorMessage("Unable to create an edge hole for add/insert edge tests.");
+        Result = NOTOK;
+    }
+
+    initialEdgeCapacity = gp_GetEdgeCapacity(theGraph);
```
wbkboyer, line 325: Unless you change the logic of what happens in an error state (see [this](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/297/changes#r3787727278) comment), you might want to use the same approach as you did on ln. 325, i.e. 

```
    initialEdgeCapacity = (Result == OK) ?gp_GetEdgeCapacity(theGraph) : 0;
```

Even though technically speaking you won't be using this value in the following `if` statement because of short-circuit evaluation (i.e. first conjunct will be falsy)

c/planarityApp/planarityCommandLine.c
```diff
@@ -261,6 +264,91 @@ int runQuickRegressionTests(int argc, char *argv[])
...
+    if (Result == OK &&
+        (gp_DynamicInsertEdge(theGraph, v, NIL, 1, v + 2, NIL, 1) != OK ||
+         gp_GetEdgeCapacity(theGraph) <= initialEdgeCapacity ||
+         gp_GetM(theGraph) != 3))
+    {
+        gp_ErrorMessage("gp_DynamicInsertEdge() did not grow edge capacity and insert once.");
+        Result = NOTOK;
+    }
+
+    edgeToDelete = Result == OK ? gp_FindEdge(theGraph, v, v + 1) : NIL;
```
wbkboyer: Please parenthesize the condition `(Result == OK)` 🙏🏼

wbkboyer (dismissed):
Just a few small tweaks regarding control flow and variable initialization, pending feedback from @john-boyer-phd !

c/graphLib/graph.c
john-boyer-phd: Ah, I see the inconsistency that you're noticing, even though I had originally said in the first bullet that it would be OK to leave these checks here. I also like your point that we want a `NOTOK` (and therefore eventually a `gp_ErrorMessage()` to be emitted from within this method so that the API caller gets feedback from the actual method they called. To rationalize these competing concerns, it would be best if this location said

```
    if (gp_InsertEdge(...) != OK)
        return NOTOK;

    return OK;
```

c/graphLib/graph.c
john-boyer-phd: And then this should say:

```
    if (gp_DynamicInsertEdge(...) != OK)
        return NOTOK;

    return OK;
```

c/graphLib/graph.c
```diff
@@ -2026,15 +1971,11 @@ int gp_DynamicInsertEdge(graphP theGraph, int u, int e_u, int e_ulink,
...
-
-        if (Result != OK)
+        if (gp_EnsureEdgeCapacity(theGraph, newEdgeCapacity) != OK)
             return NOTOK;
-
-        Result = gp_InsertEdge(theGraph, u, e_u, e_ulink, v, e_v, e_vlink);
     }
 
-    return Result != OK ? NOTOK : Result;
+    return gp_InsertEdge(theGraph, u, e_u, e_ulink, v, e_v, e_vlink);
```
john-boyer-phd: And last but not least, 

```
    if (gp_InsertEdge(...) != OK)
        return NOTOK;

    return OK;
```

c/planarityApp/planarityCommandLine.c
john-boyer-phd: I think we have a lot of places where `Result`/`RetVal`/`'retVal` are not initialized first so maybe we can avoid opening that can of worms for now. I think what most likely jumped was calling something non-trivial like `gp_New()` in variable initialization. There's nothing really 'wrong' with it, but it would be preferable, because it's all over the place in the library already, to do this instead:

```
    int Result = OK;
    graphP theGraph = NULL;

    ....

    if ((theGraph = gp_New()) == NULL)
    {
        gp_ErrorMessage(...);
        return NOTOK;
    }
```

c/planarityApp/planarityCommandLine.c
john-boyer-phd: Agreed the `!= NOTOK` can be used unfortunately because `NOTOK` still has a definition that causes it to emit an error in debug mode when there isn't an error. 

Moreover, the expected/desired `NOTOK` in `gp_DynamicInsertEdge()` also emits an error message in debug mode, so regrettably this test has to be removed.

c/planarityApp/planarityCommandLine.c
john-boyer-phd: Though the code's not wrong as-is, either the `gp_Free()`-and-`return NOTOK` or the `goto` method would be a bit cleaner. Since you are making other changes, would you mind please changing to one of these methods here?

c/planarityApp/planarityCommandLine.c
john-boyer-phd: We have other places where the condition is not parenthesized, relying instead on order of operations. That being said, I've also seen lots of code style guidelines that say it's best not to count on order of operations too much because people get it wrong sometimes. So, keep us honest, Wanda!

c/planarityApp/planarityCommandLine.c
john-boyer-phd: I agree that no change need be made here if the return methodology in this function is changed to either the `gp_Free()`-and-`return NOTOK` or the `goto` method.

c/planarityApp/planarityCommandLine.c
john-boyer-phd: I think it would be better to retain the `gp_GetEdgeCapacity(theGraph) <= initialEdgeCapacity` condition. This way, we have direct tests for both things that we expect from `gp_DynamicInsertEdge()`, namely that it must not only add the edge (M==3 test) but it must allocate the extra space to store the extra edge (as opposed to walking on unowned memory and happening not to crash).  So, once you see this comment William, it would be reasonable to hit "Resolve Comment" without making a code change.

john-boyer-phd (changes requested):
Hi William,
A few small changes in this review. When you finish and commit them, please hit "Resolve comment" on the comments that you made changes for, and please hit "Request re-review".

WilliamK112:
@john-boyer-phd @wbkboyer Applied the review notes in 5620615:

- `gp_AddEdge` / `gp_DynamicAddEdge` / `gp_DynamicInsertEdge` now wrap the delegated call as `if (... != OK) return NOTOK; return OK;` so a `NOTOK` (and debug error message) comes from the API the caller invoked.
- `runAddInsertEdgeTests()` uses `theGraph = NULL` then `if ((theGraph = gp_New()) == NULL)`, drops the `!= NOTOK` uninitialized-graph check (unsafe under `-DDEBUG`), and early-returns after `gp_Free` instead of chaining `Result == OK &&`.
- Kept the `gp_GetEdgeCapacity(...) <= initialEdgeCapacity` plus `M == 3` check as requested.

Also merged latest `master`. Please re-review when you have a chance.

john-boyer-phd (approved):
All good, thanks for making the changes, William.

PR #298, opened 2026-08-12
Add basic GraphML graph writer

### LINKED ISSUE #296: GraphML Writer, Part 1
The GraphML format has exceptional representational power that we would like to benefit from over time. To begin the journey of supporting GraphML, we start with adding the capability to write the basic vertex and edge structure of a graph.

The new GraphML writer should be placed in a new file `graphML-writer.c` that should be placed in the `io` subdirectory. The file will need to start with the copyright message, any needed header includes, and a "Private functions (exported to system)" section that gives function signatures for the methods that will be in `graphML-writer.c`. The filename `graphML-writer.c` will need to be added to the `makefile.am`, just below the line for `g6-write-iterator.c`.  

Add `#define WRITE_GRAPHML 5` to `graphIO.h` just below the declaration of `WRITE_G6`. This will give a new `writeMode` for `gp_Write()` and `gp_WriteToString()` to detect.

Add a new `int _WriteGraphMLGraph(graphP theGraph, strOrFileP outputContainer)` method into `graphML-writer.c`. It will be called by `gp_Write()` and `gp_WriteToString()` in response to receiving a `writeMode` of `WRITE_GRAPHML`, so add the extern function signature to the "Imported functions" section of `graphIO.c`. The new method will be returning `OK` for success or `NOTOK` on any failure, so that how it is called by `_WriteGraph()` (which `gp_Write()` and `gp_WriteToString()` call) will match how all of the other writers are called.

## Writing the `<graphml>` Element

Within `graphML-writer.c`, the new method `_WriteGraphMLGraph()` should invoke three subroutines, in order, given by the following signatures:

 - [ ] `int _WriteGraphMLStartTag(strOrFileP outputContainer)`
 - [ ] `int _WriteGraphMLGraphElement(graphP theGraph, int index, strOrFileP outputContainer)`
 - [ ] `int _WriteGraphMLEndTag(strOrFileP outputContainer)`
 
These are specified as subroutines so that they can be called differently when we eventually build a GraphML write iterator. Each will return `OK` for success or `NOTOK` on any failure. Each receives a `strOrFileP` parameter so they can be called by both `gp_Write()` and `gp_WriteToString()`. The `index` passed to `_WriteGraphMLGraphElement()` should be 1 (a future GraphML write iterator will need to pass varied indexes for each graph written). 

The `_WriteGraphMLStartTag()` method just has to write the following string literal (including linefeed after each line shown):

```
<?xml version="1.0" encoding="UTF-8"?>
<graphml xmlns="http://graphml.graphdrawing.org/xmlns"  
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://graphml.graphdrawing.org/xmlns
     http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd">
```

The `_WriteGraphMLEndTag()` method just has to write the following string literal (including linefeed afterward):

```
</graphml>
```

## Writing the `<graph>` Element

The `_WriteGraphMLGraphElement()` should also be written using four subroutines:

 - [ ] `int _WriteGraphMLGraphStartTag(graphP theGraph, int index, strOrFileP outputContainer)`
 - [ ] `int _WriteGraphMLGraphVertices(graphP theGraph, int index, strOrFileP outputContainer)` 
 - [ ] `int _WriteGraphMLGraphEdges(graphP theGraph, int index, strOrFileP outputContainer)` 
 - [ ] `int _WriteGraphMLGraphEndTag(graphP theGraph, int index, strOrFileP outputContainer)`

The `_WriteGraphMLGraphStartTag()` method should write a string in this format (including two starting spaces and linefeed after):

```
  <graph id="G%d" edgedefault="undirected">
```

The `%d` is to be replaced by the graph's `index`, received as a parameter to the method. The `outputContainer` object doesn't have a `printf`, so it will be necessary to use `sf_fputs()` on the substring to the left of the `%d` (including the leading two spaces), then write the integer `index`, then use `sf_fputs()` on the substring to the right of the `%d` (including the trailing linefeed).

The need to write an integer to the `outputContainer` will come up a few times, so we may as well do it the "right" way:

 - [ ] Add a method with signature `int sf_WriteInteger(int intToWrite, strOrFileP theStrOrFile)` just below `sf_fputs()` in `strOrFile.c`/`.h`
 - [ ] Return `OK` on success or `NOTOK` on any errors.
 - [ ] Use `sf_fputs()` for inspiration, such as parameter validation and switching between string or file.
 - [ ] Use `fprintf()` for the file version, and use `sb_ConcatInt()` for the string version.
 - [ ] Use `sf_WriteInteger()` to write the graph index in `_WriteGraphMLGraphStartTag()`

The `_WriteGraphMLGraphEndTag()` method should simply output the following end tag (including the two leading spaces and trailing linefeed).

```
  </graph>
```

## Writing the Vertices (the `node` Elements)

For `_WriteGraphMLGraphVertices()`, use a vertex iteration loop (up to `gp_UpperBoundVertices()`) to write a string of the following format (including the leading 4 spaces and the following linefeed):

```
    <node id="n%d"/>
```

The `%d` is to be replaced by the index of the vertex (the value of the vertex iteration loop variable), except that if the graph flags indicates that `GRAPHFLAGS_ZEROBASEDIO` flag is set, then subtract `gp_LowerBoundVertexStorage()` from the integer value to be emitted (see `_WriteAdjList()` for reference).

## Writing the Edges

For `_WriteGraphMLGraphEdges()`, use an in-use edge iteration loop (with incrementation of `e += 2`) to write a line of markup for each edge, using the first format below for undirected edges and the second format for directed edges:

```
    <edge id="e%d" source="n%d" target="n%d"/>
    <edge id="e%d" source="n%d" target="n%d" directed="true"/>
```

The integer for the edge `id` value, it would be best to run an `edgeID` counter that starts at 1 and is incremented after use outputting an in-use edge. This will ensure that the same IDs are produced whether or not the graph has any edge holes due to edge deletion. 

If the edge is undirected, then the source and target edge records can be set to be the loop iteration variable value and its twin (obtained using `gp_GetTwin()`).  If the edge is directed, then the source edge record is the out-only edge record, and the target edge record is in-only. The source edge record is the edge record contained by the source vertex. To get the source vertex, one must get the neighbor value from the source edge record's twin (which is the target edge record). Similarly, the target vertex is the neighbor value from the source edge record. 

For the `%d` in the `source` and `target` attributes, output the source and target vertex values from above, except subtract `gp_LowerBoundVertexStorage()` from each if the graph flags indicates that `GRAPHFLAGS_ZEROBASEDIO` flag is set.

## Testing

In `planarityCommandLine.c`, add a new method `runGraphMLTests()` and call it from `runQuickRegressionTests()`, just after the call to `runDigraphTests()`. Add the declaration at the end of tuen `run` method declarations at the top of the C file.

Currently, only one test is needed within `runGraphMLTests()`, but it should be encoded as a subroutine `runBasicGraphMLWriteTest()` so that other tests can be added more easily in the future. This subroutine should return `OK` on success or `NOTOK` on any error, and it should perform the following actions:

 - [ ] Use `gp_Read()` to read the sample graph called `Digraph.transposeTest.txt`
 - [ ] Use `gp_WriteToString()` with `writeMode` of `WRITE_GRAPHML`
 - [ ] Use `TextFileMatchesString() to compare to the sample file named `Digraph.transposeTest.graphml`
 
Initially, the test will fail because it is necessary to create the file `Digraph.transposeTest.graphml`, manually check that it comtains the expected GraphML markup, and then place it in the samples directory (and add it to the list of sample files in `c/samples/Makefile.am`).

To create the file, you can temporarily put a `gp_Write()` after your `gp_Read()` to write the content to a file. Then, once you have the file and have manually validated that it is correct, you can add to the samples and then remove the `gp_Write()` from your `runBasicGraphMLWriteTest()` implementation.  

To run the test, perform `planarity -test ./c/samples`, such as by uncommenting the command-line parameters in the `launch.json`.

## GraphML References

See the Official GraphML site here: https://graphml.graphdrawing.org/

Specification: http://graphml.graphdrawing.org/specification.html

Primer: http://graphml.graphdrawing.org/primer/graphml-primer.html

WilliamK112:
## Summary

- add `WRITE_GRAPHML` support to `gp_Write()` and `gp_WriteToString()`
- write GraphML documents through the requested composable start/graph/node/edge/end helpers
- preserve directed edge orientation, emit explicit per-edge direction overrides, and keep edge IDs contiguous across deleted-edge holes
- add `sf_WriteInteger()` for both file-backed and string-backed output containers
- add the checked `Digraph.transposeTest.graphml` sample and quick regression coverage

Fixes #296.

## Validation

- strict Clang warning build (`--enable-compile-warnings`, `CFLAGS=-Werror`)
- `./planarity -test ./c/samples`
- complete sample suite under AddressSanitizer
- complete sample suite under UndefinedBehaviorSanitizer
- `make distcheck`
- `xmllint --noout c/samples/Digraph.transposeTest.graphml`
- 0-based storage build with warnings as errors, plus a focused mixed-graph GraphML smoke test (10 nodes `n0`–`n9`, 15 edges, 14 directed)
- focused file/string output test after deleting an edge (14 continuous edge IDs and valid XML)
- `git diff --check`

## 0-based suite note

The repository's full `USE_0BASEDARRAYS` suite still stops at the existing Digraph transpose assertion before reaching the new GraphML test; the same failure reproduces on unmodified `origin/master`. The focused GraphML test passes in that configuration.


john-boyer-phd:
The 0-based array mode preserves the older implementation, which can only read 0-based files. The 1-based code is newer (relatively speaking, circa 2010), so it can read 1-based and also is backwards compatible with 0-based files. If a test is running on a 1-based file, it should be guarded by a 1-based ifdef. Also, one can add a #else for 0-based as long as a 0-based sample file is created and used instead. Sound ok?

WilliamK112:
Sounds good — addressed in 72324aa. The GraphML regression now uses the 1-based sample only under `USE_1BASEDARRAYS`, with a 0-based sample and expected GraphML output in the alternate branch. I applied the same selection to the adjacent Petersen digraph/transpose regression so the 0-based suite does not read the 1-based fixture before reaching the GraphML test. Both the default and `CPPFLAGS=-DUSE_0BASEDARRAYS` full sample suites pass, including the GraphML tests.

c/graphLib/io/graphML-writer.c
```diff
@@ -0,0 +1,207 @@
...
+ _WriteGraphMLStartTag()
+
+ Writes the XML declaration and GraphML root start tag.
+ ********************************************************************/
+int _WriteGraphMLStartTag(strOrFileP outputContainer)
+{
+    char const *startTag =
+        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+        "<graphml xmlns=\"http://graphml.graphdrawing.org/xmlns\"\n"
+        "    xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\"\n"
```
wbkboyer, line 50: @john-boyer-phd Do we want to use `\t` for indentation, or multiples of 4 spaces?

Edit: In reading the specification, I see that the task asks for spaces for other indentation... I'll resolve this as soon as I'm done with the initial review!

Image

c/graphLib/io/strOrFile.c
```diff
@@ -728,6 +728,33 @@ int sf_fputs(char const *strToWrite, strOrFileP theStrOrFile)
...
 
+/********************************************************************
+ sf_WriteInteger()
+
+ Writes an integer to a string-or-file output container.
+
+ Returns OK on success, NOTOK on failure.
+ ********************************************************************/
+
+int sf_WriteInteger(int intToWrite, strOrFileP theStrOrFile)
```
wbkboyer, line 739: This is a handy utility function!

@john-boyer-phd should this function definition reside up under `sf_ReadInteger()`?

The only other suggestion I have is that you might leverage `sf_fputs()` instead of conditioning on whether the output container holds a string or file, something like:

```
int result = OK;

char intStr[MAXCHARSFOR32BITINT + 1];
memset(intStr, '\0', (MAXCHARSFOR32BITINT + 1) * sizeof(char));

if (sprintf(intStr, "%d", intToWrite) < 1)
    result = NOTOK;

if (result == OK && sf_fputs(intStr, theStrOrFile) == EOF)
    result = NOTOK;

if (result != OK)
    sf_SetOutputErrorFlag(theStrOrFile);

return result;
```

Edit: I now see that the specification requested the approach supplied here, so I don't think changes are necessary other than the `NOTOK` issue below!!

---

If we decide to stick with your current implementation, then you'll have to change how you declare the return code: if the project is compiled with the `-DDEBUG` flag and you declare `int result = NOTOK`, an error message will erroneously be emitted (see `appconst.h` for the macro and `apiutils.c` for the `debugNOTOK()` function), so this should be set to `OK` initially and overridden based on the result of writing.

Currently, on MacOS with `clang` compiled using the VSCode Debug build configuration:
```
% ./Debug/planarity -test c/samples

	Starting 1-based Array Index Tests
...
Starting GraphML Tests
[ERROR] NOTOK occurred
	on line 741 of '/Users/wbkboyer/git/edge-addition-planarity-suite-fork/c/graphLib/io/strOrFile.c'
... (x56, which matches exactly the number of integers written in the sample file)
Finished GraphML Tests.

============

All tests have succeeded.

	Press return key to exit...

File: c/planarityApp/planarityCommandLine.c [hunk 3]
```diff
@@ -1622,23 +1632,39 @@ int testDigraphTranspose(void)
 {
     graphP G = gp_New();
     char *actualOutput = NULL;
+    char const *inputFileName = NULL;
+    char const *expectedOutputFileName = NULL;
+    int preservedEdgeSource;
```
wbkboyer: Please always initialize integers (i.e. `preservedEdgeSource`, `preservedEdgeTarget`, and `e`) to `NIL` (or some other sensible value for what they're supposed to represent) 🎉

c/samples/Makefile.am
```diff
@@ -58,4 +58,8 @@ dist_samples_DATA = \
   Petersen.txt.Outerplanarity.out.txt \
   Petersen.txt.Planarity.out.txt \
   Petersen.digraph.txt \
-  Digraph.transposeTest.txt
+  Petersen.digraph.0-based.txt \
+  Digraph.transposeTest.txt \
```
wbkboyer, line 62: `Digraph.transposeTest.txt` should be added to this PR 🙏🏼 

edit: Oh whoops, just double-checked the repo, and I see [Digraph.transposeTest.txt](https://github.com/graph-algorithms/edge-addition-planarity-suite/blob/master/c/samples/Digraph.transposeTest.txt) was previously added to the repo... Great catch making sure it's added to the `c/samples/Makefile.am`!!

c/graphLib/io/graphML-writer.c
```diff
@@ -0,0 +1,207 @@
...
+        "    xsi:schemaLocation=\"http://graphml.graphdrawing.org/xmlns\n"
+        "     http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd\">\n";
```
wbkboyer, line 52: I noticed there's an extra space preceding the second URL, which was present in the issue specification as well as in the GraphML Primer [2.1 A Simple Graph](http://graphml.graphdrawing.org/primer/graphml-primer.html#2.1-A-Simple-Graph)):

Image

However the [`simple.graphml`](http://graphml.graphdrawing.org/primer/simple.graphml) shows `xsi:schemaLocation` all occurring on one line with a single space between the schema instance locations... Is this something we have to worry about for consuming applications, @john-boyer-phd ?

c/graphLib/io/graphML-writer.c
```diff
@@ -0,0 +1,207 @@
...
+ Writes the graph element start tag.
+ ********************************************************************/
+int _WriteGraphMLGraphStartTag(graphP theGraph, int index, strOrFileP outputContainer)
+{
+    if (theGraph == NULL || !sf_IsValidStrOrFile(outputContainer))
+        return NOTOK;
+
+    if (sf_fputs("  <graph id=\"G", outputContainer) == EOF ||
+        sf_WriteInteger(index, outputContainer) != OK ||
+        sf_fputs("\" edgedefault=\"undirected\">\n", outputContainer) == EOF)
```
wbkboyer, line 98: Originally, I was concerned with whether we should programmatically specify the `edgedefault`, but then checked the documentation: in the [GraphML Primer](http://graphml.graphdrawing.org/primer/graphml-primer.html) section `2.3.1 Declaring a Graph`, the `edgedefault` must be specified, which is applied to an edge if no direction is specified when the edge is declared.

And from [`xsd.html`](http://graphml.graphdrawing.org/specification/xsd.html):

Image

So yep, `undirected` is a good default (since we're able to represent directed, undirected, and mixed graphs) 🎉

c/graphLib/io/graphML-writer.c
```diff
@@ -0,0 +1,207 @@
...
+ _WriteGraphMLGraphVertices()
+
+ Writes one node element for every graph vertex.
+ ********************************************************************/
+int _WriteGraphMLGraphVertices(graphP theGraph, int index, strOrFileP outputContainer)
+{
+    int v = NIL;
+    int zeroBasedVertexOffset = 0;
+
+    (void)index;
```
wbkboyer, line 114: I see you've used this notation a couple times (i.e. in `_WriteGraphMLGraphVertices()`, `_WriteGraphMLGraphEdges()`, and `_WriteGraphMLGraphEndTag()`)... I'm not familiar with this notation, but it seems like it's something to suppress an unused variable (which is a very neat trick for the future!); I noticed in the issue specification that the function signatures all include `index`, but it's only actually used when we're specifying the start of a graph (i.e. `_WriteGraphMLGraphStartTag()`). @john-boyer-phd is `index` passed in only as a reminder to the caller of these functions about which graph the vertices and edges we're specifying belong to?

c/planarityApp/planarityCommandLine.c
```diff
@@ -1687,3 +1713,52 @@ int runDigraphTests(void)
...
+    char *actualOutput = NULL;
+    char const *inputFileName = NULL;
+    char const *expectedOutputFileName = NULL;
+    int Result = OK;
+
+    if (G == NULL)
+        return NOTOK;
+
+#ifdef USE_1BASEDARRAYS
+    inputFileName = "Digraph.transposeTest.txt";
```
wbkboyer, line 1729: Good idea to save on the number of samples that we include with the project by re-using this file! However, may I suggest using the sample graph supplied in the [GraphML Primer](http://graphml.graphdrawing.org/primer/graphml-primer.html#BCfile) in section 2.1, just because we can be fairly certain that GraphML syntax is sound? When I set up the `G6(Read|Write)Iterator`, I used the `nauty_example.g6` from the [nauty & Traces project](https://users.cecs.anu.edu.au/~bdm/nauty/) documentation on supported formats (see [`nug29.pdf`](https://users.cecs.anu.edu.au/~bdm/nauty/nug29.pdf))

wbkboyer (changes requested):
I think the two changes that need to happen are:

1. Don't initialize `int result = NOTOK;` (See [this](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/298/changes#r3779247589) comment)
2. Always initialize variables (See [this](https://github.com/graph-algorithms/edge-addition-planarity-suite/pull/298/changes#r3779284877) comment)

I ran `-test` after building on MacOS using VSCode Debug build configuration, and other than the issue with 1., the tests passed 🎉 Will have to wait for reader + hooking into `TransformGraph()` for more comprehensive testing 💪🏼

c/graphLib/io/strOrFile.c
WilliamK112: Fixed in 7f38042 by initializing `result` to `OK`. I also ran the full sample suite in DEBUG builds for both 1-based and 0-based array modes; both passed, and neither test log contains `[ERROR] NOTOK occurred`.

c/planarityApp/planarityCommandLine.c
WilliamK112: Fixed in 7f38042: `preservedEdgeSource`, `preservedEdgeTarget`, and `e` are now initialized to `NIL`. The DEBUG sample suites pass in both 1-based and 0-based array modes.

WilliamK112:
Addressed the two requested changes in 7f38042: `sf_WriteInteger()` now initializes its result to `OK`, and the transpose-test integers are initialized to `NIL`. I ran the full sample suite in DEBUG builds for both 1-based and 0-based array modes; both passed, and neither log contains `[ERROR] NOTOK occurred`. The two actionable review threads are resolved and the PR is ready for re-review.

wbkboyer (approved):
Right on, thank you @WilliamK112 ! I pulled down your latest changes and compiled both 0- and 1-based arrays (i.e. built both without and with `-DUSE_0BASEDARRAYS` flag), and the `-test`s pass! I also captured those other questions I had on the issue (see [this](https://github.com/graph-algorithms/edge-addition-planarity-suite/issues/296#issuecomment-5298658345) comment), so they won't block merging this PR 🎉

PR #307, opened 2026-08-20
Enable unused parameter warnings

### LINKED ISSUE #299: Change how unused parameter warnings are handled
In the `tasks.json` used by VSCode (which means also the one in the `devEnvSetupAndDefaults` directory, we currently suppress unused parameter warnings on all compile jobs using `-Wno-error=unused-parameter`, `-Wno-used-parameter`, and (for Microsoft `cl`) `/wd4100`. The goal of this issue is to get rid of these warning suppressions and instead amend the code in a way that suppresses the warnings (by keeping the unused parameter but flagging it).

It is currently not clear how `makefile.am` and `compiler-warnings.yml` turn on all of the warnings flags we currently set in `tasks.json` (from `-Wall` through `Wunused-result`). So, it would help to investigate that to find out (A) do we need another issue to turn on more warnings than we current ask for in the make file, and (B) do we need something to turn off the suppression of unused parameter warnings in the make file, as we are seeking to do in `tasks.json` for this issue.

In the file `graphML-writer.c`, there are three occurrences of the line `(void)index;` that were added to suppress unused parameter warnings. Although those may be deleted in the future, in favor of not having the unused parameters, the technique can be used to suppress several unused parameter warnings in the rest of the graph library. For each location where this technique is used,  add a comment on the line above to say `// Suppresses an unused-parameter warning for a parameter we intend to keep`.

This will need to be checked with gcc and cl on Windows, clang on the mac, gcc on Linux, and with the `compiler-warnings.yml` job.

WilliamK112:
Resolves #299

## Type of change

### Updated

* `devEnvSetupAndDefaults/.vscode/tasks.json` removes every GCC/Clang and MSVC unused-parameter warning suppression from the development build tasks.
* `m4/ax_compiler_flags_cflags.m4` enables `-Wunused-parameter` in the Autotools warning set used by `compile-warnings.yml`; `Makefile.am` already applies that generated `WARNING_CFLAGS` value to all targets.
* `.github/workflows/dev-platforms.yml` removes `/wd4100` from the MSVC build so the Windows CI job verifies the same policy.
* The affected graph library, extension, GraphML, and application callback implementations explicitly consume intentionally retained parameters with `(void)param;`. Every such line, including the three pre-existing GraphML markers, now has the requested explanatory comment.

The warning-flag investigation found that `compile-warnings.yml` selects the compiler and passes `-Werror`, `configure.ac` enables the checked warning set through `AX_COMPILER_FLAGS_CFLAGS`, and `Makefile.am` applies the resulting `WARNING_CFLAGS`. The VS Code tasks also contain compiler-specific warnings beyond that portable, feature-probed set; aligning all of those is separable from this unused-parameter fix.

## Testing

- Apple Clang 17, default 1-based storage: `autoreconf -if && ./configure --enable-compile-warnings CC=clang CFLAGS=-Werror && make -j4 && ./test-samples.sh`
- Apple Clang 17, 0-based storage: `./configure --enable-compile-warnings CC=clang CFLAGS="-Werror -DUSE_0BASEDARRAYS" && make -j4 && ./test-samples.sh`
- Independent syntax scan of every C translation unit with `-Wall -Wextra -Wunused-parameter`; no unused-parameter diagnostics remain.
- MSVC `/Wall` coverage is provided by the updated `dev-platforms.yml` job because MSVC is not available in the local macOS environment.

<details><summary>Validation result</summary>

```text
WARNING_CFLAGS includes -Wunused-parameter and no unused-parameter suppressor.
Default build: All tests have succeeded.
0-based build: All tests have succeeded.
```

</details>

### DIFF
--- a/.github/workflows/dev-platforms.yml
+++ b/.github/workflows/dev-platforms.yml
@@ -144,7 +144,7 @@ jobs:
       - name: Build using MSVC cl
         run: |
           cl /O2 /nologo /Wall /wd4464 /wd4710 /wd4711 /wd4774 /wd4996 /wd5045 `
-          /wd4068 /wd4100 /wd4820 /external:anglebrackets /external:W0 `
+          /wd4068 /wd4820 /external:anglebrackets /external:W0 `
           /Fe./planarity.exe ./c/planarityApp/**.c ./c/graphLib/**.c `
           ./c/graphLib/extensionSystem/**.c ./c/graphLib/homeomorphSearch/**.c `
           ./c/graphLib/io/**.c ./c/graphLib/lowLevelUtils/**.c `
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -705,6 +705,9 @@ int _ClearAllVisitedFlagsOnPath(graphP theGraph, int u, int v, int w, int x)
 {
     int e, eTwin;
 
+    // Suppresses an unused-parameter warning for a parameter we intend to keep
+    (void)w;
+
     // We want to exit u from e, but we get eTwin first here in order to avoid
     // work, in case the degree of u is greater than 2.
     eTwin = _gp_FindEdge(theGraph, v, u);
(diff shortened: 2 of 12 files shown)


john-boyer-phd (approved):
Great to have this resolved with the new method you showed us previously, so that the unused parameter decision has to be made for any new code where it appears.

PR #308, opened 2026-08-21
Complete GraphML Writer Part 2

### LINKED ISSUE #300: GraphML Writer, Part 2
Now that GraphML Writer, Part 1, has been completed, there are some behavioral changes that it has become evident are needed:

 - [ ] Remove unnecessary `index` parameters.
 - [ ] The output for undirected edges should prefer to have the source as the lower numbered vertex.
 - [ ] The output of IDs for nodes and edges should be changed to be always canonical.
 - [ ] The `graph` element start tag should emit helpful parse information.
 - [ ] As a consequence of canonicalizing IDs for vertices and edges, the zero-based I/O flag must be serialized as an attribute of the `graph` element.
 - [ ] Amend the GraphML test routine and samples to reflect the changes above. 

## Remove Unnecessary `index` Parameters

The methods `_WriteGraphMLGraphVertices()`, `_WriteGraphMLGraphEdges()`, and `_WriteGraphMLGraphEndTag()` don't need the `index` parameter, so those should be removed.

## Change for Undirected Edges

For the first item above, the prior spec indicated that "If the edge is undirected, then the source and target edge records can be set to be the loop iteration variable value and its twin (obtained using gp_GetTwin())." The change we would like is that the indexes assigned to the source and target edge record variables should then be swapped if necessary so that the variable for the index of the source edge record indicates the edge record in the adjacency list of the lower numbered vertex (and consequently has the higher-numbered `neighbor` value). In this was, the undirected edge output in the file `Digraph.transposeTest.graphml` 

```
    <edge id="e1" source="n1" target="n5"/>
```

Note that after the change to canonical numbering (described below), the line will change to:

```
    <edge id="e0" source="n0" target="n4"/>
```

## Canonical ID Generation for Vertices and Edges
  
For the second item about generating canonical IDs for nodes and edges, we have a few things to change up:

 - [ ] For the `id` attribute for each node (vertex), always emit a number that is the vertex index minus the lower bound of vertex storage, so that the IDs start with `n0`. There should be no code that changes behavior based on the zero-based I/O flag (that is addressed below, after this cluster of items). 
 - [ ] Similarly, the node IDs emitted as the `source` and `target` attribute values of `edge` elements should match the IDs produced for nodes above, i.e., always using vertex index minus lower bound of vertex storage, with no special code for zero-based I/O flag.
 - [ ] The edge ID counter should be started at 0 rather than one, so that the edge IDs start at `e0`
 
## More Informative Graph Element Start Tag

Based on canonicalizing vertex and edge IDs, the `parse.nodeids` and `parse.edgeids` attributes can be specified, to help optimize those readers that choose to use it. While in the neighborhood, there are other helpful parse attributes that can be easily emitted. Amend the `graph` element start tag output to be the following:
 
```
  <graph id="G%d" edgedefault="undirected"
         parse.nodes="%d" parse.edges="%d" 
         parse.nodeids="canonical" parse.edgeids="canonical" 
         parse.order="nodesfirst">
```

For the `parse.nodes` and `parse.edges` integer values, emit the results of `gp_GetN()` and `gp_GetM()`, respectively.

## Handling the Zero-Based I/0 Flag

Although we eliminated the zero-based I/O flag processing at the vertex and edge levels, we can still preserve the zero-based I/O flag so that transformations into other formats can be serialized into 0-based formats, if they support it. To do this, we can use the GraphML feature of defining custom attributes. The following steps are needed:   

 - [ ] Add a new method `int _WriteGraphMLKeysForGraph(graphP theGraph, strOrFileP outputContainer)` that returns `OK` on success and `NOTOK` on any failure. 
 - [ ] Invoke it right after the call to `_WriteGraphMLStartTag()` (and just before the call to `_WriteGraphMLGraphElement()`)
 - [ ] Let it emit nothing currently, unless the graph's zero-based I/O flag is set. If and only if the zero-based I/O flag is set, then emit the following (including the space indenting and the trailing linefeed):
 
```
  <key id="graphflags_zerobasedio" for="graph" attr.name="graphflags_zerobasedio" attr.type="boolean">
    <default>false</default>
  </key>
``` 
  
Next, we must amend the output of the `graph` element to include output of graph-level custom graph "attributes" (via XML `data` elements), as follows:

 - [ ] Add a new method `int _WriteGraphMLGraphCustomAttributes(graphP theGraph, strOrFileP outputContainer);` that returns `OK` on success and `NOTOK` on any failure.
 - [ ] Invoke it right after the call to `_WriteGraphMLGraphStartTag()` (and just before the call to `_WriteGraphMLGraphVertices()`)
 - [ ] Let it emit nothing currently, unless the graph's zero-based I/O flag is set. If and only if the zero-based I/O flag is set, then emit the following (including the 4-space indent and the trailing linefeed):
 
```
    <data key="graphflags_zerobasedio">true</data>
```

## Changes to the GraphML Tests

In `planarityCommandLine.c`, the method `runBasicGraphMLWriteTest()` needs to be amended to reflect the changes above. The files in the `samples` directory must also be edited, as follows:

 - [ ] The line containing `#ifdef USE_1BASEDARRAYS` as well as the whole `#else` block can be removed, because there will no longer be a difference between how the basic GraphML writer operates under 1-based versus 0-based compilations.
 - [ ] The expected `Digraph.transposeTest.graphml` file has to be amended to use canonical IDs for node and edge elements, to reverse the source and target of the undirected edge, to use the canonical node IDs in the source and target attribute values of the edge elements, and to have the improved `graph` element start tag. 
 - [ ] Add a second test that reads the input file `Digraph.transposeTest.0-based.txt` and outputs a new `Digraph.transposeTest.0-based.graphml` that is expected to have the exact same content as `Digraph.transposeTest.graphml` except for the addition of the `key` and `data` elements that express the `zerobasedio` flag setting.
 
If `planarity -test` works once these changes have been made (such as by uncommenting the `-test` and samples directory command-line parameters in `launch.json`), then you're ready to submit a PR.

WilliamK112:
Resolves #300

## Type of change

### Added

- `c/samples/Digraph.transposeTest.0-based.graphml`: expected GraphML output that records the zero-based I/O graph flag.

### Updated

- `c/graphLib/io/graphML-writer.c`: removes unused index parameters, emits canonical node/edge IDs, orients undirected edges from the lower-numbered vertex, adds GraphML parse metadata, and serializes the zero-based I/O flag as graph key/data elements.
- `c/planarityApp/planarityCommandLine.c`: exercises both one-based and zero-based GraphML inputs with shared test logic.
- `c/samples/Digraph.transposeTest.graphml`: updates the canonical expected output.

### Removed

- The compile-time branch that tested only one GraphML input/storage mode.

## Testing

- Default build: `make check` — 1/1 test passed.
- Clean zero-based build: `./configure CPPFLAGS=-DUSE_0BASEDARRAYS && make check` — 1/1 test passed.
- `xmllint --noout c/samples/Digraph.transposeTest.graphml c/samples/Digraph.transposeTest.0-based.graphml`
- `git diff --check`

### DIFF
--- a/c/graphLib/io/graphML-writer.c
+++ b/c/graphLib/io/graphML-writer.c
@@ -10,12 +10,14 @@ See the LICENSE.TXT file for licensing information.
 /* Private functions (exported to system) */
 int _WriteGraphMLGraph(graphP theGraph, strOrFileP outputContainer);
 int _WriteGraphMLStartTag(strOrFileP outputContainer);
+int _WriteGraphMLKeysForGraph(graphP theGraph, strOrFileP outputContainer);
 int _WriteGraphMLGraphElement(graphP theGraph, int index, strOrFileP outputContainer);
 int _WriteGraphMLEndTag(strOrFileP outputContainer);
 int _WriteGraphMLGraphStartTag(graphP theGraph, int index, strOrFileP outputContainer);
-int _WriteGraphMLGraphVertices(graphP theGraph, int index, strOrFileP outputContainer);
-int _WriteGraphMLGraphEdges(graphP theGraph, int index, strOrFileP outputContainer);
-int _WriteGraphMLGraphEndTag(graphP theGraph, int index, strOrFileP outputContainer);
+int _WriteGraphMLGraphCustomAttributes(graphP theGraph, strOrFileP outputContainer);
+int _WriteGraphMLGraphVertices(graphP theGraph, strOrFileP outputContainer);
+int _WriteGraphMLGraphEdges(graphP theGraph, strOrFileP outputContainer);
+int _WriteGraphMLGraphEndTag(graphP theGraph, strOrFileP outputContainer);
 
 /********************************************************************
  _WriteGraphMLGraph()
@@ -30,8 +32,32 @@ int _WriteGraphMLGraph(graphP theGraph, strOrFileP outputContainer)
         return NOTOK;
 
     if (_WriteGraphMLStartTag(outputContainer) != OK ||
+        _WriteGraphMLKeysForGraph(theGraph, outputContainer) != OK ||
         _WriteGraphMLGraphElement(theGraph, 1, outputContainer) != OK ||
         _WriteGraphMLEndTag(outputContainer) != OK)
+        return NOTOK;
+
+    return OK;
+}
+
+/********************************************************************
+ _WriteGraphMLKeysForGraph()
+
+ Writes GraphML key declarations for graph-level custom attributes.
+ ********************************************************************/
+int _WriteGraphMLKeysForGraph(graphP theGraph, strOrFileP outputContainer)
+{
+    char const *zeroBasedIOKey =
+        "  <key id=\"graphflags_zerobasedio\" for=\"graph\" "
+        "attr.name=\"graphflags_zerobasedio\" attr.type=\"boolean\">\n"
+        "    <default>false</default>\n"
+        "  </key>\n";
+
+    if (theGraph == NULL || !sf_IsValidStrOrFile(outputContainer))
+        return NOTOK;
+
+    if ((gp_GetGraphFlags(theGraph) & GRAPHFLAGS_ZEROBASEDIO) &&
+        sf_fputs(zeroBasedIOKey, outputContainer) == EOF)
         return NOTOK;
 
     return OK;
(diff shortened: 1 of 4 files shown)


john-boyer-phd (changes requested):
Hi William,
I had hoped you would take this one and also see that the result is an improved writer.
Could you take a moment please and submit another commit that resolves the merge conflicts? Pretty sure these are just from now removing the unused `index` parameters that were commented in the other PR.

WilliamK112:
Resolved the merge conflicts in 8d0460d by merging current master.

Your diagnosis was exactly right: the only content conflicts were the now-removed index parameters versus the unused-parameter suppression added by #307. I kept the Part 2 signatures without those parameters, retained the canonical vertexOffset behavior, and absorbed the rest of current master, including the new Add/Insert Edge regression coverage.

Revalidated on the merged head:
- Default strict-warning build: make check — 1/1 passed
- Zero-based strict-warning build: ./configure --enable-compile-warnings CC=clang CFLAGS=-Werror CPPFLAGS=-DUSE_0BASEDARRAYS && make check — 1/1 passed
- xmllint --noout on both expected GraphML samples
- git diff --check

The PR is now conflict-free relative to current master.

john-boyer-phd (approved):
All puppies and roses, thank you William!

PR #309, opened 2026-08-22
Add directed depth-first search mode

### LINKED ISSUE #291: Add Directed Depth-First Search Method
The current method `gp_DepthFirstSearch()` performs an undirected depth-first search (DFS) that assigns pre-order depth-first indexes to all vertices and assigns edge types to all edges. This is done even if the input graph is directed because the original and ongoing purpose of the method is to prepare a graph for undirected graph operations, such as planar embedding, in which the direction of the edge does not matter.

There are other algorithms specific to directed graphs that require the differentiated behaviors and information of a directed depth-first search. Furthermore, the downstream algorithms require the differentiated information created by a directed DFS even if the input graph is undirected. Since we also strongly prefer to not want to break backwards compatibility, we should add the new capability without changing the function signature of `gp_DepthFirstSearch()`. Therefore, we need an additional depth-first search method. Finally, it is quite conceivable that we may have to create other DFS variants in the future, so the new method should preferably not be specific to the new method we currently wish to add. 

We begin with some changes to the header file `graphDFSUtils.h` to declare the new function, including a DFS method selection parameter, and some constants pertaining to the method selection and completion of  and having it call the current method if the method selection parameter indicates the original method:

 - [ ] Add a new method declaration `int gp_DepthFirstSearchEx(graphP theGraph, unsigned mode)`, just below the `gp_DepthFirstSearch()` declaration.
 - [ ] Add a declaration `#define GRAPHFLAGS_DFSNUMBERED_DIRECTED 4096` just below the other `GRAPHFLAGS` entries, and add to the comment above to state that this new flag "is set when a directed depth-first search performs the DFS numbering"
 - [ ] Just below `gp_DepthFirstSearchEx()`, add a linefeed then declare `#define DFSMODE_UNDIRECTED 1` followed by `#define DFSMODE_DIRECTED 2` (then another linefeed).
 
The implementation for `gp_DepthFirstSearchEx()` should be placed in `graphDFSUtils.c` below `gp_DepthFirstSearch()` and will be just a router for calling routines that perform the two modes of depth-first search, after some error checking. First check if the graph is `NULL`, and return `NOTOK` if so. Then, if the mode equals `DFSMODE_UNDIRECTED`, then just return the result of calling `gp_DepthFirstSearch()`. Otherwise, return `NOTOK`.

You can actually spot test the work so far by temporarily going into `planarityRandomGraphs.c` and changing the invocation of `gp_DepthFirstSearch()` to `gp_DepthFirstSearchEx(theGraph, DFSMODE_UNDIRECTED)`. Then, run the planarity executable, click `p` for planarity, `100000` for number of graphs, and `20` for the number of vertices. When that succeeds, you will know that you are doing well so far, and then you can change the invocation back to just `gp_DepthFirstSearch()`.

Next, we need to add the method for directed depth-first search into `graphDFSUtils.c` as follows:

 - [ ] Just above the declaration `int _SortVertices(graphP theGraph)`, add the declaration `int _DepthFirstSearchDirected(graphP theGraph)`
 - [ ] Just below `gp_DepthFirstSearchEx()`, add the method `_DepthFirstSearchDirected()` and implement it by having it just return `NOTOK` for the moment.
 - [ ] In `gp_DepthFirstSearchEx()`, add an `else` clause that invokes `_DepthFirstSearchDirected()` if the `mode` parameter equals `DFSMODE_DIRECTED`
 
Before we turn to filling in the implementation of `_DepthFirstSearchDirected()`, we need to amend `gp_DepthFirstSearch()` now that it is clear that we expected `gp_DepthFirstSearch()` to perform its current operations on even on directed graphs. Specifically, 

 - [ ] At the beginning, of `gp_DepthFirstSearch()`, remove the `if` test that returns `NOTOK` if the graph contains directed edges.
 - [ ] At the end, just after the assignment `theGraph->graphFlags |= GRAPHFLAGS_DFSNUMBERED`, do an `if` test to see if `GRAPHFLAGS_DFSNUMBERED_DIRECTED` is set and, if it is, then do two things: 1) use bitwise-and with `~GRAPHFLAGS_DFSNUMBERED_DIRECTED` to turn off that flag because, and 2) invoke `_FillVertexVisitedIndexes(theGraph, 0)`. 
 - [ ] There is a custom, optimized version of undirected depth-first search in `_EmbeddingInitialize_Optimized()` that also sets `theGraph->graphFlags |= GRAPHFLAGS_DFSNUMBERED`, so that location also needs the test of the `GRAPHFLAGS_DFSNUMBERED_DIRECTED` bit and the two operations in the preceding step if the bit is set.
 
 Now we turn to implementing `_DepthFirstSearchDirected()` according to the following:
 
  - [ ] The discovery times are placed in the vertex index fields, and the finish times are placed in the vertex `visitedIndex` fields
  - [ ] Initialize the `index` and `visitedIndex` fields to 0 (not yet discovered and not finished)
  - [ ] Use an in-use edge iteration loop (look for one containing `gp_EdgeInUse()`) to clear the edge types and marked bits of all in-use edges.
  - [ ] Perform a for loop/while loop similar to that in `gp_DepthFirstSearch()` with some exceptions described in a list below. 
  - [ ] When the for loop/while loop is finished, set `GRAPHFLAGS_DFSNUMBERED_DIRECTED`, clear the `GRAPHFLAGS_DFSNUMBERED` bit, and return OK.
  
Key differences for a directed DFS are:

 - [ ] Rather than a variable `DFI`, use a `timer` initialized to 1 and incremented *after* every time it is assigned to a discovery time or finish time
 - [ ] Use of the visited bit can be replaced by whether the discovery time (vertex index) has been set (the vertex is unvisited or undiscovered if the discovery time is still zero). 
 - [ ] Only push edge records that are out-only or undirected. Push from the back of the list to the front, so they are pushed right to left order relative to what you see in the adjacency list file. This is because we are using an explicit stack instead of recursion.
 - [ ] The first pushed edge record should have its marked bit set so that when you pop it, you will know you're on the last edge record for the vertex and can set its finish time (and then increment the timer). If no edge records are pushed for a vertex, then you can set its finish time (and then increment the timer).
 - [ ] When an edge record `e = (v -> w)` is popped that has the marked bit set, the process it as below, and then clear the marked bit of `e` and set the finish time of the source vertex `v` (because you just finished exploring everything that can be reached from `v`).
 - [ ] When an edge record `e = (v -> w)` is popped, this is logically equivalent to having a recursive depth-first search that looping left-to-right through the adjacency list of `v` and has just performed a recursive call to visit vertex `w`. At this moment, the edge type for `e` can be determined based on discovery and finish time stamp analysis, according to the next two bullet points.
 - [ ] If the `e = (v -> w)` is popped and `w` does not yet have a discovery time (still 0), then label the edge record `e` with `EDGE_TYPE_TREE`, set its discovery time of `w` (its index) using the timer (and increment the timer), set the parent of `w` to `v` (the vertex that owns the edge record), and then push all the out-only and undirected edges of `w`.
 - [ ] If the `e = (v -> w)` is popped and `w` does have a discovery time (!= 0), then we will only be setting the type of edge record `e`. If `w` is not finished (its finish time in its `visitedIndex` is zero), then set the type of `e` to `EDGE_TYPE_BACK`. Otherwise, `w` is finished, so set the type of `e` to `EDGE_TYPE_CROSS` if the discovery time of `w` is less than the discovery time of `v`, and to `EDGE_TYPE_FORWARD` otherwise.

 Last but not least, we can test that this work is ready for a PR by adding a regression test into `runDigraphTests()` as follows:
 
  - [ ] Add a file `DirectedDFSTest.txt` with content below into the samples directory, and add it to the `makefile.am`
  - [ ] Add a function call to a new function `testDirectedDFS()` to the execution sequence of `runDigraphTests()`
  - [ ] Make `testDirectedDFS()` call `gp_DepthFirstSearchEx()` with `DFSMODE_DIRECTED` (and test for `OK`/`NOTOK`)
  - [ ] Programmatically check that `GRAPHFLAGS_DFSNUMBERED_DIRECTED` has been set (returning NOTOK if the check fails)
  - [ ] Programmatically check that the discovery times are 1, 11, 2, 3, 7, 4
  - [ ] Programmatically check that the finish times are times are 10, 12, 9, 6, 8, 5
  - [ ] Programmatically check that all out-only and undirected edge records are of type `EDGE_TYPE_TREE` except that 5->1 and 5->3 are `EDGE_TYPE_BACK`, 2->3 and 2->4 are `EDGE_TYPE_CROSS`, and 1->4 is `EDGE_TYPE_FORWARD`
  - [ ] If this work is being done after loop edge support is added, then there will exist a method called `runLoopEdgeTests()`.  Add to that test a call to `gp_DepthFirstSearchEx()` on the graph G<sub>1</sub> being manipulated in that test, and ensure that it returns OK and that all loop edges are marked as back edges.
  - [ ] Change the `launch.json` setting for the debug mode of your compiler to uncomment the command line parameters that enable you to run planarity with `-test` on the samples directory.
  - [ ] Run in debug mode and see if you get a successful test or if it is necessary to debug. 
  
The content of the `DirectedDFSTest.txt` file should be:

```
N=6
1: 3 4 0
2: 3 4 0
3: 4 5 0
4: 6 0
5: 1 3 0
6: 0
```

**NOTE:** To get the discovery and finish times listed above, you'll have to push the edges from the ends of the lists to the beginning. For example, notice that the discovery time of vertex 3 is 2. This is because the edges from vertex 1 must be pushed in the order 4 and then 3 so that the edge to 3 is popped first and exploration proceeds to vertex 3 before going to vertex 4. We are using an explicit stack rather than using recursion, and a directly recursive implementation would recurse to exploring vertex 3 and then go onto vertex 4 after the recursion to vertex 3 returned.

WilliamK112:
Closes #291

## Summary

- add `gp_DepthFirstSearchEx()` with undirected and directed modes plus the directed-DFS completion flag
- implement iterative directed DFS discovery/finish timestamps and tree, back, cross, and forward edge classification
- preserve the legacy `gp_DepthFirstSearch()` behavior on directed input and cleanly replace directed DFS state when legacy DFS or embedding initialization runs
- add directed DFS regression coverage, including exact timestamps, edge types, mode switching, and embedding-state replacement

## Implementation notes

The final outgoing edge for a vertex is marked as requested. When that edge discovers a new child, a compact deferred finish event keeps the source active until the child subtree completes; this produces recursive DFS finish-time semantics without recursion and stays within the existing graph stack capacity.

The current codebase names the issue-described full embedding path `_EmbeddingInitialize_Optimized()`, so the directed-state cleanup is applied there.

The issue sample snippet needs the normal adjacency-list header and final colon to be readable by `gp_Read()`; `DirectedDFSTest.txt` therefore adds `N=6` and uses `6: 0` while preserving the specified graph.

## Validation

- `make check` with 1-based arrays and compile warnings enabled
- `make check` with `-DUSE_0BASEDARRAYS`
- DEBUG build: `planarity -test c/samples`
- ASan/UBSan build: `make check`
- `make distcheck`
- `git diff --check`


WilliamK112:
Fixed the GCC 14 signedness failure in eda8fb7.

The test stored the expected edge type in a signed int, while gp_GetEdgeType() returns the masked unsigned flags field. I changed only expectedType to unsigned so the comparison uses matching types; directed DFS behavior is unchanged.

Local validation:
- ./configure --enable-compile-warnings CFLAGS=-Werror
- make -j2 check (1/1 passed)
- git diff --check

john-boyer-phd (approved):
This looks great, William. A couple of tiny tweaks. I especially liked the deferred finish event and how you immunized the test from 1-based versus 0-based array indexing compiles.

PR #311, opened 2026-08-26
Test single random graph endpoints

WilliamK112:
## Summary

- exercise the single-graph generator used by the `-rm` and `-rn` command-line endpoints from the quick regression suite
- use five vertices so the maximal-planar case is expected to return `OK` and the one-extra-edge case is deterministically `NONEMBEDDABLE`
- temporarily suppress interactive save prompts and restore the caller's quiet-mode setting afterward
- avoid creating output files in the source samples directory during out-of-tree or `distcheck` runs

This fills the endpoint gap identified while reviewing whether #178 is ready to close. It complements the existing `RandomGraphs()` checks rather than restoring the older shell-level temporary-file tests.

## Validation

- Clang build with `-Wall -Wextra -Werror -g -O1` and full `./test-samples.sh`
- 20 additional consecutive full regression runs
- AddressSanitizer build and full regression
- UndefinedBehaviorSanitizer build and full regression
- `USE_0BASEDARRAYS` warning-as-error build and full regression
- `git diff --check`

Refs #178

### DIFF
--- a/c/planarityApp/planarityCommandLine.c
+++ b/c/planarityApp/planarityCommandLine.c
@@ -345,6 +345,7 @@ int runAddInsertEdgeTests(void)
 int runRandomGraphsTests(void)
 {
     int retVal = OK;
+    unsigned quietModeCache = gp_GetQuietMode();
 
     gp_Message("Starting Random Graph Tests");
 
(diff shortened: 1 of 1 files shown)


wbkboyer (approved):
Nice work temporarily suppressing the prompts (`planarityRandomGraphs.c` ln. 677-694) for the purposes of automated testing!

On MacOS with `clang` 22.1.6, I used VSCode to produce the Debug, Release, and AddressSanitizer builds with 1-based arrays, and ran `-tests` for Debug and AddressSanitizer, the latter with the following command:

```
ASAN_OPTIONS=detect_leaks=1:symbolize=1 ASAN_SYMBOLIZER_PATH=/opt/homebrew/opt/llvm/bin/llvm-symbolizer ./AddressSanitizer/planarity -test c/samples
```

I then re-compiled the AddressSanitizer build using `-DUSE_0BASEDARRAYS` and re-ran the tests, and no leaks were found.

Finally, I uncommented the quiet-mode suppression to see what output files were being made and if they made sense, and it seems like the expected 5-vertex random max planar and random nonplanar graphs are being made.

PR #312, opened 2026-08-28
Name remaining function declaration parameters

### LINKED ISSUE #188: Add parameter names to all function declarations that don't have them

WilliamK112:
## Summary

Closes #188.

Add names to all 25 graph function-table callback signatures and to the 14 remaining unnamed parameters in C forward declarations. The follow-up covers the six source files identified in review; a repository-wide audit found and corrected the same pattern in six additional source files.

This remains a names-only change: parameter types, order, return types, and behavior are unchanged.

## Validation

- Checked each callback signature and forward declaration against its corresponding implementation.
- Audited tracked C headers and source forward declarations for remaining unnamed parameters.
- `make -j2 check` — passed; all sample tests succeeded.
- `./configure --enable-compile-warnings CC=clang CFLAGS=-Werror && make -j2` — passed with the full project warning set.
- `make -j2 check` under the strict warning configuration — passed.
- Header syntax checks with C11 and C++17, using `-Wall -Wextra -Werror -pedantic` — passed.
- `git diff --check` — passed.

### DIFF
--- a/c/graphLib/extensionSystem/graphFunctionTable.h
+++ b/c/graphLib/extensionSystem/graphFunctionTable.h
@@ -24,37 +24,37 @@ extern "C"
         // These function pointers allow extension modules to overload some of
         // the behaviors of protected functions.  Only advanced applications
         // will overload these functions
-        int (*fpEmbeddingInitialize)(graphP);
-        void (*fpEmbedBackEdgeToDescendant)(graphP, int, int, int, int);
-        void (*fpWalkUp)(graphP, int, int);
-        int (*fpWalkDown)(graphP, int, int);
-        int (*fpMergeBicomps)(graphP, int, int, int, int);
-        void (*fpMergeVertex)(graphP, int, int, int);
-        int (*fpHandleInactiveVertex)(graphP, int, int *, int *);
-        int (*fpHandleBlockedBicomp)(graphP, int, int, int);
-        int (*fpEmbedPostprocess)(graphP, int, int);
-        int (*fpMarkDFSPath)(graphP, int, int);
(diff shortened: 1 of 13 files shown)


john-boyer-phd (changes requested):
It's great to have the param names in there, William, thank you.

To fully close #188, there are just a few C files I noted that have a function without a name for the parameter: The _extension.c files for K23Search, K33Search, K4Search and Drawplanar; graphDFSUtils.c; and the g6 api utils C file.

Could you add a commit to add those parameter names so we can close #188? May want to tweak the PR name to match, but not required.

WilliamK112:
Thanks, John. I added the requested parameter names in the K23Search, K33Search, K4Search, and DrawPlanar extension files, `graphDFSUtils.c`, and the graph6 API utilities file on `efcb6e0`.

I also audited the remaining C forward declarations and found the same unnamed-parameter pattern in six additional source files, so those are included as names-only fixes as well. No parameter types, ordering, return types, or behavior changed.

The full Clang compiler-warning build passes with `-Werror`, and `make check` passes under that configuration. Ready for another look when convenient.

john-boyer-phd (approved):
Thanks for adding those other params William, looks great.

PR #313, opened 2026-08-28
Propagate strOrFile read errors instead of accepting truncated input

### LINKED ISSUE #187: Add more robust error detection and handling to `strOrFile` container API
## Problem statement

Originally, in `c/graphLib/io/graphIO.c`'s `_ReadGraph()`, when `extraDataAllowed` for a graph input file, we loop a call to `sf_fgets()` to get all additional input lines and append them to the `extraData` `strBuf`. For an input file, this means we are calling [`fgets()`](https://en.cppreference.com/w/c/io/fgets) under the covers, which returns `NULL` if the file contents are exhausted *or* if a read error has occurred. In the latter case, the `errno` is set (see [`fgets()` ISO standard](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fgets.html)); however, we currently don't check this flag, and so we are unable to differentiate between these two cases.

## Task description

* [ ] Add `errorFlag` to `strOrFile` struct
* [ ] Update `sf_fgets()` so that for a `strOrFile` container that holds a `FILE *`, check for error flags when using `fgets()` and indicate an error state to the caller.
* [ ] Update consumers of `strOrFile` containers so that they detect and handle error flags being set

WilliamK112:
## Summary

Closes #187.

Distinguish a file read failure from normal EOF throughout the `strOrFile` input path.

- Add a sticky `inputErrorFlag`, named alongside the existing `outputErrorFlag`. Both `sf_getc()` and `sf_fgets()` check `ferror()` and refuse further reads after an error, including buffered pushback. Normal EOF and string-backed input retain their existing behavior.
- Propagate failures through whitespace/integer readers and graph extra-data processing. Do not invoke `fpReadPostprocess` with partially read data or overwrite an earlier failure with its return value.
- Make the graph6 iterator return `NOTOK` for a read error instead of marking normal end-of-input; preserve that result in the single-graph wrapper. Reject incomplete multibyte order bytes before shifting them.
- Put the fault-injection coverage in `c/planarityApp/planarity-test/testReadErrors.c` and call it from the existing `planarity -test` flow.
- Keep the existing test, warning, and sanitizer workflows unchanged. The only Makefile change registers the new source with the `planarity` executable.

## Review follow-up

- Merged current `master` through `27bf82c`.
- Removed the separate `test-read-errors` executable and Automake test target.
- Restored `compile-warnings.yml`, `runtests.yml`, `sanitisers.yml`, and `.gitignore` to current-master behavior.
- Moved the test implementation into the requested `planarityApp/planarity-test` subdirectory.
- Added `runReadErrorTests()` to `runQuickRegressionTests()`, so the existing `test-samples.sh` and sanitizer workflows execute it automatically.

## Validation

- The integrated read-error suite passes **94 checks** from `planarity -test`, including normal EOF, pushback, partial lines, sticky error state, integer/whitespace readers, adjacency-list/matrix extra data, graph6 iteration/wrapping, and incomplete graph6 order bytes.
- `autoreconf -fi && ./configure && make -j2 check` — passed.
- `./configure --enable-compile-warnings=error CC=clang && make -j2 && make -j2 check` — passed.
- AddressSanitizer build plus `test-samples.sh` — passed.
- UndefinedBehaviorSanitizer build plus `test-samples.sh` with `halt_on_error=1` — passed.
- No-callback C11 `-Wall -Wextra -Werror -pedantic` compile of `testReadErrors.c` — passed.
- `make -j2 distcheck` — passed, including out-of-tree distribution build/test/install/uninstall.
- `git diff --check` — passed.

Local execution is on macOS using `funopen`. The `fopencookie` path is wired for Linux CI but has not been executed locally. Platforms without either callback API skip only the fault-injection portion while the rest of `planarity -test` continues.


john-boyer-phd (changes requested):
Thank you for taking this on, William. 

You wired in the tests in a really interesting way. However, I think it would be better to refactor this so that the test-read-errors code can be run from `planarity -test`. There are a couple of reasons. One is I'd prefer not to build other executables (and not sure if make will still work on Windows, given that libPlanarity is a dll, but didn't test that). Another is that any additional code testing that is being done can be routed through the sanitizers. A third is that all the test stuff can end up in one place under one methodology.

Along those lines, there is a habit I've got us into that we will also have to get out of around just putting all kinds of tests into `planarityCommandLine.c`. It would be nice if this refactor could start us down the road to recover by putting the `test-read-errors` code into a `planarity-test` subdirectory **within** the `planarityApp` directory. Later, we can migrate all the other tests from `planarityCommandLine.c` and into C files in `planarityApp/planarity-test`.

Once `test-read-errors` is moved to this location and then called from `planarityCommandLine.c`, then all the build and yml file modifications could be rolled back. (My apologies, but it looks like I merged master before realizing this, which will make that roll-back a little harder). Would this be OK, William?

WilliamK112:
Yes, absolutely — thanks for the direction, John. I refactored this in `f1862983`:

- moved the fault-injection tests to `c/planarityApp/planarity-test/testReadErrors.c`
- called `runReadErrorTests()` from the existing `runQuickRegressionTests()` path
- removed the separate test executable and Automake test target
- restored all three workflow files and `.gitignore` to current-master behavior
- merged current `master` through `27bf82c`

The 94 read-error checks now run as part of `planarity -test`, including under the existing sanitizer workflow path. Fresh local validation passed for normal and strict-warning builds, ASan, UBSan, no-callback compilation, and `make distcheck`.

This also creates the requested `planarity-test` location without adding more test implementation to `planarityCommandLine.c`; that file only declares and invokes the new test entry point. Ready for another look when convenient.

john-boyer-phd:
Hi @WilliamK112 , With apologies, we had some important bugs to resolve that caused a merge conflict with your PR. Please could you fix them in another commit?

c/planarityApp/planarity-test/testReadErrors.c
```diff
@@ -0,0 +1,294 @@
...
+static int postprocessCalls = 0;
+static const char *expectedExtraData = NULL;
+
+#define CHECK(condition)                                                        \
+    do                                                                         \
+    {                                                                          \
+        checks++;                                                              \
+        if (!(condition))                                                      \
+        {                                                                      \
+            fprintf(stderr, "Line %d: %s\n", __LINE__, #condition);              \
```
john-boyer-phd: Any places like this with fprintf to give errors should be expressed as `gp_ErrorMessage()`.

c/planarityApp/planarity-test/testReadErrors.c
```diff
@@ -0,0 +1,294 @@
...
+#ifdef HAVE_FOPENCOOKIE
+    cookie_io_functions_t functions = {readInput, NULL, NULL, NULL};
+    stream = fopencookie(input, "r", functions);
+#else
+    stream = funopen(input, readInput, NULL, NULL, NULL);
+#endif
+    if (container == NULL || stream == NULL)
+    {
+        fprintf(stderr, "Unable to create test input stream.\n");
+        exit(EXIT_FAILURE);
```
john-boyer-phd: Any of these `exit()` lines will need to be more in the `return NOTOK` style.

c/planarityApp/planarity-test/testReadErrors.c
```diff
@@ -0,0 +1,294 @@
...
+    }
+    sb_Free(&container->theStrBuf);
+    container->pFile = stream;
+    return container;
+}
+
+static graphP newGraph(void)
```
john-boyer-phd: May not need this anymore.

john-boyer-phd (changes requested):
A few changes based on moving the tests into being part of `planarity -test` (in addition to the ask to please resolve the latest merge conflicts from the other PRs).

WilliamK112:
Thanks, John — I addressed the requested changes and the new master conflicts in 39c583f (with 0fd8c01 preserving the existing remote PR history):

- merged master through a71c3b6 and combined its int-based sf_getc/sf_ungetc EOF handling with this PR’s sticky read-error propagation
- changed the test CHECK diagnostics from fprintf to gp_ErrorMessage
- replaced all exit calls with propagated OK/NOTOK returns and cleanup
- removed the redundant newGraph helper
- kept the PR diff limited to the same eight implementation/test/build files

Fresh validation passed:

- Clang warnings-as-errors build plus full planarity -test: 94/94 read-error checks and all quick regressions
- warnings-as-errors with --enable-unsigned-char: all tests passed
- forced no-fopencookie/no-funopen build: compiled and correctly skipped only the fault-injection tests
- AddressSanitizer and UndefinedBehaviorSanitizer: all tests passed
- make distcheck -j4: passed

The branch now merges cleanly with current master and is ready for another look.

john-boyer-phd (changes requested):
Hi William,

This is really coming along, getting close.

I'd like to ask if you could review the commits I've done on this PR, for your info on future PRs (e.g., make sure your dev env is using the gitignore, add linefeeds on tests). Also, please double-check in case I missed any of the EOF checks that needed to be added.

More importantly, please could you look into the following:

 * Clean on all jobs now, except one of the sanitizers is annoyed about something. Are you able to diagnose and fix? [**edit**: it is almost certainly fixable by doing the next bullet point below]
 * Please could you use the technique now shown in `runCapacityLimitTests()` to suppress error messages coming from the tests, when the errors are the intended outcomes of the tests you are running?
 * Do you have any way to fill in the `#else` part at the bottom of `testReadErrors.c` that currently emits "Custom FILE read callbacks are unavailable..."? It would be nice to see a version of your tests running on Windows, for example. (**edit**: The GitHub actions do run these tests, so the answer to this can be "Not really, no", but it seemed best to ask because it's definitely no if you don't ask).
 * **edit**: For consistency, setting the `inputErrorFlag` should be done with a new method `sf_SetInputErrorFlag()` that also checks for `NULL`, as this will save a lot of typing in the subsequent points. 
 * **edit**: It seems that numerous of the `sf_` input functions should, for completeness, set the `inputErrorFlag` to `TRUE` if their initial parameter check fails or if a subordinate function fails, rather than just returning `EOF` or `NOTOK` or assigning `NOTOK` to an `exitCode`. In the case of calling a subordinate function, it may seem redundant, but it's best to not make assumptions about what non-local code does. AFAICT, the functions that need modification are: `sf_ungets()`, `sf_fgets()`, `sf_ungetc()`, `sf_ReadInteger()`, `sf_ReadSingleDigit()`, `sf_ReadSkipWhitespace()`, `sf_ReadSkipChar()`, and `sf_getc()`.
 * **edit**: In `sf_closeFile()`, the failure of of `fflush()` or `fclose()` should be followed by deciding whether to set `theInputErrorFlag` for an input container, or `theOutputErrorFlag` for an output container.
 * **edit**: There are a few places in `sf_ReadInteger()` where the result of `sf_getc()` is not checked for `EOF` or if it has set the `inputErrorFlag`. It can be legit to read an integer that ends at the end of file, so `EOF` should be allowed to end a number validly, but if the `EOF` is accompanied by a `TRUE` `inputErrorFlag`, then that should cause propagation of the error.  

Thank you,
John

john-boyer-phd:
Dear @WilliamK112 
One thing I realized I should mention is that I tend to commit to a PR mainly to remove relatively "piddly" stuff out of the way. Stuff that if I had instead done it as a review comment, it would be reasonable to wonder "why did he waste time saying that in words instead of code?!" If you'd rather this sort of stuff as review comments, just let me know, but my intention is to keep your awesome talent focused on the things that most require it.

Along those lines, it looks like I didn't realize until now that `sf_ReadInteger()` needs work to address the possibility of `EOF` returns from `sf_getc()`. A wrinkle is that it may be legit to get an EOF as the way to end reading of an integer (e.g., there may not be a trailing linefeed at the end of an adjacency list formatted graph file). This means that `EOF` has to be handled with care within `sf_ReadInteger()` because it may be a true `EOF` but it's necessary to check the input container's `inputErrorFlag` to be sure. If you're doing another commit for any of the other bullet points from the latest review, please could you add to it to address this?

WilliamK112:
Thanks, John — I reviewed the commits you added and addressed the latest review in `831edd2`.

- Added `sf_SetInputErrorFlag()` and routed the listed input helpers through it for invalid parameters and subordinate-operation failures.
- Reworked `sf_ReadInteger()` so ordinary EOF can validly terminate an integer, while EOF accompanied by the sticky input-error flag fails without changing the caller's output value. The tests now cover positive/negative values at both the ordinary loop EOF and the maximum-length look-ahead boundary.
- Made `sf_closeFile()` mark `inputErrorFlag` or `outputErrorFlag` according to the container direction, and added an injected input-close failure test.
- Applied the `runCapacityLimitTests()` quiet-mode pattern only around calls expected to fail, restoring the prior quiet mode before assertions. This fixes the sanitizer job's false failure: the Valgrind workflow greps for “Invalid”, and the expected “invalid graph order” diagnostics were triggering it despite Valgrind reporting zero errors.
- Added focused failure-state checks for `sf_ReadSkipChar`, `sf_ReadSingleDigit`, `sf_ungetc`, `sf_ungets`, and `sf_fgets`.

For the no-`fopencookie`/no-`funopen` branch: I do not know a portable stdio mechanism on Windows that can inject a `FILE *` read failure at a chosen byte boundary, so I left the explicit skip in place. I did build and run that branch to verify it remains clean.

Fresh validation passed:

- Clang warnings-as-errors build and full `planarity -test`
- warnings-as-errors with `--enable-unsigned-char`
- forced no-custom-callback build and full tests
- AddressSanitizer and UndefinedBehaviorSanitizer full tests
- `make distcheck -j4`
- 128/128 read-error regression checks

Ready for another look.

john-boyer-phd:
This is looking great, @WilliamK112 . Just so you know what's going on, we're going to pause a beat and roll out the 5.1 release, and once that tag is created (which will be pretty quickly), then I'll refresh this PR with in and then it can roll in as one of the first things in 5.2.

WilliamK112:
Thanks for the update, John — that sounds great. I will leave the branch in place for the 5.2 queue and take another look after the 5.1 tag if anything needs reconciling.

c/planarityApp/planarity-test/testReadErrors.c
john-boyer-phd: This has been resolved in the latest commit.

c/planarityApp/planarity-test/testReadErrors.c
john-boyer-phd: This has been resolved in the latest commit.

c/planarityApp/planarity-test/testReadErrors.c
john-boyer-phd: This has been resolved in the latest commit.

john-boyer-phd (changes requested):
I had to make sure to scrutiize `testReadErrors.c` as well. It's a spark-tactular piece of work, William. Really. I appreciated how much I learned reading this.

I'm pretty much ready to merge this in, but I'd like to ask you to please consider these little additions:

1. As additional tests, it seems possible to encode tests for failure on close graph reading by setting `errno` to 0 immediately before a `_ReadGraph()` and a `g6_FreeReader()` then looking for non-zero `errno` immediately after the call. These would be desirable as a way to illustrate for graph application developers that they could IRL do the same thing with `gp_Read()` and `g6_FreeReader()`.
2. Your PR contains the code that fixes #328 so it seems like an ideal opportunity to add a read error test that demonstrates that the failure now happens on the specific graph mentioned in the issue. It will be another case of having to go to all-quiet mode so you can get the `!= OK` result without emitting any error messages, because we _want_ the error to happen.

Would you be able to add these, William?

WilliamK112:
Thanks, John. I added both requested close-failure checks: each test clears errno immediately before _ReadGraph() or g6_FreeReader(), then verifies that the injected close failure leaves EIO. I also added a public gp_ReadFromString() regression for issue #328 using the truncated order-5 graph6 line D?. The expected rejection runs in all-quiet mode.

I merged current master through 8b5bb3b, preserving its sparse6 reader and test coverage; PR #313 is conflict-free again. Local validation: warnings-as-errors build, full planarity -test ./c/samples (137/137 read-error checks and sparse6 tests), and make check all pass. The update is at 32c8bb8.

WilliamK112:
CI note: the only failing check is Windows-MINGW32. It failed during msys2/setup-msys2, before the project build or tests: the ftp2.osuosl.org mirror timed out downloading two MinGW package signatures, and pacman aborted the transaction. The other hosted checks passed. I tried to rerun that job, but GitHub returned "job 104897196634 cannot be rerun" for my account. Could a maintainer rerun the failed job when convenient? No source change appears warranted for this infrastructure failure.

john-boyer-phd (approved):
Fantastic work, once again, William!

Issue #316, opened 2026-08-28
`compile-warnings.yml` cannot fail on warnings that need `-O2`, and is hiding two `chdir` warnings


MODSetter:
The compiler warning job passes, but that is an artefact of how `CFLAGS` is handed to `configure`
rather than evidence that the tree is clean.

`.github/workflows/compile-warnings.yml`:

```yaml
    env:
      CC: ${{ matrix.compiler }}
      CFLAGS: -Werror
    ...
          ./configure --enable-compile-warnings CC="$CC" CFLAGS="$CFLAGS"
```

Assigning `CFLAGS` on the `configure` line replaces autoconf's default `-g -O2` rather than adding to
it, so the job builds unoptimized. On Ubuntu that matters: gcc auto-defines `_FORTIFY_SOURCE` when
optimizing, fortification is what makes glibc attach `warn_unused_result` to `chdir`, and without
`-O2` none of that happens. So the class of warning the job exists to escalate never gets emitted.

Four configurations, same tree, Ubuntu 24.04 with gcc 13.3.0:

| `./configure` arguments | `make` exit | diagnostics |
|---|---|---|
| none, which is what a user runs | 0 | 2 warnings |
| `--enable-compile-warnings CC=gcc CFLAGS=-Werror`, the job verbatim | 0 | none |
| `--enable-compile-warnings CC=gcc CFLAGS="-O2 -Werror"` | 2, fails | 2 errors |
| `--enable-compile-warnings CC=gcc` | 0 | 2 warnings |

## The two warnings it is hiding

A plain `./configure && make` on Ubuntu today:

```
c/planarityApp/planarityCommandLine.c:218:17: warning: ignoring return value of 'chdir' declared with attribute 'warn_unused_result' [-Wunused-result]
c/planarityApp/planarityCommandLine.c:263:5: warning: ignoring return value of 'chdir' declared with attribute 'warn_unused_result' [-Wunused-result]
```

Both are the `chdir(origDir)` calls that restore the working directory. Line 218 is on the path where
the samples directory could not be entered, and line 263 runs after the test loop. If either fails the
process carries on in the wrong directory, which is a real, if unlikely, failure mode rather than
noise.

## Suggested fix

Two parts, either of which can go first:

1. Give the job the optimization level a user gets, `CFLAGS: "-O2 -Werror"`, or append `-Werror`
   through `AM_CFLAGS` so autoconf's `-g -O2` survives. Note this will fail the build until part 2
[...]

Issue #310, opened 2026-08-23
Improvements for Compiler Flags


john-boyer-phd:
The recent work on #304 has raised awareness of a few ways in which we could improve our compiler warnings, i.e., make the tests even more stringent.

Most of the items that were here have now been transferred to Issue #324 .

 - [x] Ignoring `-fno-strict-aliasing` and the double `-Wall`, the `tasks.json` lists have 45 warnings, but the `AX_APPEND_COMPILE_FLAGS` used in our make file and hence our `compiler-warnings.yml` only has 33 entries. Add the missing entries to `AX_APPEND_COMPILE_FLAGS` and determine whether it causes any difficulty for any of the GitHub actions (not just `compiler-warnings.yml`).

Deln0r:
I will take the second item in the warnings group: bringing `AX_APPEND_COMPILE_FLAGS` up to the `tasks.json` set (the 33 vs 45 gap) and checking what the added flags do to each of the GitHub workflows, per John's suggestion.

john-boyer-phd:
The item in this issue was resolved with PR #318

PR #318, opened 2026-08-31
Align AX_APPEND_COMPILE_FLAGS with the tasks.json warning set

Deln0r:
Part of #310 (second item in the warnings group).

## Type of change

- Build system change (non-breaking; no library or application code touched)

### Updated

* `m4/ax_compiler_flags_cflags.m4` adds the five warning flags that the VS Code tasks enable and the Autotools set omitted: `-Wunused-result`, `-Wignored-qualifiers`, `-Wdiscarded-qualifiers`, `-Wdeprecated-non-prototype` and `-Wincompatible-pointer-types-discards-qualifiers`. The gcc-only and clang-only spellings sit in the same list because `AX_APPEND_COMPILE_FLAGS` probes every flag and drops what the active compiler rejects; the probe result is in the testing section below.
* The same file records why the remaining `tasks.json` entries are deliberately not mirrored, per the first bullet of #310: `-Wno-deprecated-declarations` is a suppression rather than a warning and would weaken the gate `compile-warnings.yml` exists to enforce, and `-fsanitize=address`, `-fdiagnostics-color=always`, `-fcolor-diagnostics` and `-fansi-escape-codes` are not warning flags.

After this the two lists differ only by those documented entries. This is the follow-up #307 set aside as separable.

## Testing

Each configuration is a clean `make distclean && autoreconf -if && ./configure --enable-compile-warnings CC=<cc> CFLAGS=-Werror && make -j4`.

- Apple clang 21, macOS arm64: build clean, `make check` 1/1 pass.
- gcc 14 and clang 18 on Ubuntu 24.04 x86_64, matching the `compile-warnings.yml` matrix: both build clean, `./test-samples.sh` OK in both.
- Control on unpatched `master` in the same containers, so a clean result is attributable to the change rather than to the environment.

Two failures the added flags did **not** cause, both checked against unpatched `master` in the same container before attributing them:

- `-O2 -Werror` fails on `chdir` return values at `c/planarityApp/planarityCommandLine.c:218` and `:263`. Unpatched `master` fails identically. That is #316.
- `-funsigned-char -Werror` fails at `c/graphLib/io/graphIO.c:560`. Unpatched `master` fails identically. Filed separately.

<details><summary>Probed flag sets</summary>

```text
gcc-14   : -Wunused-result -Wignored-qualifiers -Wdiscarded-qualifiers
clang-18 : -Wunused-result -Wignored-qualifiers -Wdeprecated-non-prototype -Wincompatible-pointer-types-discards-qualifiers
```

</details>

### DIFF
--- a/m4/ax_compiler_flags_cflags.m4
+++ b/m4/ax_compiler_flags_cflags.m4
@@ -86,6 +86,21 @@ AC_DEFUN([AX_COMPILER_FLAGS_CFLAGS],[
         fi
 
         # "yes" flags
+        #
+        # This list is kept in step with the compiler flags in
+        # devEnvSetupAndDefaults/.vscode/tasks.json. Compiler-specific spellings
+        # are safe to list together because AX_APPEND_COMPILE_FLAGS probes each
+        # flag and drops the ones the active compiler does not accept, e.g.
+        # -Wdiscarded-qualifiers is gcc and
+        # -Wincompatible-pointer-types-discards-qualifiers is its clang analogue.
+        #
+        # Deliberately not mirrored from tasks.json:
+        #   -Wno-deprecated-declarations  a suppression, not a warning; the
+        #     editor tasks quiet it for local builds, and adding it here would
+        #     weaken the gate that compile-warnings.yml exists to enforce.
+        #   -fsanitize=address, -fdiagnostics-color, -fcolor-diagnostics,
+        #     -fansi-escape-codes  not warning flags; sanitisers have their own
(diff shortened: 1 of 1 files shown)


john-boyer-phd (approved):
It's great to have these extra warnings requested, and will be especially helpful once `-Werror` is added by the PR that will solve #316 

Thanks for adding all the comments, in addition to the extra warning flags, and really good job working out the details of the unsigned-char issue, in #319 .

PR #320, opened 2026-09-01
Return int from sf_getc and sf_ungetc so EOF survives unsigned char

### LINKED ISSUE #319: gp_Read rejects a valid graph on unsigned-char targets when an extension is attached
## Describe the bug

`sf_getc()` and `sf_ungetc()` are declared to return `char` (`c/graphLib/io/strOrFile.h:42` and `:50`), and `sf_getc()` returns `EOF` as a `char` (`strOrFile.c:284`). Plain `char` is unsigned on AArch64, ARM, PowerPC and s390x, so on those targets `EOF` becomes 255 and every `== EOF` test against the result is decided by the type, not by the data.

At true end of input in `_ReadGraph()` (`graphIO.c:559`) the branch is therefore always taken, `sf_ungetc()` does not recognise 255 as `EOF` and pushes it back, and `sf_fgets()` hands the caller a one-byte string `{0xff, '\0'}` as extra data.

With the default post-processor this is invisible: extra data is ignored and the read succeeds, which is why the test suite passes on AArch64 and adjacency-list output is byte-identical to x86-64. Attach an extension first and the fabricated byte is delivered to its post-processor, which rejects it.

## Reproduction steps

`one.txt`, a valid one-vertex graph:

```text
N=1
1: -1
```

```c
graphP g = gp_New();
gp_ExtendWith_DrawPlanar(g);   /* supported before gp_Read; see graphDrawPlanar_Extensions.c:152 */
int r = gp_Read(g, "one.txt"); /* OK = 1 */
```

## Expected behavior

`gp_Read()` returns `OK` on every platform. The file is valid and carries no extra data.

## Terminal logs

Same source, same commit, same compiler, two container architectures:

```text
x86_64  (signed char)    no extension: 1    DrawPlanar: 1
aarch64 (unsigned char)  no extension: 1    DrawPlanar: 0
```

Without the extension both agree, which isolates the post-processor path rather than the graph read.

## Environment

- OS: Ubuntu 24.04 containers, `linux/amd64` and `linux/arm64`
- Built from source at `36762b5`, `autoreconf -if && ./configure && make`
- gcc 13.3.0 in both containers

## Additional context

`compile-warnings.yml` cannot see this. It runs `ubuntu-latest` only, where `char` is signed and the comparison is in range. The AArch64 leg of `dev-platforms.yml` does run on the affected target, but calls plain `./configure` without `--enable-compile-warnings`, and its pull-request trigger is commented out. Building x86-64 with `-funsigned-char` reproduces the diagnostic at `graphIO.c:560` and would give architecture-independent coverage in the existing gate.

I found this while testing #318 and checked it against unpatched `master`, so it is not caused by that change. Related: #316, which reports the same gate failing to fail.

The usual fix is `int` in both wrappers, keeping the value as `int` through the `EOF` check and narrowing only after it is known not to be `EOF`. I am happy to prepare that if you want it, though it touches every caller and is a larger change than the warnings work.

Deln0r:
Fixes #319. Follows the plan agreed there: full int conversion rather than the `(char)EOF` casts.

## Type of change

### Updated

* `c/graphLib/io/strOrFile.h`, `strOrFile.c`: `sf_getc()` and `sf_ungetc()` now take and return `int`, following the stdio `getc()`/`ungetc()` contract, so EOF stays distinguishable from the byte 0xFF on builds where plain `char` is unsigned (the Linux ABIs for AArch64, ARM, PowerPC, s390x). All EOF-sensitive temporaries in `strOrFile.c` are `int` now.
* The conversion also closes the same defect in the opposite direction: a literal 0xFF byte from the string container sign-extended to EOF on signed-char builds, and bytes round-tripped through the unget buffer did the same. Both paths now normalize through `unsigned char`.
* `c/graphLib/io/g6-read-iterator.c`, `graphIO.c`: callers updated; the `(char)` casts at `sf_ungetc()` call sites are gone.
* `c/planarityApp/planarityCommandLine.c`: two quick-regression tests. `runReadWithExtensionAtEofTest()` reads a valid adjacency-list file with DrawPlanar attached, which is the #319 reproducer. `runHighByteRoundTripTest()` feeds the byte 0xFF through the string container and the unget buffer directly; this one bites on today's signed-char CI, not only on unsigned-char builds.

## Testing

- Positive control for both tests: `master`'s `graphLib` built against this branch's tests fails the suite (exit 255, the 0xFF round-trip test) on plain x86-64, and fails the extension-at-EOF test under `-funsigned-char`; with the patched `graphLib` the full suite passes 23/23 in both configurations.
- The #319 reproducer flips from `NOTOK` to `OK` on native linux/arm64 (Ubuntu 24.04 container) and on x86-64 with `-funsigned-char`; both checked against unpatched `master` in the same containers.
- CI-faithful matrix: gcc-14 and clang-18, `--enable-compile-warnings CFLAGS=-Werror`, Ubuntu 24.04: both build clean. Apple clang 21 on macOS arm64: build clean, `make check` and `test-samples.sh` pass.

The `--enable-unsigned-char` configure switch discussed in #319 is left for the CI follow-up, so this MR stays the mechanical type conversion plus its tests.

### DIFF
--- a/c/graphLib/io/g6-read-iterator.c
+++ b/c/graphLib/io/g6-read-iterator.c
@@ -249,7 +249,7 @@ int _g6_InitReaderWithStrOrFile(G6ReadIteratorP theG6ReadIterator, strOrFileP *p
 
 int _g6_InitReader(G6ReadIteratorP theG6ReadIterator)
 {
-    char charConfirmation = EOF;
+    int charConfirmation = EOF;
     int firstChar = '\0';
     int lineNum = 1;
     int order = NIL;
@@ -262,7 +262,7 @@ int _g6_InitReader(G6ReadIteratorP theG6ReadIterator)
     }
     else
     {
-        charConfirmation = sf_ungetc((char)firstChar, inputContainer);
+        charConfirmation = sf_ungetc(firstChar, inputContainer);
 
         if (charConfirmation != firstChar)
         {
@@ -283,7 +283,7 @@ int _g6_InitReader(G6ReadIteratorP theG6ReadIterator)
     }
 
     firstChar = sf_getc(inputContainer);
-    charConfirmation = sf_ungetc((char)firstChar, inputContainer);
+    charConfirmation = sf_ungetc(firstChar, inputContainer);
 
     if (charConfirmation != firstChar)
     {
@@ -430,7 +430,7 @@ int _g6_DetermineOrderFromInput(strOrFileP inputContainer, int *order)
             return NOTOK;
         }
 
-        sf_ungetc((char)graphChar, inputContainer);
+        sf_ungetc(graphChar, inputContainer);
 
         for (int i = 2; i >= 0; i--)
         {
--- a/c/graphLib/io/graphIO.c
+++ b/c/graphLib/io/graphIO.c
@@ -556,7 +556,7 @@ int _ReadGraph(graphP theGraph, strOrFileP *pInputContainer)
     // was OK.
     if (extraDataAllowed)
     {
-        char charAfterGraphRead = EOF;
+        int charAfterGraphRead = EOF;
         if ((charAfterGraphRead = sf_getc((*pInputContainer))) != EOF)
         {
             if (sf_ungetc(charAfterGraphRead, (*pInputContainer)) != charAfterGraphRead)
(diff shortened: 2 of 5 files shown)


john-boyer-phd (approved):
This is a great PR, Ian; thanks for doing it the best way! Now I can remove the line from the upcoming GraphML Reader Part 1 issue that said we wouldn't support 0xFF in ISO-8859-1 files.

And, chess-style !! for the typecasts on lines 320 and 612 of `strOrFile.c`.

PR #321, opened 2026-09-01
Fix signed integer overflow in random graph generation at N >= 46342

### LINKED ISSUE #315: Signed integer overflow in random graph generation: `planarity -rm N` fails for N >= 46342
`gp_CreateRandomGraphEx()` computes the maximum simple graph edge count in a 32-bit `int`, and the
multiplication happens before the shift, so `N * (N - 1)` has to fit rather than the halved value.

`c/graphLib/graph.c:1333` (master `36762b5`):

```c
    int N, maxNumEdges, maxPlanarEdges, numPlanarCoreEdges;
    ...
    maxNumEdges = (N * (N - 1)) >> 1;
```

The largest safe vertex count is 46341, because `46341 * 46340 = 2,147,441,940` fits in an `int` and
`46342 * 46341 = 2,147,534,622` does not. Above that the product is undefined behaviour and in
practice wraps. When it wraps negative, the clamp two lines later sets `numEdges` negative, the
`numEdges < N - 1` guard returns `NOTOK`, and the command fails outright:

```
$ planarity -rm 46342 /tmp/out.g6
Creating the random graph...
[ERROR] gp_CreateRandomGraphEx() failed
	on line 596 of 'c/planarityApp/planarityRandomGraphs.c'
$ echo $?
255
```

No output file is written. Whether a given N works depends only on the sign of the wrapped value, so
the failures are not contiguous:

| N | `N*(N-1)` | as int32 | `maxNumEdges` | `planarity -rm N` |
|---|---|---|---|---|
| 1,000 | 999,000 | fits | 499,500 | works |
| 46,341 | 2,147,441,940 | fits | 1,073,720,970 | works, last good N |
| 46,342 | 2,147,534,622 | -2,147,432,674 | -1,073,716,337 | fails, exit 255 |
| 46,343 | 2,147,627,306 | -2,147,339,990 | -1,073,669,995 | fails, exit 255 |
| 65,536 | 4,294,901,760 | -65,536 | -32,768 | fails, exit 255 |
| 100,000 | 9,999,900,000 | +1,409,965,408 | +704,982,704 | works |
| 1,000,000 | 999,999,000,000 | -728,379,968 | -364,189,984 | fails, exit 255 |
| 10,000,000 | 99,999,990,000,000 | +266,447,232 | +133,223,616 | works |

A UBSan build confirms the cause:

```
c/graphLib/graph.c: runtime error: signed integer overflow: 46342 * 46341 cannot be represented in type 'int'
c/graphLib/graph.c: runtime error: signed integer overflow: 1000000 * 999999 cannot be represented in type 'int'
c/graphLib/graph.c: runtime error: signed integer overflow: 10000000 * 9999999 cannot be represented in type 'int'
```

It is not specific to one platform or word size. The boundary sits at the same N on every target I
tried, which is what you would expect since it depends on `int` being 32 bits rather than on the
pointer width:

| Platform | compiler | `-rm 46341` | `-rm 46342` |
|---|---|---|---|
| Ubuntu 24.04 x86_64 | gcc 13.3.0 | exit 0 | exit 255 |
| Ubuntu 24.04 aarch64 | gcc 13.3.0 | exit 0 | exit 255 |
| Windows 11 x64 | MSVC 19.44 | exit 0 | exit -1 |
| Windows 11 x86 | MSVC 19.44 | exit 0 | exit -1 |

## Two related things in the same area

The same expression appears at two other places, with the same 46342 ceiling:

```
c/graphLib/graph.c:1229-1230   if (M > N * (N - 1) / 2)  /  M = N * (N - 1) / 2;   gp_CreateRandomGraph()
c/graphLib/graph.c:1978-1979   if (candidateEdgeCapacity > ((N * (N - 1)) >> 1))   gp_DynamicInsertEdge()
```

UBSan also flags `gp_GetRandomNumber()` in `c/graphLib/lowLevelUtils/apiutils.c`, where `N` starts at
`rand()` and can already be near `INT_MAX` before the folding adds to it:

```
c/graphLib/lowLevelUtils/apiutils.c:101:7: runtime error: signed integer overflow: 255 + 2147483516 cannot be represented in type 'int'
```

Line 100 has the same shape (`N += ((N & 0xFFFF0000) >> 16);`) though I did not observe it trip. The
mask to `0x7FFFFFFF` on line 102 looks like it was meant to keep the value in range, but it runs after
the additions rather than before them.

## Why CI does not catch it

`sanitisers.yml` runs ASan and UBSan only over `c/samples`, which is graphs on 8 vertices. `N` there is
four orders of magnitude below the threshold, so the job that would have caught this instantly cannot
reach the code path. Adding a single large N case would cover the whole class, and it is fast:
`planarity -rm 46342` under UBSan takes under a second.

## Suggested fix

Widen the arithmetic, or guard N. Something like:

```c
    long long maxNumEdgesLL = ((long long)N * (N - 1)) >> 1;
    maxNumEdges = maxNumEdgesLL > INT_MAX ? INT_MAX : (int)maxNumEdgesLL;
```

and the same at the other two sites. For `gp_GetRandomNumber()`, applying the `0x7FFFFFFF` mask to
`rand()` before the additions, or doing the folding in `unsigned`, removes the overflow.

## Where this came from

This is not long-standing code. `git log -S` puts the expression in `a785616` (2026-07-13), which
reached master through #279 ("Improve `gp_CreateRandomGraphEx` generation"). I mention it only because
it narrows the regression window, and because the test matrix in that PR tops out at N = 20, so
nothing in it goes near the threshold.

## Reproducing

```bash
git clone https://github.com/graph-algorithms/edge-addition-planarity-suite.git
cd edge-addition-planarity-suite
autoreconf -fi && ./configure && make

./planarity -rm 46341 /tmp/a.g6 ; echo "46341 -> $?"
./planarity -rm 46342 /tmp/b.g6 ; echo "46342 -> $?"

# under UBSan
make distclean
./configure CFLAGS="-O1 -g -fsanitize=undefined,signed-integer-overflow -fno-omit-frame-pointer"
make
UBSAN_OPTIONS=print_stacktrace=1 ./planarity -rm 46342 /tmp/c.g6
```

Verified on Ubuntu 24.04, gcc 13.3.0, x86_64, against master `36762b5` and the `JOSS` branch.

Raised while reviewing for JOSS, openjournals/joss-reviews#11163.

MODSetter:
Resolves #315

Part of the JOSS review, openjournals/joss-reviews#11163.

## Type of change

### Added

* None.

### Updated

* `c/graphLib/graph.c`
  * `gp_CreateRandomGraphEx()`: `maxNumEdges` becomes `long long` and the product is formed in that type, so the clique bound no longer wraps. The clamp on `numEdges` casts back to `int`, which is safe because that branch only runs when the bound is smaller than a value that already fit in an `int`.
  * `gp_CreateRandomGraph()`: the same bound computed the same way, before `M` is clamped.
  * `gp_DynamicInsertEdge()`: removed the cap on the doubled edge capacity, following the suggestion in #315 that the pitstop is no longer needed now that parallel edges are supported.
* `c/graphLib/lowLevelUtils/apiutils.c`
  * `gp_GetRandomNumber()`: `N` becomes `unsigned int` and the `0x7FFFFFFF` mask moves from before the modulus into the `return`, also following #315.

### Removed

* None.

## Testing

Ubuntu 24.04, gcc 13.3 for the default build, gcc-14 14.2 and clang-18 18.1 for the warning checks.

- I built a pristine `master` tree next to the patched tree and ran `planarity -rm N` on both. `-rm` asks for a random maximal planar graph, so the reported edge count has to be exactly `3N - 6`. Below the boundary the two trees agree exactly, so inputs that already worked are unaffected. At 46342 and above, `master` exits 255 without writing a graph while the patched tree returns the correct count.
- I rebuilt both trees with `-fsanitize=undefined`. On `master`, UBSan reports the overflow at `graph.c:1333`. At N = 100000 it reports the overflow and the program still exits 0, which is the quiet case: the wrapped bound lands above the requested edge count, so the output happens to look right. The patched tree produces no UBSan reports at any N I tried.
- `make check` passes on the patched tree.
- I built with `--enable-compile-warnings` on gcc-14 and clang-18. gcc-14 reports the two pre-existing `chdir` warnings and nothing else, clang-18 reports nothing. `long long` is new to this codebase and `-pedantic` is in that flag set, so this confirms it adds no warning of its own.
- For `gp_GetRandomNumber()` I compared the old and new versions over 2 million draws each for the ranges [0,1], [0,9], [1,6], [0,999], [-5,5], [100,100] and [0,65535]. Both stay inside their range and cover the same set of values.

<details><summary>Edge counts and UBSan reports, unpatched vs patched</summary>

```
############ functional: reported edge count must equal 3N-6 ############
N          | UNPATCHED (control)          | PATCHED
-----------+------------------------------+-----------------------------
1000       | 2994 edges  OK               | 2994 edges  OK
46341      | 139017 edges  OK             | 139017 edges  OK
46342      | exit 255 (no graph)          | 139020 edges  OK
46343      | exit 255 (no graph)          | 139023 edges  OK
65536      | exit 255 (no graph)          | 196602 edges  OK
100000     | 299994 edges  OK             | 299994 edges  OK

############ UBSan: signed overflow reports ############

--- N = 46342 ---
  UNPATCHED exit=255  runtime-error lines=1
c/graphLib/graph.c:1333:22: runtime error: signed integer overflow: 46342 * 46341 cannot be represented in type 'int'
  PATCHED   exit=0  runtime-error lines=0

--- N = 100000 ---
  UNPATCHED exit=0  runtime-error lines=1
c/graphLib/graph.c:1333:22: runtime error: signed integer overflow: 100000 * 99999 cannot be represented in type 'int'
  PATCHED   exit=0  runtime-error lines=0
```

</details>

<details><summary>gp_GetRandomNumber output range, old vs new</summary>

```
  old  [0, 9]  min=0  max=9      in-range=yes  values-never-hit=0/10
  new  [0, 9]  min=0  max=9      in-range=yes  values-never-hit=0/10
  old  [-5, 5]  min=-5  max=5    in-range=yes  values-never-hit=0/11
  new  [-5, 5]  min=-5  max=5    in-range=yes  values-never-hit=0/11
  old  [0, 65535]  min=0  max=65535  in-range=yes  values-never-hit=255/65536
  new  [0, 65535]  min=0  max=65535  in-range=yes  values-never-hit=255/65536
```

</details>

<details><summary>How often the old fold was actually undefined</summary>

```
seeds in [INT_MAX-40000, INT_MAX] that overflow line 2: 255
  first such seed: 2147450626 (0x7FFF7F02)
  last  such seed: 2147450880 (0x7FFF8000)
  probability per rand() call: 1.19e-07
scan.c:9:7: runtime error: signed integer overflow: 2147483393 + 255 cannot be represented in type 'int'
```

</details>

## Notes for review

Three places where I did not follow #315 exactly.

* The mask in the suggested `return` line reads `0x7FFFFFF`, seven Fs, where the existing code has `0x7FFFFFFF`, eight. I took the short one as a typo and kept eight. Seven clears bits 27 through 31, which would change the result once the range passes 2^27.
* I wrote `return (int)(N & 0x7FFFFFFF) + NMin;` instead of `return (int)((N & 0x7FFFFFFF) + NMin);`. Adding a negative `NMin` inside the unsigned expression wraps and then converts back, which works on two's complement but is not guaranteed. Leaving `+ NMin` outside the cast keeps that addition signed. I can switch to the original form if you prefer it.
* In `gp_DynamicInsertEdge()` I deleted more than the two lines holding the product. Removing them leaves `int N` unused, and `-Wunused-variable` would then break the `-Werror` build that #316 sets up, so the declaration had to go with them. That in turn leaves the following `if (newEdgeCapacity <= gp_GetEdgeCapacity(theGraph))` block unable to change anything, since `newEdgeCapacity` is now always the doubled capacity, and its comment opens with "However" and refers to a cap that is gone. I removed that block too. I can restore it if you would rather handle it separately.

Two things I noticed and left alone.

* Moving the mask after the modulus changes which sequence a given seed produces. Nothing in `make check` depends on that, but it is a visible change for anything that seeds `rand()` and expects particular graphs.
* `gp_GetEdgeCapacity(theGraph) << 1` in the same function can still overflow for a large enough capacity. Same class of problem, but outside #315, so I have not touched it.

### DIFF
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -1200,6 +1200,7 @@ graphP gp_DupGraph(graphP theGraph)
 int gp_CreateRandomGraph(graphP theGraph)
 {
     int N, M, u, v, m;
+    long long maxSimpleEdges;
 
     if (theGraph == NULL)
     {
@@ -1226,8 +1227,14 @@ int gp_CreateRandomGraph(graphP theGraph)
 
     M = gp_GetRandomNumber(7 * N / 8, theGraph->edgeCapacity);
 
-    if (M > N * (N - 1) / 2)
-        M = N * (N - 1) / 2;
+    /* N * (N - 1) overflows a 32-bit int at N == 46342, before the halved
+            result would, so the simple undirected edge bound is formed in a
+            wider type. */
+
+    maxSimpleEdges = ((long long)N * (N - 1)) / 2;
+
+    if (M > maxSimpleEdges)
+        M = (int)maxSimpleEdges;
 
     for (m = N - 1; m < M; m++)
     {
@@ -1316,7 +1323,8 @@ int gp_CreateRandomGraphEx(graphP theGraph, int numEdges)
 {
     randomGraphEdgeRec *optionalEdges = NULL;
     randomGraphFaceRec *faces = NULL;
-    int N, maxNumEdges, maxPlanarEdges, numPlanarCoreEdges;
+    int N, maxPlanarEdges, numPlanarCoreEdges;
+    long long maxNumEdges;
     int lowerVertex, upperVertex, faceCapacity, optionalEdgeCapacity;
     int optionalEdgeCount = 0, faceCount = 0, addAllPlanarEdges;
     int Result = OK;
@@ -1330,14 +1338,16 @@ int gp_CreateRandomGraphEx(graphP theGraph, int numEdges)
     N = gp_GetN(theGraph);
     lowerVertex = gp_LowerBoundVertices(theGraph);
     upperVertex = gp_UpperBoundVertices(theGraph);
-    maxNumEdges = (N * (N - 1)) >> 1;
+    // Formed in a wider type because N * (N - 1) overflows a 32-bit int at
+    // N == 46342, before the halved result would.
+    maxNumEdges = ((long long)N * (N - 1)) >> 1;
     maxPlanarEdges = 3 * N - 6;
 
     if (numEdges > theGraph->edgeCapacity)
         numEdges = theGraph->edgeCapacity;
 
     if (numEdges > maxNumEdges)
-        numEdges = maxNumEdges;
+        numEdges = (int)maxNumEdges;
 
     if (numEdges < N - 1)
         return NOTOK;
@@ -1967,23 +1977,10 @@ int gp_DynamicInsertEdge(graphP theGraph, int u, int e_u, int e_ulink,
 
     if (gp_GetM(theGraph) >= theGraph->edgeCapacity && sp_IsEmpty(theGraph->edgeHoles))
     {
-        // The candidate edge capacity is double the current capacity
-        int candidateEdgeCapacity = gp_GetEdgeCapacity(theGraph) << 1;
-        int N = gp_GetN(theGraph);
-        int newEdgeCapacity = candidateEdgeCapacity;
-
-        // If the candidate edge capacity exceeds the number of edges
-        // needed in an undirected clique on N vertices, then attempt
-        // to use that as the new edge capacity.
-        if (candidateEdgeCapacity > ((N * (N - 1)) >> 1))
-            newEdgeCapacity = ((N * (N - 1)) >> 1);
-
-        // However, if the edge capacity is already greater than or
-        // equal to that maximum capacity needed for an undirected
-        // clique on N vertices, then we allow the capacity to double
-        // beyond the simple undirected graph limit.
-        if (newEdgeCapacity <= gp_GetEdgeCapacity(theGraph))
-            newEdgeCapacity = candidateEdgeCapacity;
+        // The new edge capacity is double the current capacity. Parallel
+        // edges are supported, so the capacity is not capped at the number
+        // of edges needed for an undirected clique on N vertices.
+        int newEdgeCapacity = gp_GetEdgeCapacity(theGraph) << 1;
 
         if (gp_EnsureEdgeCapacity(theGraph, newEdgeCapacity) != OK)
             return NOTOK;
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -1980,9 +1980,19 @@ int gp_DynamicInsertEdge(graphP theGraph, int u, int e_u, int e_ulink,
         // The new edge capacity is double the current capacity. Parallel
         // edges are supported, so the capacity is not capped at the number
         // of edges needed for an undirected clique on N vertices.
-        int newEdgeCapacity = gp_GetEdgeCapacity(theGraph) << 1;
+        long long newEdgeCapacity = ((long long)gp_GetEdgeCapacity(theGraph)) << 1;
 
-        if (gp_EnsureEdgeCapacity(theGraph, newEdgeCapacity) != OK)
+        // If left shift would overflow signed integer, then cap it at INT_MAX.
+        if (newEdgeCapacity > INT_MAX)
+        {
+            newEdgeCapacity = INT_MAX;
+            // If unable to allocate more edges due to already being at INT_MAX,
+            // then return failure.
+            if (newEdgeCapacity <= gp_GetEdgeCapacity(theGraph))
+                return NOTOK;
+        }
+
+        if (gp_EnsureEdgeCapacity(theGraph, (int)newEdgeCapacity) != OK)
             return NOTOK;
     }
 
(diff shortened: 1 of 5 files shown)


john-boyer-phd (approved):
Looks great Rohan. I also liked your suggestion in the issue to do a larger case of N for the sanitizers, and it turns out we have an easy way to add that. I added one to test for signed overflow on clique size calculation, and another larger graph for unsigned. Both are really fast-- for release compiling under 0.1 seconds for all random graph testing including the two bigger graphs, and only a couple of seconds for sanitizing.

PR #326, opened 2026-09-03
Add --enable-unsigned-char and an unsigned-char CI leg

Deln0r:
The CI follow-up agreed in #319, now that the int conversion (#320) is in.

## Type of change

- Build system / CI change (non-breaking; no library code touched)

### Updated

* `configure.ac` gains `--enable-unsigned-char`, which appends `-funsigned-char` to `WARNING_CFLAGS` via `AX_APPEND_COMPILE_FLAGS`. Riding the existing `AM_CFLAGS = $(WARNING_CFLAGS)` plumbing means the workflow needs no `CFLAGS` assignment, so the autoconf default `-g -O2` stays intact, which was the concern raised in #319 after #322. Off by default: no build is affected unless the switch is given.
* `compile-warnings.yml` gains one matrix leg, gcc-14 with the switch, named `gcc-14 (unsigned char)`. Char-signedness regressions of the #319 class now surface on the x86-64 runner on every pull request.

## Testing

- With the switch: `WARNING_CFLAGS` contains `-funsigned-char`; full build under `--enable-compile-warnings=error` is clean; `make check` and `test-samples.sh` pass, including the two #320 regression tests, on Apple clang 21 and on gcc-14/Ubuntu 24.04.
- Without the switch: `WARNING_CFLAGS` does not contain the flag, byte-identical to master behavior.
- Control from the #320 work: the pre-#320 tree fails exactly this configuration, so the leg is known to be able to fail.

### DIFF
--- a/.github/workflows/compile-warnings.yml
+++ b/.github/workflows/compile-warnings.yml
@@ -12,7 +12,7 @@ concurrency:
 
 jobs:
   tests:
-    name: ${{ matrix.compiler }}
+    name: ${{ matrix.compiler }}${{ matrix.unsigned-char == 'yes' && ' (unsigned char)' || '' }}
     timeout-minutes: 60
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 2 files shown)


john-boyer-phd (approved):
YES!!

PR #327, opened 2026-09-04
Validate vertex and edge capacities before int arithmetic

### LINKED ISSUE #325: Text readers pass an unvalidated vertex count into int capacity arithmetic
## Describe the bug

The three text readers take the vertex count straight from the file and hand it to `gp_EnsureVertexCapacity()`, which only rejects `N <= 0` (`graph.c:257`). Everything above that goes into capacity arithmetic that is done in `int`, so a large count overflows before anything checks whether it is plausible.

On the default edge capacity the first overflow is at **N = 178,956,971**, in `(theGraph->edgeCapacity << 2) + 2` at `graph.c:285`. Higher counts reach the others: `3 * N` at `graph.c:277`, `2 * 2 * 3 * N` at `graph.c:287`, and the bound macros `gp_UpperBoundVertexStorage` and `gp_UpperBoundEdgeStorage` in `graph.h`.

The `.g6` reader does bound its order, rejecting `n > 100000` with a diagnostic at `g6-read-iterator.c:421`. The text readers have no equivalent, which is what makes this easy to hit by accident with a truncated or edited file.

## Reproduction steps

Any of these, with a UBSan build (`./configure CFLAGS="-fsanitize=undefined -fno-sanitize-recover=all -g -O1" LDFLAGS="-fsanitize=undefined"`):

```bash
printf 'N=2147483647\n' > n.txt          # adjacency list
planarity -s -q -p n.txt out.txt

printf '2000000000\n' > m.txt            # adjacency matrix
planarity -s -q -p m.txt out.txt

printf 'LEDA.GRAPH \nvoid\nvoid\n-2\n\n2000000000 \n' > l.txt
planarity -s -q -p l.txt out.txt
```

```
c/graphLib/graph.c:285:41: runtime error: left shift of 536870913 by 2 places cannot be represented in type 'int'
c/graphLib/graph.c:277:113: runtime error: signed integer overflow: 3 * 2000000000 cannot be represented in type 'int'
```

The library entry points reach it too: `gp_ReadFromString(g, "N=2000000000\n")` produces the same report, and `gp_EnsureVertexCapacity()` is public in `graph.h`.

Boundary, measured: `N=178956970` is clean, `N=178956971` reports at `graph.c:285`.

## Expected behavior

A vertex count whose capacity arithmetic does not fit in `int` is rejected with a diagnostic, the way the `.g6` reader already rejects large orders, rather than being computed with.

## Terminal logs

What a normal build does today, same inputs, no sanitizer:

| N | result |
|---|---|
| 100,000,000 | `NOTOK`, exit 255, after attempting roughly 25 GB of allocations across `V`, `E`, `edgeHoles` and the stack |
| 178,956,971 and above | `NOTOK`, exit 255 |
| negative, or 0 | `NOTOK`, rejected cleanly by the existing guard |

So on the builds I tested nothing is corrupted: the allocations fail and `_ClearGraph()` runs before `_InitVertices()` or `_InitEdges()` touch anything. I would not state that more strongly than "observed to fail cleanly", because the wrapped values are not guaranteed to be the two's-complement ones this reasoning assumes, they need not be negative, and `sp_New()` multiplies the capacity by `sizeof(int)` where a narrower `size_t` could wrap to a small allocation instead of failing.

## Environment

- Built from source at `875b82b`, `autoreconf -if && ./configure && make`
- Apple clang 21 on macOS arm64, and gcc 13.3.0 on Ubuntu 24.04 x86-64

## Additional context

Same class as #315, which you fixed for `-rm`, and reached from file input rather than the command line. The `sanitisers.yml` job runs over `c/samples`, where the largest order is 8, so it cannot see this; the large-N cases you added for #315 exercise the generator rather than the readers.

Three extension overloads repeat `3 * N` before delegating, so a check placed only inside `_EnsureVertexCapacity()` would come too late for them: `graphK4Search_Extensions.c:224`, `graphK33Search_Extensions.c:260`, `graphDrawPlanar_Extensions.c:378`.

The smallest fix that covers every route looks like a validation in the public `gp_EnsureVertexCapacity()` wrapper before dispatch: compute the capacity, the two bounds and the stack size in `long long`, reject if any exceeds `INT_MAX`. That matches the widened-arithmetic style already used at `graph.c:1230` and `graph.c:1980`. `gp_EnsureEdgeCapacity()` has the same gap for capacities that cannot support `2 * capacity + 2` edge indexes.

Happy to prepare that patch with a test if you want it, though it touches a public entry point, so tell me if you would rather scope it differently.

Deln0r:
Fixes #325, following the design set there: the vertex-side check uses the greater of a pre-set `edgeCapacity` and `DEFAULT_EDGE_CAPACITY_FACTOR * N`, and `gp_EnsureEdgeCapacity()` gets a stack-capacity check stringent enough to shield the `gp_UpperBoundEdgeStorage()` uses in the extensions.

## Type of change

### Updated

* `c/graphLib/graph.c`: both public wrappers validate in `long long` before dispatch. One test carries the load: the stack bound `(edgeCapacity << 2) + 2` is the strict maximum of every quantity later derived from `N` and the capacity, in the default and `USE_0BASEDARRAYS` builds and in the three extension overloads, so nothing downstream can overflow once it passes. The comments carry the dominance argument so it stays maintainable.
* `c/planarityApp/planarityCommandLine.c`: a quick-regression test pinning both boundaries. First rejected vertex count is 178,956,971 and first rejected edge capacity is 536,870,912, both derived in-code from `INT_MAX` rather than hardcoded; the edge-capacity accept side is exercised at its exact maximum, which allocates nothing because the capacity is stored before `gp_EnsureVertexCapacity()`.

The observable change for the text readers: `N=2147483647` in an adjacency-list, adjacency-matrix or LEDA file now returns `NOTOK` immediately, where each previously ran signed arithmetic past `INT_MAX` at `graph.c:277`/`:285` and attempted multi-gigabyte allocations.

## Testing

- UBSan build: the three reproducers from #325 are clean and fail fast; `N=178956970` (last valid) stays clean; full `-test` suite passes 24/24 under the sanitizer.
- Both gates on Ubuntu 24.04: gcc-14 and clang-18 under `--enable-compile-warnings=error`, plus the new `--enable-unsigned-char` leg, all clean, suite 24/24.
- macOS Apple clang 21: build clean under the error gate, `make check` and `test-samples.sh` pass.
- Boundary exactness cross-checked independently: `12N+2 = 2,147,483,642` at the last accepted `N`, `2,147,483,654` at the first rejected.

### DIFF
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -256,12 +256,29 @@ void _InitFunctionTable(graphP theGraph)
 
 int gp_EnsureVertexCapacity(graphP theGraph, int N)
 {
+    long long effectiveEdgeCapacity = 0;
+
     // valid params check
     if (theGraph == NULL || N <= 0)
         return NOTOK;
 
     // Should not call init a second time; use reinit
     if (gp_GetN(theGraph) > 0)
+        return NOTOK;
+
+    // Reject a vertex count whose capacity arithmetic cannot be represented
+    // in int (issue #325). The effective edge capacity is the greater of a
+    // pre-set edgeCapacity and DEFAULT_EDGE_CAPACITY_FACTOR * N, because
+    // gp_EnsureEdgeCapacity() may legitimately have stored a lower value
+    // before this call. The stack holds (edgeCapacity << 2) + 2 entries,
+    // which strictly exceeds every other derived quantity (vertex storage,
+    // edge storage and the 2 * 2 * DEFAULT_EDGE_CAPACITY_FACTOR * N + 2
+    // fallback), so one test on it shields them all, including the
+    // gp_UpperBoundEdgeStorage() uses in the algorithm extensions.
+    effectiveEdgeCapacity = (long long)DEFAULT_EDGE_CAPACITY_FACTOR * N;
+    if (theGraph->edgeCapacity > effectiveEdgeCapacity)
+        effectiveEdgeCapacity = theGraph->edgeCapacity;
+    if ((effectiveEdgeCapacity << 2) + 2 > INT_MAX)
         return NOTOK;
 
     return theGraph->functions->fpEnsureVertexCapacity(theGraph, N);
@@ -424,6 +441,15 @@ void _ResetGraphStorage(graphP theGraph)
 int gp_EnsureEdgeCapacity(graphP theGraph, int requiredEdgeCapacity)
 {
     if (theGraph == NULL || requiredEdgeCapacity <= 0)
+        return NOTOK;
+
+    // Reject a capacity whose stack arithmetic cannot be represented in int
+    // (issue #325). Together with the guard in gp_EnsureVertexCapacity(),
+    // this keeps every stored capacity able to support its stack and edge
+    // storage arithmetic: the vertex-side guard covers the derived
+    // quantities computed from N, and this test covers the ones computed
+    // from a stored or requested capacity
+    if ((((long long)requiredEdgeCapacity) << 2) + 2 > INT_MAX)
         return NOTOK;
 
     if (theGraph->edgeCapacity >= requiredEdgeCapacity)
--- a/c/graphLib/graph.c
+++ b/c/graphLib/graph.c
@@ -276,7 +276,7 @@ int gp_EnsureVertexCapacity(graphP theGraph, int N)
     // fallback), so one test on it shields them all, including the
     // gp_UpperBoundEdgeStorage() uses in the algorithm extensions.
     effectiveEdgeCapacity = (long long)DEFAULT_EDGE_CAPACITY_FACTOR * N;
-    if (theGraph->edgeCapacity > effectiveEdgeCapacity)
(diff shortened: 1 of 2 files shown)


john-boyer-phd (approved):
This looks great, Ian, thank you for flagging and solving this issue. 

I really only made one minor change to the test routine to prevent emission of error messages during subtests that are intended to cause the APIs to return `NOTOK`. 

Eventually, every `NOTOK` should get a `gp_ErrorMessage()` so that an explanation of the problem is given. This is now possible due to the better quiet mode handling added to 5.0, but it's about a 1,000 things to add, so it's on the backburner. Meanwhile, `DEBUG` mode has had (for a  very long time) a compile-time overload of `NOTOK` that emits an error message every time it happens, so we have to either prevent `NOTOK` from showing up in the happy paths of the code, or suppress the error messages, as I did, when the happy path is _intended_ to produce a `NOTOK`.

Issue #38, opened 2024-03-25
EPIC: Support iterative reading of sparse6 and incremental sparse6 formats


john-boyer-phd:
Nauty supports output of a ["sparse6" and incremental "sparse6" file formats](https://users.cecs.anu.edu.au/~bdm/data/formats.txt) to contain all graphs generated according to its parameterization. Subject to some experimentation, it is likely that the incremental "sparse6" format could significantly improve performance of algorithm testing, i.e., performance and output from command lines of the form `planarity -t [-q] C I O`. 

- [ ] Begin by using Nauty to generate all simple undirected graphs of order 10 and each of the possible numbers of edges in separate files. Do this for both the "graph6" and the incremental "spase6" formats so that we can get a sense of when "graph6" is less versus more efficient in terms of storage than incremental "sparse6". It may be necessary to also do this for order 9 and 11 to get a better sense.

The incremental "sparse6" format is expected to helps in the following ways:
- More of the files generated may be able to fit entirely into memory such that read iteration of graph reading occurs over an in-memory string rather than against a file. 
- Rather than reinitializing and fully rebuilding each graph to be tested with an algorithm, the incremental format offers the possibility of  simply adding an edge or edges to a previously given graph.

If the incremental "sparse6" format looks promising enough, then further work will be done to define the work items in this epic.

For transformation of the first graph in a file to the "sparse6" format, the letter 's' is used, i.e., `planarity -t [-q] -ts I O`. Since only one graph is transformed, the incremental format is not applicable as an output format.

Deln0r:
I would like to take this EPIC, starting with the measurement checkbox: orders 9 through 11 generated in graph6 and incremental sparse6, storage and read-throughput compared, results posted here before any code. The reader itself I would build the way I build wire-format ports elsewhere: a fixture set pinned against nauty's own tools as the reference, exercised in both directions, then the iterator implemented to the same API shape as the existing g6 one. If you or Wanda have design expectations beyond what is written in the issue, tell me before I start.

Deln0r:
Measurements for the first checkbox, before any code. Method: `geng -q` in its canonical order, conversions with `copyg -s` and `copyg -i`, read cost measured as `copyg -g` into `/dev/null`, so the reader is nauty's own. Apple M3, single runs; sizes are exact, times are indicative.

| order | graphs | graph6 | sparse6 | incremental s6 | is6/g6 | read g6 | read is6 |
[...]

Issue #37, opened 2024-03-25
Epic: Improve Graph Read/Write Error Detection and Reporting


john-boyer-phd:
There are several improvements that can be made throughout the `graphIO.c` code:

- [ ] All instances of `return NOTOK` and `RetVal = NOTOK` should be preceded with an explanatory `ErrorMessage()` to indicate what has gone wrong, in keeping with the level of information now provide by the new "graph6" reader. For example, there should be an error message in `_ReadAdjList()` that indicates loop edges are not supported (until/unless they are). 
- [ ] The adjacency matrix reader should be amended to detect the failure of `fscanf()` and `sscanf()` as well as successful reading of integer values other than 0 and 1, and provide an appropriate `ErrorMessage()` and `return NOTOK` when such events occur.
- [ ] Amend `ReadAdjList()` and `_ReadLEDAGraph()` to efficiently detect and report an error  and `return NOTOK` if parallel edges are detected (until/unless those are supported).
- [ ] Amend the LEDA graph format reader to detect a loop edge and report an error and `return NOTOK`
- [ ] Amend `_WriteAdjMatrix()` and the new "graph6" writer to report an error and `return NOTOK` if the graph being written is a digraph.

Issue #253, opened 2026-06-23
Increase Support of Parallel Edges


john-boyer-phd:
Support the addition of a parallel edge, i.e., an edge of the form (_v, w_) when another edge (_v, w_) already exists. This is particularly important for many directed graph use cases.

- [ ] The edge addition methods (e.g.,  `gp_AddEdge()` and `gp_InsertEdge()`) already create a new pair of edge records to represent a new edge, whether or not it is parallel to a pre-existing edge. This is intended behavior. Consequently, parallel edges contribute to degree counting methods (e.g.,  `gp_GetVertexDegree()`) according to their multiplicity, not as one. This design allows separate metadata and directionality to be imparted with ease.

- [ ] Add a flag `#define GRAPHFLAGS_PARALLELEDGEDETECTED 4` just above `gp_GetGraphFlags()` in `graph.h`. This flag will initially be clear due to the use of `calloc()` in `gp_New()`. Ensure that the default `_ResetGraphStorage()` method clears the flag. Ensure that `gp_CopyAdjacencyLists()` and `gp_CopyGraph()` (and hence `gp_DupGraph()`) copy the flag value from the source to the destination graph.

- [ ] Create new `graphEdgeDetector.c/.h` source files. Add to `makefile.am`, with the `.h` being package private, not public. Add a `#include` for `graphEdgeDetector.h` to `graph.private.h`. Add the copyright messages to both, and add the `extern "C"` construct to the header file.

- [ ] The `graphEdgeDetectorStruct` should be declared in the header file and contain an `unsigned *edgeDetector` and an `int edgeDetectorCapacity`. Also declare an `graphEdgeDetectorP` pointer type, as is done throughout the graphLib.

- [ ] Declare in the header file and implement in the source file a method with signature `graphEdgeDetectorP ged_New(int theCapacity)`. Returns `NULL` on error, such as memory allocation failure or non-positive `theCapacity` parameter value. Use `calloc` to allocate `theCapacity` unsigned integers, and assign to the `edgeDetector` data member. Assign `theCapacity` parameter value to the `edgeDetectorCapacity` data member.

[...]

PR #329, opened 2026-09-05
Add a sparse6 and incremental sparse6 read iterator (#38)

Deln0r:
Second step of #38, after the measurements posted there.

**What this adds**

A streaming reader for nauty's sparse6 (`:` lines) and incremental sparse6 (`;` lines) formats, `S6ReadIterator` in `c/graphLib/io/s6-read-iterator.[ch]`, with the same API shape as the graph6 read iterator: `s6_NewReader()`, `s6_InitReaderWithString()`, `s6_InitReaderWithFileName()`, `s6_ReadGraph()`, `s6_EndReached()`, `s6_FreeReader()`. `gp_Read()` and `gp_ReadFromString()` dispatch a first line beginning with `:` or `;`, or with the `>>sparse6<<` header, to it, so `planarity -x -(gam)` accepts sparse6 input as of this PR. `planarity -t` over sparse6 files is the next step and is not in here.

**Design points I would like you to check**

- The bit stream is decoded straight from the input container rather than from a buffered line: a graph6 line has a fixed length per order, a sparse6 line grows with the edge count, so there is no line length to allocate for. Decoding follows the specification and nauty's `stringtograph()`: running vertex `v`, `x > v` moves `v`, pairs that push `v` to the order or beyond are padding, an incomplete final pair is discarded.
- A `;` line toggles each decoded edge on the graph left by the previous line, so the graph is not reset there; a later `:` line resets it as the graph6 reader does. Toggling is `gp_FindEdge()` then `gp_DeleteEdge()` or `gp_DynamicAddEdge()`, so its cost is the adjacency list walk; that is fine for the `-t` sweeps this is meant for (order 12 and below) and I have measured it on dense random graphs of order 1000 at about 0.2 s per graph, in case that matters to you.
- Loop edges and repeated edges on a `:` line are errors, with the vertex and line number in the message. nauty itself is of two minds on repeats: its library decoder `stringtograph()` collapses a repeated pair into one edge, while `copyg` and `listg` go through `stringtograph_inc()`, which flips the edge twice and drops it (`:Ab`, the order-2 graph with edge 0-1 written twice, comes out of `copyg -g` as `A?`). I did not think either silent behaviour was worth matching, and #37 asks for parallel edges to be reported. The repeat check is a per-vertex stamp that is exact in constant time: an edge `{x, v}` is decoded from a pair whose `x` and `v` are fixed by the edge and `v` never decreases along a line, so a repeat can only occur while `v` is unchanged. My first version walked the adjacency list per edge and was 27 times slower than the graph6 reader on dense order-1000 input; with the stamp it is within 3x, and the difference is `gp_DynamicAddEdge()` bookkeeping.
- Line terminators may be LF, CR or CRLF, as in the graph6 reader. nauty rejects CR.

**Tests**

`N5-all.s6`, `N5-all.is6` and `nauty_example.s6` were produced from the existing `.g6` samples with nauty's `copyg -s` and `copyg -i`, and I confirmed `copyg -g` maps them back byte for byte. `runSparse6ReadTests()` reads them in lockstep with the graph6 reader, from a file and from a string, comparing every graph by graph6 encoding and edge count (the encoding alone would not see a doubled edge). String cases cover the specification example, header, CRLF, a missing terminator, an incomplete final pair, all graphs of orders 2 and 4 (the `n = 2^k` padding rule), the four-byte order encoding, the incremental toggle, order 1 through its zero-width vertex field, sixteen inputs that must be rejected (loops, repeats adjacent and interleaved, bytes outside 63..126 including 0xFF, order 0, the eight-byte order, order change, blank line, digraph6, header followed by a newline), and reinitialization of a reader after a failed initialization. Outside the repository I ran the same lockstep over every graph of orders 2 to 8 (12,346 at order 8), random graphs of orders 16 to 1000 and the full order-10 set (12,005,168 graphs), all with zero mismatches, and I checked the tests themselves by breaking the reader in fourteen deliberate ways: eleven of them fail a named case, and the other three are either masked by a neighbouring guard (a `;` first line still fails the `:` check, order 0 still fails the capacity check) or need more input than a unit test should carry (the cap that stops `v` counting past the order).

Read-only throughput, Apple M3, single runs, the same 12,005,168 graphs of order 10 as in the measurement comment:

| file | size | read time |
|---|---|---|
| `n10.g6` (graph6 reader) | 120.1 MB | 3.0 s |
| `n10.is6` (this reader) | 65.3 MB | 1.6 s |

Two things I noticed on the graph6 side while building the oracle, which I have left alone here: `_g6_ReadGraphFromStrOrFile()` returns `OK` after `g6_ReadGraph()` fails, so `planarity -x` on a truncated `.g6` line exits 0 and writes an empty graph, and neither the graph6 reader nor writer handles order 1 (`@`, which `geng 1` emits). I can file those separately if you want them tracked.


john-boyer-phd (changes requested):
This is a phenomenal contribution, thank you Ian! I am really looking forward to including sparse6 (for completeness) and especially incremental sparse6 (the exciting one).

Could you please make minor adjustments to this PR to resolve the merge conflict that came from merging the prior vertex/edge capacity PR, and to rename the `.is6` file to end with `.inc.s6`? Apparently, incremental sparse6 files are also just ended with `.s6` in nauty, but since we do have to distinguish it from the non-incremental, just using something before the suffix would suffice. 

Regarding your design point questions:

 * Yes, reading straight from the container rather than a buffered line is just fine. FILEs are buffered anyway, and so we just did it that way with Graph6 because it seemed like it made things a bit easier for us at the time. 100K vertices per graph seemed like more than enough per graph, so that's sufficient. Collections of bigger graphs should probably use incremental sparse6 anyway, due to the file size for adjacency matrix representations in graph6.
 * Using `gp_FindEdge()` before `gp_DeleteEdge()` seemes unavoidable so that's all good. It's conceivable to add data structures that would cause `gp_FindEdge()` to be expected constant time, but this is I/O anyway, so best to keep the data structures lighter weight and less memory and optimize `gp_FindEdge()`-then-`gp_DeleteEdge()` only if/when we hit a use case where it would make a big difference. Until then, good to go as is.
 * Good to error out on loop edges and parallel (repeated) edges at this time, as the initial implementation can be adapted once those are supported or better supported. Technically, we do support adding parallel edges, but someone has to write their own code to detect them and remove them before calling methods like `gp_Embed()`, so it would be better to have something along the lines of #253 in place before going any farther with parallel edge support anyway. 
 * Your line terminator support seems preferable. Even though nauty would never generate it, there doesn't seem to be a reason to forbid legacy CR-only linefeeds. Nobody will ever mean anything different by it. We just won't generate them when we write sparse6 files either.
 
Thank you again, Ian, for moving EAPS forward with this work.

Deln0r:
Done: rebased onto master after #327 and renamed the incremental sample to N5-all.inc.s6 (file, samples/Makefile.am, the tests and the commit message). While rebasing I also adopted the QUIETMODE_ALL convention from your change to runCapacityLimitTests for the sparse6 reject cases, so DEBUG builds stay quiet on the intended NOTOKs. The same four CI-equivalent builds (gcc-14, gcc-14 with -funsigned-char, clang, clang with ASan+UBSan) are green on the new commit.

Deln0r:
One more commit, found while wiring `-t` over incremental input: `gp_DeleteEdge()` leaves a hole unless the pair was the last one, and DrawPlanar declines any graph with holes (the check at graphDrawPlanar.c:576), so `-t -d` on n8.mALL converted to incremental sparse6 failed at the ninth graph, two edges in eighteen records. The reader now fills each hole with the last pair right away, so a graph read incrementally has the same dense storage as a freshly read one; the lockstep test asserts it and fails at the sixth graph of N5-all.inc.s6 with the fill disabled.

john-boyer-phd:
This is looking great, Ian (@Deln0r ). Just so you know what's going on, we're going to pause a beat and roll out the 5.1 release, and once that tag is created (which will be pretty quickly), then I'll refresh this PR with in and then it can roll in as one of the first things in 5.2.

Deln0r:
Sounds good, thanks for the heads-up. The -t support for sparse6 input and the writer are rebased on your merged head locally and will follow as separate PRs once this lands in 5.2.

c/graphLib/io/s6-read-iterator.c
```diff
@@ -0,0 +1,957 @@
...
+                                "expected a printable ASCII character in the "
+                                "range 63 to 126.",
+                                lineNum);
+                return NOTOK;
+            }
+
+            n = (n << 6) | (orderGroups[i] - 63);
+        }
+
+        if (n > 100000)
```
john-boyer-phd: It would be reasonable to allow n <= 258047 here. The limit of 100K over in graph6 is because it is doing a large line allocation that must accommodate a form of adjacency matrix storage, whereas here you are reading character by character, so the graph6 limit need not apply. Technically, you could go larger, but it seems best to let someone have the problem before we try to accommodate it.

c/graphLib/io/s6-read-iterator.c
```diff
@@ -0,0 +1,957 @@
...
+
+    // One entry per vertex, used to detect a repeated edge on a ':' line
+    // in constant time. Each edge {x, v} with x <= v is decoded from a
+    // pair whose x and v are fixed by the edge, and v never decreases
+    // along a line, so a repeat of the edge can only occur while v is
+    // unchanged. The entry for x records the (line, v) of the last edge
+    // decoded on x as lineNum * (order + 1) + v + 1, so the check is a
+    // comparison with no per-line reset. Needs long long because
+    // lineNum and order can each exceed 2^16.
+    long long *edgeStamps;
```
john-boyer-phd: This is a very  cool way of doing this, and works. Nonetheless, because of the facts on which you rely, I'd like to ask if you could please remove it in favor of using the vertex visited flags. 

At the start of reading a graph, all vertex visited flags are clear. During processing of v, the visited flag of each x can be checked. If already set, then parallel edge error. If clear, then set it and add the edge. At the moments immediately before when v is increased, add a little routine that goes through the adjacency list and clears the neighbors' visited flags. Also has to be done for the last vertex when the end of the `:` line is reached, such that the visited vertices of the graph are all back to being clear.

john-boyer-phd (approved):
This is a fantastic piece of work, Ian. I can't see any changes. I _really_ like your `edgeStamps` idea. I was going to suggest that it could be replaced with a simple technique involving the vertex visited flags, but your way is a bit faster at the present time, and the reality is that it will almost certainly be replaced with a faster method that also works for incremental graphs, only once parallel edge support is increased by the resolution of issue #253. I also really liked your thoroughness in testing and creativity being thorough, too, such as your Lockstep test.

PR #332, opened 2026-09-07
Add CITATION.cff

goelakash:
Adds a `CITATION.cff` so GitHub shows a "Cite this repository" button and citation tooling can read machine-readable metadata.

Author names and ORCIDs come from `paper.md`, and the licence from the LICENSE file. Validated with `cffconvert --validate` against schema 1.2.0.

Worth checking the given-name/family-name split per author, since some papers give a single `name` field.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,12 @@
+cff-version: 1.2.0
+message: If you use this software, please cite it as below.
+title: Edge Addition Planarity Suite and Generalized Graph Library
+type: software
+authors:
+- given-names: John M.
+  family-names: Boyer
+  orcid: https://orcid.org/0000-0002-4755-5535
+- given-names: Wanda B. K.
+  family-names: Boyer
+repository-code: https://github.com/graph-algorithms/edge-addition-planarity-suite
+license: BSD-3-Clause


john-boyer-phd:
This is a good idea, Akash, thank you. The reason it is a good idea is that my hope with the JOSS paper was that it would give authors a recent source to cite when using this software. Based on this reason, I'm going to briefly hold this PR in the back pocket, so to speak. We're on the verge of doing a 5.1.0.0 release, so that the JOSS paper can have a new archival tag that includes the improvements made during the review. Assuming and once the paper is published, then I'll come back here with a PR review to ask you to slightly amend the `cff` file with the JOSS paper info (volume number, doi, etc.).

My hope is that authors will cite the JOSS paper whether they use EAPS in C (including via `planarity.exe`) or in Python via the `planarity` PyPI package. Therefore, once the final `cff` file is made here, a PR would be most welcome to contribute the same `cff` file on the [Planarity Python repo](https://github.com/graph-algorithms/planarity) ;-).

john-boyer-phd (approved):
Thanks for getting us started on this! If you'd like to do a PR over on the planarity repo with this cff change and the DOI markdown for the readme, it would be gratefully received.

PR #333, opened 2026-09-07
Run distcheck for manually dispatched test workflows

### LINKED ISSUE #330: Fix skipped distcheck in runtests.yml
Please investigate and fix the reason why the `distcheck` job in `runtesets.yml` is reporting being skipped.

WilliamK112:
Resolves #330

## Type of change

### Updated

- `.github/workflows/runtests.yml` — remove the stale pull-request-only condition from the `distcheck` job. The workflow currently enables only `workflow_dispatch`, so the old condition made the job unconditionally skip on every available trigger.

## Testing

- Confirmed [Tests run 34053173815](https://github.com/graph-algorithms/edge-addition-planarity-suite/actions/runs/34053173815) used `workflow_dispatch` and reported `distcheck` as skipped with no steps.
- Parsed `.github/workflows/runtests.yml` successfully with Ruby Psych.
- Confirmed the `distcheck` job no longer contains an `if` event guard.
- Ran the exact job commands in an isolated source archive: `mkdir -p m4`, `autoreconf -if`, `./configure CC=gcc`, and `make distcheck -j4`; all completed successfully and produced the `planarity-5.1.0.0.tar.gz` distribution archive.
- `git diff --check`

The pull-request trigger remains disabled, preserving the repository's current Actions-minutes policy; this only makes `distcheck` participate in manually dispatched Tests runs.

### DIFF
--- a/.github/workflows/runtests.yml
+++ b/.github/workflows/runtests.yml
@@ -39,7 +39,6 @@ jobs:
   distcheck:
     timeout-minutes: 15
     runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'pull_request' }}
     env:
       CC: gcc
     steps:


john-boyer-phd (approved):
Thank you, William!
I find your Testing to be very consistently excellent, so thank you for that too ;-)

As an aside, I recently found out that apparently the 2000-minute limit may only apply to private repos. Still, it's nice not having to wait for compiler-versions and especially dev-platforms, so we can continue to run those now and again or when needed.

PR #340, opened 2026-09-19
Keep sf_fgets() inside the caller's buffer when reading pushback

### LINKED ISSUE #339: sf_fgets() writes one byte past the caller's buffer when reading from the pushback buffer
## Describe the bug

`sf_fgets()` stores up to `count` characters from the pushback buffer and writes the null terminator after the last of them, so a pushback buffer holding at least `count` characters makes it write `str[count]` (`strOrFile.c:775`, the loop that begins at `:766`). Its file path stores at most `count - 1` characters, as stdio does, so the two paths of the same function disagree about the bound.

`gp_Read()` reaches it. `_ReadGraph()` reads the first line of the input and pushes it back for the reader that the line selects (`graphIO.c:517` and `:522`), and the graph6 reader's line buffer is sized from the order in the first character, nine bytes for order 8 (`g6-read-iterator.c:354`). A `.g6` file whose first line is longer than its order allows fills the pushback buffer past that, and the next `sf_fgets()` writes one byte beyond it.

## Reproduction steps

With an ASan build (`./configure CC=clang CFLAGS="-fsanitize=address -fno-omit-frame-pointer -g -O1"`):

```bash
python3 -c "open('overlong.g6','w').write('G' + '?'*300 + chr(10) + 'G?????' + chr(10))"
planarity -s -q -p overlong.g6 out.txt
```

## Expected behavior

The line is malformed, so it should be rejected the way the same file is rejected when it is read without the pushback, i.e. through `planarity -t`: "Invalid line length read on line 1".

## Terminal logs

```
==20355==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000000199
WRITE of size 1 at 0x602000000199 thread T0
    #0 sf_fgets strOrFile.c:775
    #1 g6_ReadGraph g6-read-iterator.c:534
    #2 _g6_ReadGraphFromStrOrFile g6-read-iterator.c:761
    #3 _ReadGraph graphIO.c:558
    #4 gp_Read graphIO.c:461
    #5 SpecificGraph planaritySpecificGraph.c:112

0x602000000199 is located 0 bytes after 9-byte region [0x602000000190,0x602000000199)
allocated by thread T0 here:
    #0 calloc
    #1 _g6_InitReader g6-read-iterator.c:355
```

## Environment

macOS 26 on Apple silicon, Apple clang 21, `planarity` built from `911af7a`. The same input is clean on the file path, so nothing here depends on the platform beyond having a sanitizer to see the one byte.

The bound is one expression, and the read-error tests are the place for the case: a pushback buffer longer than the caller's buffer, a guard byte after the requested count, the terminator inside the buffer. I have that on a branch and will open the PR unless you would rather take it yourself.

One more thing from the same reading, which I have deliberately left out of that branch because it is a different defect: the string-backed path uses `strncpy(str + charsToReadFromUngetBuf, theStrBuf, charsToReadFromStrOrFile)` (`strOrFile.c:808`), which copies exactly that many bytes and adds no terminator, so a caller whose buffer is `count` bytes can be handed an unterminated string.

Deln0r:
Fixes #339 by giving the pushback path of `sf_fgets()` the bound its file path already has.

## Type of change

### Updated

* `c/graphLib/io/strOrFile.c`: the pushback loop reads at most `count - 1` characters, so the terminator it writes after them stays inside a buffer of `count` bytes. The remainder taken from the file or the string buffer is computed against that same cap, so the total written is unchanged for every input that did not overflow. The function comment claimed stdio parameter order while documenting a `count + 1` allocation; it now states the stdio bound the function keeps on all three paths.
* `c/planarityApp/planarity-test/testReadErrors.c`: a case for the boundary. It pushes a string longer than the caller's buffer, fills the buffer with a guard character, asks for `sizeof(buffer) - 1`, then checks that the guard byte after the requested count is untouched, that the terminator is inside, and that the characters returned are the leading ones pushed back.

The observable change: a `.g6` file whose first line is longer than its order allows is now rejected with "Invalid line length read on line 1" instead of writing one byte past the graph6 reader's line buffer. Well-formed input is unaffected, since the pushback buffer is then shorter than the buffer the reader offers.

## Testing

- ASan build, the reproducer from #339: clean with this change, heap-buffer-overflow without it. `planarity -test` under ASan is clean too.
- The new case fails with the bound removed (3 of 143 read-error checks) and passes with it, so it is the bug it claims to cover.
- Containers on Ubuntu 24.04: gcc-14 and clang-18 under `--enable-compile-warnings=error`, gcc-14 with `--enable-unsigned-char`, plain gcc and clang, and `make distcheck`. Ubuntu 26.04: clang with `-fsanitize=address` and with `-fsanitize=undefined -fno-sanitize-merge`. All clean, suite green in each.
- macOS Apple clang 21: build clean under the error gate, `make check` and `planarity -test` pass.


john-boyer-phd (approved):
Great catch, Ian, thank you.
Based on also having the other error with `strncpy()`, it seemed like an appropriate time to pause a beat and improve the code with better variable names, more comments, simplified steps, and of course changing to `strncat()` to get the null terminator.

PR #341, opened 2026-09-19
Test algorithms on sparse6 and incremental sparse6 input

Deln0r:
Third step of #38, after the read iterator in #329: `planarity -t` runs its algorithms over sparse6 and incremental sparse6 files, not only `.g6`.

Stacked on #340, which this needs. Reading the first line and pushing it back is what `gp_Read()` already does, and it is what exposes the overflow fixed there, so without that bound the same input would reach it through `-t` as well.

## Type of change

### Added

* `c/samples/n8.mALL.inc.s6`: the existing `n8.mALL.g6` sample in incremental sparse6, made with `copyg -i`. `copyg -g` maps it back to the `.g6` file byte for byte.

### Updated

* `c/graphLib/io/s6-read-iterator.[ch]`: the first-line test that `_ReadGraph()` applies is now `s6_IsSparse6Input()`, so its two callers cannot drift apart.
* `c/graphLib/io/graphIO.c`: `_ReadGraph()` calls it instead of repeating the condition.
* `c/planarityApp/planarityTestAllGraphs.c`: a facade over the two read iterators, chosen by the content of the input rather than by the name of the file. The input is opened once: the first line is pushed back into the input container and the container is handed to the iterator that matches it, so an input that cannot be reopened keeps its first graph. That is why the facade calls `_g6_InitReaderWithStrOrFile()` and `_s6_InitReaderWithStrOrFile()`, which are package private today, and it is the one decision I would like from you: promote those two to the public headers, or move the facade into the io layer next to the iterators. I will do whichever you prefer.
* `c/planarityApp/planarityCommandLine.c`: `runTestAllGraphsTest()` takes the expected result line, with `NULL` keeping the counts for `n8.mALL.g6`. Six commands run over the new sample, and planarity runs over `N5-all.s6` for the plain sparse6 case.
* `c/planarityApp/planarityHelp.c`, `planarityMenu.c`, `planarity.1`: the formats `-t` accepts.

## Testing

- The six algorithms over `n8.mALL.inc.s6` produce the counts the `.g6` file produces, drawing included, which matters because it requires edge storage without holes and the reader fills them.
- With the format detection disabled, those six fail, so the tests do exercise the new path.
- Thirteen inputs by hand, each compared with master where master supported it: the three formats as a file, as `stdin`, through a redirect of `/dev/fd/0` and behind a pipe, an empty file, a missing file, and the 12,346-graph sample. The first version of this opened the file a second time to look at it, which cost the first graph of `/dev/fd/0`, 33 graphs where master reads 34; the container version reads every case correctly.
- Measurement for the epic, on the 12,005,168 graphs of order 10 from `geng -q 10`, three interleaved rounds each:

| input | size | `-t -p` user time | result line |
|---|---|---|---|
| `n10.g6` | 120.1 MB | 22.92, 23.26, 23.29 s | `12005168 1140916 10864252` |
| `n10.inc.s6` | 65.3 MB | 23.47, 23.22, 23.16 s | `12005168 1140916 10864252` |

  The incremental file is 46 percent smaller and reads faster on its own, 1.84, 1.83, 1.81 s against 2.11, 2.15, 2.08 s through `copyg -g` into `/dev/null`, but the run takes the same time either way: `gp_CopyGraph()` and `gp_Embed()` dominate and the read is a few percent of it. I would rather write that in the epic than imply a speedup that is not there.
- Containers on Ubuntu 24.04: gcc-14 and clang-18 under `--enable-compile-warnings=error`, gcc-14 with `--enable-unsigned-char`, plain gcc and clang, and `make distcheck`. Ubuntu 26.04: clang with ASan and with UBSan. macOS Apple clang 21: error gate, `make check`, `planarity -test`. All clean.


john-boyer-phd (changes requested):
This is a great PR, Ian. You're saving me the trouble of having to file this issue later. As soon as I saw the s6 API, I realized I had made a mistake in making the g6 API public. Not until looking at the first line does the API user doesn't know whether they have a G6 file, an S6 file, or eventually a GraphML file or GML file  containing multiple graphs. By then, it's too late. 

I'd like to ask if you could please refactor this code to "promote" it into the graphLib? 

If this is OK to you, then please consider the following specific points:
 * Add `gp-read-iterator.c/.h` to the `io` subdirectory to contain your "allGraphsReader" code currently in `planarityTestAllGraphs.c`. 
 * Rename functions and structures as necessary to exactly mirror the `g6_`/`s6_` APIs except starting with `gp_` instead (instead of `allGraphs...`). 
 * The only changes to `testAllGraphs()` should be to change `g6_` to `gp_` (and to retain your more generalized comment about initialization). 
 * It seems best to have the matching function `g6_IsGraph6Input()` method that the allGraphs `gp_`-level initializer could call to decide whether to create a G6 read iterator. (And `_ReadGraph()` could call that too).
 * The iterator structure should probably have a data member and set of file type macros that identify G6 and S6 for now. When the initializer decides the file type, it could set that variable and then make decisions from it, rather than based on whether there is an S6 read iterator.

The huge advantage to promoting this code to the `gp_` level in `graphLib` is that it will make it a lot easier to also amend the Python version of test all graphs in the EAPS-T repo to use the generalized iterator and get support for iterating with G6, S6, and any other supported format in the future.  

This sound OK?

Deln0r:
Yes, and it answers the question I put in the PR, so I will refactor it that way. All six names are free in the tree today.

One point on `g6_IsGraph6Input()`: graph6 is the fallback in `_ReadGraph()` today, so a positive test changes what happens to input that matches nothing. In the `gp_` initializer I would test sparse6, then graph6, then report an unrecognized format with the first bytes of the input, rather than have the graph6 reader report an invalid first character for, say, a GraphML file.

The file type member also lets read, end-reached and free stop asking which iterator exists.

Deln0r:
Done, and it reads better in the library than it did in the application. Pushed as two commits on this branch.

The iterator is `c/graphLib/io/gp-read-iterator.[ch]`: `gp_NewReader()`, `gp_InitReaderWithString()`, `gp_InitReaderWithFileName()`, `gp_ReadGraph()`, `gp_EndReached()`, `gp_FreeReader()`, with `GPReadIteratorP` and the `GP_FILE_TYPE_` macros. The type is decided once, at initialization, and every other method dispatches on it, so a third format is a case in three switches and one predicate. `testAllGraphs()` keeps its loop and changed `g6_` to `gp_`, as you asked.

`g6_IsGraph6Input()` sits beside `s6_IsSparse6Input()`. Since graph6 marks its lines with nothing but the order, it recognizes a candidate: the header, or a first byte in 63 to 126. `_ReadGraph()` asks it instead of falling through, and an input in none of the formats now gets its own message there rather than the graph6 reader's complaint about a first character.

Two notes on what that predicate cannot do, both in the tests so they stay visible. An adjacency list begins with `N`, which is a legal order byte, so a file of graphs starting that way is taken for graph6 and fails when the line is read; in `gp_Read()` nothing changes, because the `N=` branch comes first. And `>>graph6<<` is the only positive marker graph6 has.

A file is still opened once: the first line is read to decide the type and pushed back, and the same container is handed to the chosen reader, so stdin and pipes keep their first graph. A string is classified directly, since looking at it consumes nothing.

While writing the tests I found that `gp_NewReader()` would allocate over an existing iterator, where `g6_NewReader()` and `s6_NewReader()` refuse it; that is the second commit, with the case that catches it.

`runGeneralReadIteratorTests()` covers the two predicates over a table of first lines, both formats from a string and from a file, reading each sample to its last graph, an input refused for having no format, one that looks like graph6 and fails when read, a refused second initialization after which the iterator still reads, and a refused allocation over an existing reader.

Checked by hand again after the move: the three formats as a file, as `stdin`, through a redirect of `/dev/fd/0` and behind a pipe, plus digraph6, GraphML, an empty file and a missing file. Local gate green on all eight legs: gcc-14 and clang-18 under `--enable-compile-warnings=error`, gcc-14 with `--enable-unsigned-char`, plain gcc and clang, `make distcheck`, and clang with ASan and with UBSan on 26.04.

The EAPS-T side is yours to decide, but the Python test-all can now mirror this one for its own formats.

c/graphLib/io/g6-read-iterator.c
```diff
@@ -63,6 +63,40 @@ struct G6ReadIteratorStruct
...
+
+    if (strncmp(firstLine, ">>graph6<<", strlen(">>graph6<<")) == 0)
+        return TRUE;
+
+    firstChar = (unsigned char)firstLine[0];
+
+    if (firstChar == ':' || firstChar == ';' || firstChar == '&')
+        return FALSE;
+
+    return (firstChar >= 63 && firstChar <= 126) ? TRUE : FALSE;
```
john-boyer-phd: It would be possible to enable this function to reject the adjacency list format in a generic way. At this point, you could say:
```
    if (firstChar >= 63 && firstChar <= 126)
    {
        int secondChar = (unsigned char)firstLine[1];
        return (secondChar >= 63 && secondChar <= 126) ? TRUE : FALSE;
    }
    return FALSE;
```
This would work because the adjacency list format starts with `N=` and the equal sign is not in range. I would have just made the change, but I suspect you may want to pair this with adding another test?

c/graphLib/io/gp-read-iterator.c
```diff
@@ -0,0 +1,323 @@
...
+    // type is decided from it directly and the read iterator of that type is
+    // then initialized with the whole string.
+    if (_gp_SetFileType(theGPReadIterator, inputString) != OK ||
+        _gp_NewChildReader(theGPReadIterator) != OK)
+    {
+        _gp_FreeChildReader(theGPReadIterator);
+        return NOTOK;
+    }
+
+    if ((theGPReadIterator->fileType == GP_FILE_TYPE_S6
```
john-boyer-phd: Would you mind flipping this around a bit to not use the ternary operator? In other words to say if the file type is `GP_FILE_TYPE_G6` then try to `g6_InitReaderWithString()` and handle failure _else_ if file type is `GP_FILE_TYPE_
S6` then try to `s6_InitReaderWithString()` and handle failure, _else_ fail. This rearrangement will make it easier to add other else clauses for formats like GraphML in the future. The `switch` statements you're using everywhere else might be the best way to implement this for consistency.

c/graphLib/io/gp-read-iterator.c
```diff
@@ -0,0 +1,323 @@
...
+
+    if (_gp_SetFileType(theGPReadIterator, lineBuff) != OK ||
+        _gp_NewChildReader(theGPReadIterator) != OK)
+    {
+        _gp_FreeChildReader(theGPReadIterator);
+        sf_Free(&inputContainer);
+        return NOTOK;
+    }
+
+    Result = (theGPReadIterator->fileType == GP_FILE_TYPE_S6)
```
john-boyer-phd: And also here, using a `switch` instead.

c/planarityApp/planarityMenu.c
```diff
@@ -62,7 +62,7 @@ int menu(void)
 
             gp_Message(
                 "X. Transform single graph in supported file to .g6, adjacency list, or adjacency matrix\n"
-                "T. Perform an algorithm test on all graphs in .g6 input file\n"
+                "T. Perform an algorithm test on all graphs in .g6, .s6 or .inc.s6 input file\n"
```
john-boyer-phd: It would be better here to just say "Perform an algorithm test on all graphs in an input file\n".

I didn't want to "muddy the waters" and take the focus off of sparse6, but there's nothing really to stop us from having the `gp_` level iterator recognize adjacency list, adjacency matrix, etc. and just iterate over the one graph in the file. That generalization could be filed as a separate issue, though.

Edit: looks like the string above it for the `X` command will need help too because the addition of sparse6 will make the current wording become even more unwieldy. How about "X. Transform a single graph in an input file to a supported output format."

c/graphLib/io/gp-read-iterator.c
```diff
@@ -0,0 +1,323 @@
...
+/********************************************************************
+ _gp_SetFileType()
+
+ Decides the format from the first line of the input and stores it,
+ so that the other methods dispatch on the type rather than on which
+ read iterator happens to exist.
+ ********************************************************************/
+int _gp_SetFileType(GPReadIteratorP theGPReadIterator, char const *const firstLine)
+{
+    if (s6_IsSparse6Input(firstLine))
```
john-boyer-phd: I think this is the only place where s6 is done before g6 (well, it will be after you turn two ternary operator constructs into switches below). Just for consistency, would it be ok to flip the order here?

john-boyer-phd (changes requested):
This new formulation feels _sooo_ good, Ian, thank you! Just a few really minor changes.

Also, when it comes time to add the `s6_RetrieveGraphChange()` method, it can also be added at the `gp_` level you have just created, because G6 and other formats can just behave the same way as non-incremental sparse6 will have to behave. It'll just be no increment available so they call `gp_ReadGraph()` to read a whole graph. Only sparse6 will behave differently and only if/when incremental sparse6 content is encountered.

c/graphLib/io/g6-read-iterator.c
Deln0r: Done, with one allowance. Your check would also reject a line that is only its order byte, graph6 order 0 or 1, because the byte after it is the newline; the existing predicate case pins "?" as graph6, so NUL, LF and CR pass as the second byte too. "N=3" joined the table, and the adjacency list moved from the read-failure cases to the reject cases.

Deln0r:
All five are in SHA. The second byte now has to be in the range too, with the one allowance explained on that thread, so the adjacency list is refused at the predicate: it moved from the read-failure cases to the reject cases, "N=3" joined the predicate table, and the read-failure case is now a truncated graph6 line. graph6 is tested before sparse6 everywhere, both initializers dispatch with a switch, and the menu says what you suggested.

Agreed on the lookahead at the gp_ level: every format but sparse6 reports no increment, and the caller reads a whole graph.

john-boyer-phd (approved):
An excellent piece of work, Ian. Thanks again for saving the work of having to file this later, and good catch on the order 0 and 1 graph6 files.

PR #343, opened 2026-09-23
Add GraphML reader Part 1

### LINKED ISSUE #342: GraphML Reader Part 1
GraphML is an XML markup language capable of expressing many aspects and attributes of graphs, their vertices (nodes) and their edges. It is a reasonable to respond to the functional requirement to read GraphML files by adopting an open source XML parser and then processing its output, such as a document object model (DOM) containing GraphML elements and attributes. However, there are a few reasons why we will instead be creating our own GraphML-specific parser. First, we don't want to have to fight with how much potential modification may be needed to compile an existing XML parser at the level of warning stringency we have in this project. Second, and more important, XML parsers have notoriously long histories of required security patches and, sometimes, delayed response to patching; we don't want to be on a treadmill of security patches, nor on the possibly delayed security patch schedule. Furthermore, by writing our own parser that is specific to GraphML, our solution will be less susceptible to security issues than a generalized XML parser and because our parser can stop immediately on seeing just a few bytes of XML content that is not expected in a GraphML file.

The GraphML Reader is also to be based on the `strOrFile` input container, like all of our other readers, so that it can seamlessly read from a string or `FILE`. Our parser will also be able to take advantage of the large unget buffer that the `strOrFile` object supports, which will help with looking ahead to see that we are getting only the content we support.

## Technical Overview of GraphML Support in Part 1

This section identifies broadly what is and is not supported in this GraphML Reader Part 1.

### Supported

* Optional byte-order mark (BOM).

* ISO 8859-1 and UTF-8 that is convertible to ISO 8859-1.

* Optional XML Declaration; if present, the required `version` attribute must have a value of `1.0`, the optional `encoding` attribute if given must indicate UTF-8 or ISO-8859-1), and the optional `standalone` attribute if given must have the attribute value `yes`.

* XML comments and whitespace characters, which are skipped and not retained, after the XML Declaration and before the root document element's start tag, before any other supported GraphML element's start tag, and after any supported GraphML element's end tag.

* XML whitespace characters, which are skipped and not retained, before any attribute name, on either side of the equal sign separating an attribute name and value, and after any attribute value. This applies to attributes in the XML declaration and in XML element start tags.

* In character data for attribute values and elements whose content is character data, the following character entity references only are supported (converted to the byte they represent): &lt;, &amp;, &gt;, &apos;, and &quot;. It is an error if a `<` character appears in the raw data or if `&` appears without being followed by one of these five character entity references.

* In character data, if the content encoding is UTF-8, then any two-byte UTF-8 characters that can be converted to ISO-8859-1 are so converted, and otherwise an error occurs for an unsupported UTF-8 character. Specifically, UTF-8 encoded bytes up to 0x7F are unchanged in ISO-8859-1, a UTF-8 encoded byte of 0xC2 if followed by a byte between 0xA0 and 0xBF is converted to a single ISO-8859-1 byte between 0xA0 and 0xBF, and a UTF-8 encoded byte 0xC3 followed by a byte between 0x80 and 0xBF is convered to a single ISO-8859-1 byte between 0xC0 and 0xFF.

* In general, the GraphML Reader supports various GraphML elements and attributes (e.g., `<graphml>`, `<key>`, `<graph>`, `<node>`, `<edge>`, `id`, `source`, `target`, etc.) only when declared without namespace qualification. As will be evident below, the default namespace must be set to the GraphML URI in the root document element and not changed in any descendant element.

* For any attribute with the name `id`, the attribute value's leading and trailing whitespace are removed (and not retained), and the characters are checked to ensure they are only capital and lower case letters, digits, the underscore, the hyphen, the period, or the colon. Further checks are performed per GraphML element, as described below.

* The `graphml` element is supported as root document element only, with no namespace qualification, and with a required `xmlns` attribute set to the GraphML URI. The `xsi` namespace and the `xsi:schemaLocation` attributes may optionally appear in the `<graphml>` start tag only, and they are skipped if present. No other attributes are supported in the `<graphml>` start tag.

* The only allowed child content of the `<graphml>` element (other than whitespace and comments) is a sequence of zero or more `<key>` elements followed by a sequence of one or more `<graph>` elements. when the GraphML Reader is configured to read only one graph, any subsequent content, such as other `<graph>` elements is skipped (and therefore not checked for well-formedness, etc.). The Part 1 GraphML Reader is configured to read only one graph from the content.

* A `<key>` element must have the following attributes, in any order: `id`, `for`, `attr.name`, and `attr.type`. The `id` and `attr.name` attribute values are normalized and then tested using the rules for `id` attributes above. The `attr.type` value must be one of `boolean`, `int`, `long`, `float`, `double`, and `string`. The subset of allowed values for the `for` attribute value that should be tested are `graph`, `node`, and `edge`. A `<key>` element can be expressed as an empty tag (the start tag ends with `/>` after the attributes), or it can have a single child element called `<default>` (plus optional leading and trailing whitespace and/or comments). The `<default>` element content can only be character data (which includes resolving the supported character entity references and converting UTF-8 characters to ISO 8859-1).

* In this GraphML Reader Part 1, there can be only one `<key>` element (or none), and if it is given, it must be for the `graphflags_zerobasedio` graph-level attribute. So, after recognizing a valid `<key>` element, test for the specific values of this one key, produce an error for a non-match, and then produce an error if the start of a second key element is detected.

* A `<graph>` element can have the following attributes in any order: `id`, `edgedefault` (values of `undirected` or `directed` only), `parse.nodes` (positive integer only), `parse.edges` (positive integer only), `parse.nodeids` (`canonical` only), `parse.edgeids` (`canonical` only), and parse.order (`nodesfirst` only). The `id` is optional; if given, then in this GraphML Reader Part 1, it is an error for the value to be `graphflags_zerobasedio` (conflicts with the `<key>` id), or to have the format of a canonical `<node>` or `<edge>` id (to ensure `id` uniqueness). The `edgedefault` is optional and defaults to `undirected`. The `parse.nodes` and `parse.edges` are optional and default to how many child `<node>` and `<edge>` elements, respectively, that the `<graph>` contains. In this Graph ML Reader, it is an errors for `parse.nodes` to be less than 2 and for `parse.edges` to be less than 1. In this GraphML Reader Part 1, `parse.nodeids`, `parse.edgeids`, and `parse.order` are required and must have the given attribute values (because they have different effective defaults in GraphML).   

* The allowed children of the `<graph>` element (other than whitespace and comments) are: zero or more `<data>` elements, then a sequence of `<node>` elements, then a sequence of `<edge>` elements. In this Graph ML Reader, it is an error to have less than 2 `<node>` child elements or to have less than 1 `<edge>` elements, and it is an error for the number of `<node>` and `<edge>` elements to differ from the values of `parse.nodes` and `parse.edges`, respectively, if those were expressed. In this GraphML Reader Part 1, all `<node>` elements must precede all `edge` elements because only `nodesfirst` parse order is supported at this time. 

* In this GraphML Reader Part 1, the only `data` element that can appear is the one that exactly expresses the `graphflags_zerobasedio`. It is an error for the data value to be other than `true`, `false`, `1`, or `0`. The graph's zero-based I/O flag should be set if the value is `true` or `1`, and it should be explictly cleared otherwise.

* In this GraphML Reader Part 1, each `<node>` element must be expressed as an empty tag or with only optional whitespace and comments between the start and end tags. It must have an `id` attribute and no other attributes. The `id` attribute value must be canonical, i.e., start with "n" followed by a string expressing the 0-based position of the `<node>` element in the list of `<node>` element children of the parent `<graph>` element (e.g., "n0", "n1", "n2", ...).

* In this GraphML Reader Part 1, each `<edge>` element must be expressed as an empty tag or with only optional whitespace and comments between the start and end tags. The `<edge>` element can have the following four attributes only: `id`, `source`, `target`, and `directed`. The `id` attribute optional; if it is expressed, then the attribute value must be canonical, i.e., start with "e" followed by a string expressing the 0-based position of the `<edge>` element in the list of `<edge>` element children of the parent `<graph>` element (e.g., "e0", "e1", "e2", ...). The `source` and `target` attributes are required. They must contain canonical node id references, which must refer to actual nodes of the graph, i.e., the integer after the letter "n" must be in the range 0 to N-`, where N is the number of vertices (nodes) in the graph. 

* Once the last `<edge>` element has been processed, any whitespace and comments must be skipped to enable the processor to recognize the `</graph>` end tag and advance to the byte after it (or produce an error on failure). 
 
* After the GraphML reader has finished reading the one or many `<graph>` elements, the `</graphml>` end tag and trailing XML content are skipped (and therefore not checked for well-formedness, etc.).

### Not Supported

A consuming application should remove or replace the following content items, or note that they will produce errors from the GraphML Reader Part 1:
 
* Encodings other than those supported, e.g., UTF-16, EBCDIC, UTF-8 characters not convertible to ISO 8859-1, other encodings declared in the XML declaration.
* XML DTDs (and hence no entity declarations or expansions).
* XML processing instructions.
* XML namespace declarations, except the default namespace set equal to the GraphML URI and the appropriately assigned `xmlns:xsi` attribute.
* XML namespace usage in element start tags (no colons).
* XML namespace usage in attribute names, except the `xsi:schemaLocation` attribute (which is ignored).  
* CDATA sections.
* Entity references other than the supported character entity references.
* Attributes in a start tag other than the supported attributes for the given element.
* Any `id` attribute value characters other than the supported characters.
* Non-canonical id values for nodes and edges (which will be changed to supported in a near-term future version).
* Parse orders other than `nodesfirst`, such as `free` or `adjacencylist` order.
* Any `<key>` and `<data>` elements other than those supported. 
* Child elements or character content other than the supported child elements or character content per supported GraphML element. 

In particular, this means no nested graphs, no `<port>` elements, no `<hyperedge>` elements, and no foreign-namespaced content, such as content in the `svg` namespace.

The following will be changed to supported in a near-term future version of this GraphML Reader:

* Arbitrary `<key>` elements, and more `<data>` element children of `<node>`, `<edge>`, and `<graph>`.
* The `adjacencylist` parse order.
* Non-canonical id values for nodes and edges.

 
## Location of and Routing to GraphML Reader

In `graphIO.c`, the method `_ReadGraph()` routes to different readers based on reading (and then ungetting) a line up to MAXLINE (currently 1024) characters, and then using the first few characters to decide on the reader. Just before the `else` clause that routes to the Graph6 reader, add a new `else if` that routes to a new method `_ReadGraphMLGraph()` if the first few bytes indicate that the input is XML content. If the first character of the line matches one of the following characters, then route to GraphML reader: the less-than symbol ('<'), a space (0x20), a tab (0x08), a return (0x0D), a new line (0x0A), the start of a UTF-8 byte order mark (0xEF), the start of a UTF-16 byte order mark (0xFE or 0xFF), or a null terminator (0x00). Otherwise, also test for the following three-byte sequence: 0x4C 0x6F 0xA7, and route to the GraphML reader if found because it is likely EBCDIC-encoded XML and definiitely not Graph6 (because 0xA7 is illegal in Graph6 files). 

The method `_ReadGraphMLGraph()` and the new subroutines it calls should be placed in a new file `graphML-reader.c` in the `io` subdirectory. The name of the file should be added to `makefile.am`, just above the entry for `graphML-writer.c`. 

The function signature should be `int _ReadGraphMLGraph(graphP theGraph, strOrFileP inputContainer, int readGraphElemOnly)`.  This function signature must also be added to the top of `graphIO.c`, between the declarations for `_ReadLEDAGraph()` and `_ReadPostprocess()`.

The invocation of `_ReadGraphMLGraph()` in `_ReadGraph()` should pass `FALSE` to `readGraphElemOnly` because it will be necessary to read and process all the markup from the start of the content, not just a `<graph>` element at or immediately after the `inputContainer`'s current content pointer location.  The `TRUE` setting is for future use of a GraphML read iterator.

 
## The `_ReadGraphMLGraph()` Method (and Subroutines)

The `_ReadGraphMLGraph()` method should return `OK` on success or `NOTOK` on any error. Invalid parameter errors include receiving `NULL` for `theGraph` or an `inputContainer` that fails `sf_IsValidStrOrFile()`. Since these parameters are being checked at this top-level within `graphML-reader.c`, they don't need to be rechecked in subroutines within this file. Any other errors in `_ReadGraphMLGraph()` will come from subroutines returning a value `!= OK`.

**NOTE:** All subroutines added to `graphML-reader.c` except for `_ReadGraphMLGraph()` should be declared `static`.

**NOTE:** All subroutines in `graphML-reader.c` should return an `int` that is either `OK` on success or `NOTOK` on failure. Each failure should be coupled with an approriate error message delivered with `gp_ErrorMessage()`. 

**NOTE:** All subroutines in `graphML-reader.c` should have at least the following parameters in their signature without this specification needing to say so: `(graphP theGraph, strOrFileP inputContainer, int *pLineNum)`. Most will not need more, but it will become evident that some methods may need an extra subsequent parameter or so. 

In `_ReadGraphMLGraph()`, declare a variable `lineNum` initialized to 1. The address of `lineNum` will be passed around to all subroutines so that it can be used in error messages to indicate what line the reader is on when an error occurs and so that it can be incremented when a linefeed (0x0A) is consumed, which means not just read for lookahead but read and then not pushed into the unget buffer. Such linefeed increments will occur when skipping whitespace (such as between attribute values or between element tags), skipping comments (such as within the comments), or reading character data (such as attribute values or element content).

If `readGraphElemOnly` is `FALSE`, then the next lines of code need to invoke the following subroutines (and check for `OK`/`NOTOK` results): 

* `_ReadGraphMLProlog()`
* `_ReadGraphMLStartTag()`
* `_ReadGraphMLKeys()`

Then, regardless of `readGraphElemOnly` parameter value, invoke `_ReadGraphMLGraphElement()`.

Finally, note that it's preferable not to proceed with trying to read and skip whitespace, comments, and the `</graphml>` end tag (if `readGraphElemOnly` is `FALSE`) because it adds no value and because it would also be necessary to 1) read and skip any more `<graph>` elements that may be before the `</graphml>` end tag, 2) read and skip any whitespace and comments after the `</graphml>` end tag, and 3) read and produce errors for any processing instructions that may occur after the `</graphml>` end tag.

  
## Lookahead Strategy

Throughout the implementations of subroutines, it will only be necessary to look at most five characters ahead to decide whether to call a subroutine, directly consume some specific expected content, or produce an error indicating that unexpected content has been found (on a particular line number of the GraphML string or file). 

For example, if one gets a first lookahead character that is a whitespace character, and whitespace is allowed at the current point in parsing (such as before reading an element start tag), then you would unget the character and then call a method to `_ReadGraphMLSkipWhitespace()`. That method would then read zero or more whitespace characters until it read a non-whitespace character (or EOF), and it would then unget the non-whitespace character just read and return `OK` to the caller. 

After skipping whitespace, it's often the case that the next character read is a less-than symbol (0x3C), such as the beginning of an element start to end tag, or the beginning of an XML declaration, an XML DTD, a processing instruction, a CDATA section, or a comment. A second lookahead character helps to distinguish between these cases. For example, if it is an exclamation point ('!'), then the upcoming content may be a comment, or it may be unsupported content such as an XML DTD or a CDATA section. One would check for `<!--` in total; a match means unget the four characters and go to a `_SkipGraphMLComment()` subroutine that can get past the comment. If four lookahead characters don't match `<!--` (including getting an unexpected EOF), then it's time to emit an "Unsupported GraphML content found on line %d" error message and return `NOTOK`. 

The character after the less-than symbol (0x3C) may instead be a question mark. If it's the first line of the content, and five lookahead characters match `<?xml`, then you have an XML declaration, so you would unget the five characters and invoke `_ReadGraphMLXMLDecl()`. Otherwise, if it's not the first line of the content or if the next characters after `<?` aren't exactly `xml`, then it's time to emit an unsupported content error message and return `NOTOK`.

The character after the less-than symbol (0x3C) may be what the XML specification calls a `NameStartChar`. However, within GraphML, the parser will typically be looking for only one or two possible XML start tags. For example, after processing the `<graph>` element start tag, the only acceptable elements in `nodesfirst` GraphML content (after skipping whitespace and comments) are `<data>` or `<node>` (if there are no graph-level `<data>` elements). So, if the character after the less-than symbol is neither an exclamation mark nor a question mark, then it can only be a `d` or an `n`. Otherwise, it's time to emit an "Unsupported element or content found on line %d" error and return `NOTOK`.

Finally, the character after the less-than symbol (0x3C) maybe a forward slash ('/'), which can only occur if the parser is at the beginning of an end tag, such as `</graph>`. The code in the specific parsing subroutine will have to decide whether it is expecting an end tag at all, and if so, then does the end tag match the expected one. 

More generally, the code in specific parsing routines will generally be looking ahead only for the specific content item or items that are allowed to happen next. For example, the subroutine `_ReadGraphMLGraphElementStartTag()` would be called because the caller is expecting a graph element as possible next content and its lookahead found `<g`. Within `_ReadGraphMLGraphElementStartTag()`, the code would start by attempting to match `<graph` and, failing that, then it would emit an "Expected graph element start tag on line %d" and return `NOTOK`. On the other hand, if the characters match `<graph` are matched in the input, then those characters are consumed. At that point, the only acceptable next input is either at least one character of whitespace before a first attribute or a greater-than symbol ('>') to end the start tag. In this GraphML Reader Part 1, the `<graph>` element is required to have some attributes, so the next input must be whitespace. Hence, the next lookahead character must either match a whitespace character, or `_ReadGraphMLGraphElementStartTag()` should emit an "Unexpected content in <graph> element start tag on line %d" error and return `NOTOK`. Assuming a whitespace character was found, `_ReadGraphMLGraphElementStartTag()` should unget it and then invoke `_ReadGraphMLSkipWhitespace()`. Crucially, the next expected content after skipping the whitepace is an attribute name, not a comment. So, if the next lookahead character after skipping the whitespace is not a valid character for starting one of the supported attribtues of a `<graph>` element, then `_ReadGraphMLGraphElementStartTag()` would again emit an error message and return `NOTOK`.

## Skipping Whitespace

It will be useful to have a function named `_ReadGraphMLSkipWhitespace()`. It should return `OK` for success or `NOTOK` for failure. It should look ahead for each character using `sf_getc()`. If it is a whitespace character (space, tab, return, or newline), then discard it and keep going until a non-whitespace character is found, `EOF` is found, or a read error is found. Return `NOTOK` on the read error. 

If a whitespace character is found and it is a newline (0x0A), then increase the line number count. If a return (0x0D) is found, then you have to look ahead one character and only increment the line number if the next character is not a newline. And, if the next lookahead is other than `EOF` (a -1 integer, not 0xFF character), use `sf_ungetc()` to put it back.  

Once a non-whitespace character or `EOF` is found, use `sf_ungetc()` if it is other than EOF. Then, once `EOF` or a non-whitespace character was found, then the decision whether to return `OK` or `NOTOK` must be based on an additional parameter below.

Add an additional boolean parameter `requireWhitespace`. If `TRUE`, then at least one whitespace character must have been read to return `OK`. Otherwise, `OK` can be returned even if zero whitespace characters were read (provided there was no input error from the underlying `sf` container).  

## Reading XML Atributes for GraphML

In general, GraphML elements as well as the XML Prolog, if present, have at least one required attribute. However, there are a few elements that may be supported in the future that may have zero attributes. The implementations of the start tag processors (and the prolog processor) know which attributes they support and will generally continue reading attributes until they hit they end of start tag or prolog marker, or until they hit an unsupported attribute. Reading a single attribute consists of the following:

1. Reading an attribute name
2. Returning NOTOK unless the found attribute name matches one of the expected attributes
3. Skipping optional whitespace
4. Recognizing and then skipping a required equal sign (`NOTOK` if not found before `EOF`)
5. Skipping optional whitespace
6. Recognizing and then skipping a required double quote or single quote mark  (`NOTOK` if not found before `EOF`)
7. Finding the matching end double quote or single quote (must match the opening quote; `NOTOK` if not found before `EOF`)
8. If an expected attribute value to match is given, `strncmp()` match it or return `NOTOK` for a mismatch (no need to normalize for this matching case.
9. If the expected attribute value is NULL, then extract and normalize the attribute value (`NOTOK` for memory failure or normaization failure)

To avoid memory churning, the caller of `_ReadGraphMLAttribute()` is expected to provide a pre-allocated buffer and buffer size, such as by using `MAXLINE`. The implementation of `_ReadGraphMLAttribute()` is expected to return `NOTOK` if the attribute value extraction attempts to exceed the given buffer size.

Once an attribute value is extracted, `_ReadGraphMLAttribute()` must perform attribute value normalization, which consists of:

1. If the encoding is UTF-8, then convert UTF-8 encoded characters to single byte characters, including 0xFF (`NOTOK` if the UTF-8 is unable to be converted to a single-byte char)
2. Convert the following five character encodings to their single-byte equivalents, if found: &amp;, &lt;, &gt;, &apos;, and &quot;
3. It is `NOTOK` to have a bare `<`, `&` or quote mark matching the quotes used for the attribute value (preceding list, step 6) 

Name the method that performs this work `_ReadGraphMLAttribute()`. Let this method receive a list of acceptable attributes that could be read, as well as the list length. Because the exact supported attributes for each supported element are known, these lists can be created by a static declaration to avoid lots of costly `malloc()` and `free()` calls that churn memory. Each attribute structure needs to have the following:

 * The attribute name, such as `id`, `source`, or `target`
 * An `int` boolean flag for whether the attribute is required
 * A `char *` pointer to an expected attribute value, or NULL if the attribute value is to be extracted 

The caller of `_ReadGraphMLAttribute()` is a start tag or XML prolog processor that must be responsible for several things:

1. Look ahead to see if there is whitespace, and if so, put the character back then skip required whitespace
2. For an element start tag, if the next character is `>` or the next two are `/>` then the end of the start tag has been found
3. For an XML prolog, the characters `?>` signify the end of the prolog
4. Construct the required attribute list and size, passing them to `_ReadGraphMLAttribute()`
5. Pass a character buffer and buffer size to receive an extracted attribute value, if the attribute found has a NULL expected attribute value   
6. It is 'NOTOK` to get `EOF` in the middle of `/>` or `/?`
7. It is `NOTOK` to get `EOF` before the start tag or prolog end-marker is recognized
8. It is `NOTOK` to receive an attribute value more than once
9. It is `NOTOK` to end reading attributes but not have values for all required attributes

## The XML Prolog for GraphML

The `_ReadGraphMLProlog()` method has a few responsibilities. The first is to test for and deal with content encoding information that may be at the beginning of the content. Most notably, there is the byte order mark (BOM), if present. If the first byte is 0xFE, 0xFF, 0x00, or if the first two bytes are 0x3C followed by 0x00, then the file is marked as being UTF-16 encoded, so emit an error message about UTF-16 being unsupported and then return `NOTOK`. If the first byte is 0x4C, then since we have already been routed here by `_ReadGraph()`, the full test of EBCDIC encoding has already been done, so emit an error message about EBCDIC being unsupported and then return `NOTOK`. If the first byte is 0xEF and the two subsequent bytes are 0xBB then 0xBF, then the file is UTF-8, so record that the graph encoding is UTF-8, consume the three bytes, and move to the next step. Otherwise, the expected first content of the file must be either `<?xml` for an XML declaration, or `<!--` as the start of a comment, or a whitespace character. If there are any other starting bytes, emit an unsupported GraphML content message (with line number) and return `NOTOK`.

If the start of an XML declaration was detected, then process it, as always using a subroutine `_ReadGraphMLXMLDeclaration()`. An error may occur, such as if the `version` attribute is absent or not `1.0`, if the `standalone` is expressed but not `yes`, if the `encoding` is other than `UTF-8`(the default) or `ISO-8859-1`, or if an expressed `encoding` is other than `UTF-8` if the UTF-8 BOM was detected above, or if there is any unrecognized content in the XML declaration (such as other attributes, unexpected characters or `EOF`, or failure to end with `?>`). If reading the XML declaration was successful, then it is expected that its bytes will be consumed, including changes of line number that may occur due to whitespace in the XML declaration (which should happen naturally due to calling `_ReadGraphMLSkipWhitespace()`). Record the graph encoding for later use (`UTF-8` or `ISO-8859-1`).
     
After the XML declaration, whether or not present, the desired next token is the `<graphml>` start tag, but there may legitimately be whitespace and comments that have to be skipped before that. There may also be unsupported content like an XML DTD or XML Processing Instruction that, if present, must result in an unsupported content error message and returning `NOTOK`. A white space lookahead character should result in invocation of `_SkipGraphMLWhitespace()`. A lookahead of `<!--` should result in invocation of `_SkipGraphMLComment()`. Any other lookahead content except `<g` should cause an unsupported content error and a return of `NOTOK`. This process should occur until `<g` is found (unless EOF is erroneously reached). Once a lookahead of `<g` is found, unget the two bytes and return `OK`. 

**NOTE:** It will be the job of `_ReadGraphMLStartTag()` to ensure that the next bytes match `<graphml` rather than `<g` plus anything else. 

## The `<graphml>` Element and Content

The full GraphML Reader Part 1 should be articulated consistently with the descriptions above based on a functional decomposition that reflects the attribute and element children of each element described in the Technical Overview above.

This begins with the previously mentioned methods `_ReadGraphMLProlog()`, `_ReadGraphMLStartTag()`, `_ReadGraphMLKeys()`, and `_ReadGraphMLGraphElement()`.

The method `_ReadGraphMLStartTag()` has to start by recognizing and moving past the initial `<graphml` token, or reporting failure if the token is not recognized (because the method is not called until the caller has used the lookahead strategy to determine that `<graphml` is indeed the expected next token). This Part 1 reader requires at least one attribute (`xmlns`). 

After successfully reading and handling all given attributes of the `graphml` tag, it is necessary to skip optional whitespace and comments to find the next element declaration inside of the `graphml` tag. This is expected to be either a `<key>` element or a `<graph>` element. If the lookahead strategy has determined that a `<key>` element is present, then unget the lookahead characters and invoke `_ReadGraphMLKeys()` followed by skipping optional whitespace and comments. It is `NOTOK` to detect other than the expected elements.

Once the `key` elements have been read, if any, then we enter the main logic for reading a graph. We skip optional whitespace and comments, then look for the start of an expected `<graph>` element. It is `NOTOK` for it to not be next. If the lookahead strategy has determined that a `<graph>` element is present, then unget the lookahead characters, and invoke `_ReadGraphMLGraphElement()`. Once it finishes successfully, there is no need to skip optional whitespace and look for a `</graphml>` end tag because it will be useful for a graph read iterator to have this state when trying to read more than one graph from a GraphML file (future feature). 

In the present implementation, `_ReadGraphMLKeys()` need only recognize the key for `graphflags_zerobasedio`, which must be exact, except note that there may be optional whitespace and comments to discard before and after the child `<default>` element.

The `_ReadGraphMLGraphElement()` implementation should stsart with invoking a subroutine to focus on the start tag. Only the `edgedefault` attribute is required. Some of the other attributes, if present, need to have their attribute values validated according to previously-mentioned restrictions, the only `parse.order` is `nodesfirst` and the `parse.nodeids` and `parse.edgeids` must be `canonical`. Some attribute values, if given, can be partially validated at first but then must be saved for later completion of validation, such as `parse.nodes` and `parse.edges` must be positive integers that are later checked to ensure they match the number of `<node>` and `<edge>` elements found. Finally, some attribute values must be saved to guide later processing, such as the `edgedefault` that indicates whether `<edge>` elements should indicate directed versus undirected edges by default.

After the `<graph>` start tag, there maybe optional whitespace and comments to ignore before and after each child element. There may be zero or more `<data>` elements as the first element children of `<graph>`, so the lookahead strategy must be used to decide whether to call `_ReadGraphMLGraphDataElements()`. Currently, the only supported data element is for `graphflags_zerobasedio`. Then, because we will start by only supporting `nodesfirst` order, we expected to keep looking ahead and deciding to read `<node>` elements with `_ReadGraphMLGraphNode()`. The only other non-whitespace, non-comment element that should be acceptable is the start of the first `<edge>` element. Then, use the lookahead strategy to keep calling `_ReadGraphMLGraphEdge()` as long as the next non-whitespace, non-comment element continues to be `<edge>`. Once the last `edge` element is read, then skip optional whitespace and comments, then recognize a required `</graph>` end tag.

Recognizing each `<node>` element is straighforward. It must have an `id` attribute which has an attribute value that exactly matches `n` plus an index value starting at 0 and incremented by the caller, and passed to `_ReadGraphMLGraphNode()` so that it can construct the attribute value to match. No other attribute values are supported. The end of the `<node>` element can be signalled by the empty tag syntax (`/>`) or by a true empty tag (`> _optional ignored whitespace and comments_ </node>`).

Once all nodes are read, the number can be compared to `parse.nodes` if given. It is `NOTOK` to mismatch. This is the moment when `gp_EnsureVertexCapacity()` can be invoked.

Reading each `<edge>` element should include validating that the `id` attribute value is canonical if the `id` attribute is given, obtaining the required `source` and `target` attribute values to use in a call to `gp_DynamicAddEdge()`, and obtaining the `directed` attribute value, if given. The `directed` attribute value, if given, should be validated against the values `true` and `false`. It is used if given and otherwise default to the graph's `edgedefault`. If the edge should be directed, then `gp_SetDirection()` must be used to set the direction from the source to the target. The `source` and `target` attribute values must be references to canonical `id` values for the vertices (nodes), so it is easy to validate the attribute values, make sure they are in the 0 to N-1 range, and then add an input offset as is done in `_ReadAdjList()` (e.g., with `inputOffset = gp_LowerBoundVertexStorage(theGraph) - (zeroBased ? 0 : 1);`).     

## Testing

TBD

WilliamK112:
Resolves #342

## Type of change

### Added

* `c/graphLib/io/graphML-reader.c` implements the deliberately restricted GraphML Part 1 parser described in the issue, including supported prolog encodings, comments and character entities, root/key/graph/node/edge validation, zero-based graph flags, canonical IDs, and directed edges.
* GraphML regression coverage includes round trips of both existing writer samples, valid string inputs for the supported encoding and syntax variants, and rejection cases for unsupported or inconsistent content.

### Updated

* `c/graphLib/io/graphIO.c` detects XML-like input without losing embedded or leading bytes and routes it to the GraphML reader.
* `Makefile.am` builds the new reader.
* `c/planarityApp/planarityCommandLine.c` runs GraphML read and validation tests as part of the quick regression suite.

## Testing

- Configured and built with `--enable-compile-warnings=error`.
- Ran `./planarity -test c/samples`; all tests succeeded, including the existing 143 read-error checks and the new GraphML cases.
- Ran `make check`; 1/1 tests passed.
- Rebuilt with Clang AddressSanitizer and UndefinedBehaviorSanitizer and reran the full quick regression suite; all tests succeeded with no sanitizer findings.

<details><summary>Regression summary</summary>

```
Starting GraphML Tests
Finished GraphML Tests.

============

All tests have succeeded.

Testsuite summary for planarity 5.1.0.0
# TOTAL: 1
# PASS:  1
# FAIL:  0
# ERROR: 0
```

</details>


c/graphLib/io/graphIO.c
```diff
@@ -46,6 +47,50 @@ int _WritePostprocess(graphP theGraph, char **pExtraData);
 /* Private functions */
 char _GetEdgeTypeChar(graphP theGraph, int e);
 char _GetObstructionMarkChar(graphP theGraph, int v);
+static int _ReadInputLooksLikeGraphML(strOrFileP inputContainer, int *pResult);
```
john-boyer-phd: Let's declare this as extern, rename the function to _IsGraphMLInput() and change the signature to be like s6_IsSparse6Input(). Receive a lineBuff and return bool. See below for different hook-in.

c/graphLib/io/graphIO.c
```diff
@@ -46,6 +47,50 @@ int _WritePostprocess(graphP theGraph, char **pExtraData);
...
+
+static int _ReadInputLooksLikeGraphML(strOrFileP inputContainer, int *pResult)
```
john-boyer-phd: Please rename, change signature and move to graphML-reader.c

c/graphLib/io/graphIO.c
```diff
@@ -46,6 +47,50 @@ int _WritePostprocess(graphP theGraph, char **pExtraData);
...
+    for (index = byteCount - 1; index >= 0; index--)
+    {
+        if (sf_ungetc(bytes[index], inputContainer) != bytes[index])
+            return NOTOK;
+    }
+
+    if (byteCount == 0)
+        return OK;
+
+    if (bytes[0] == '<' || bytes[0] == ' ' || bytes[0] == '\t' ||
```
john-boyer-phd: Please add comment: These are the legal first bytes of an XML file that are not in common with any other currently supported file format.

c/graphLib/io/graphIO.c
```diff
@@ -46,6 +47,50 @@ int _WritePostprocess(graphP theGraph, char **pExtraData);
...
+        bytes[0] == '\b' || bytes[0] == '\r' || bytes[0] == '\n' ||
+        bytes[0] == 0xEF || bytes[0] == 0xFE || bytes[0] == 0xFF ||
+        bytes[0] == 0)
+        *pResult = TRUE;
+    else if (byteCount == 3 && bytes[0] == 0x4C &&
```
john-boyer-phd: Please add comment: This is a test for XML in EBCDIC encoding. Not supported but detected.

c/graphLib/io/graphIO.c
```diff
@@ -508,13 +553,26 @@ int _ReadGraph(graphP theGraph, strOrFileP *pInputContainer)
     int RetVal = OK;
 
     int extraDataAllowed = FALSE;
+    int graphMLInput = FALSE;
```
john-boyer-phd: Delete this line.

c/graphLib/io/graphIO.c
```diff
@@ -508,13 +553,26 @@ int _ReadGraph(graphP theGraph, strOrFileP *pInputContainer)
...
     char lineBuff[MAXLINE + 1];
 
     memset(lineBuff, '\0', (MAXLINE + 1));
 
     if (!sf_IsValidStrOrFile((*pInputContainer)))
         return NOTOK;
 
+    if (_ReadInputLooksLikeGraphML((*pInputContainer), &graphMLInput) != OK)
```
john-boyer-phd: Once this function signature is changed, it will be more evident how to wire it in further below, as an else clause just above the one that routes to sparse6.

The key here is to wire it in the same way as all the other formats rather than doing it differently.

c/graphLib/io/graphIO.c
```diff
@@ -567,6 +625,8 @@ int _ReadGraph(graphP theGraph, strOrFileP *pInputContainer)
         RetVal = NOTOK;
     }
 
+readComplete:
```
john-boyer-phd: Delete this, you won't need it anymore.

c/planarityApp/planarityCommandLine.c
```diff
@@ -3051,6 +3057,234 @@ int runBasicGraphMLWriteTest(void)
...
+            "attr.name='graphflags_zerobasedio' attr.type='boolean'/>"
+            "<graph edgedefault='directed' parse.nodeids='canonical' "
+            "parse.edgeids='canonical' parse.order='nodesfirst'>"
+            "<data key='graphflags_zerobasedio'> false </data>"
+            "<node id='n0'/><node id='n1'/>"
+            "<edge source='n0' target='n1' directed='false'/></graph>",
+            FALSE, FALSE
+        }
+    };
+    static char const *rejectCases[] = {
```
john-boyer-phd: Please add a linefeed and a comment to each rejection case to say what it is intended to reject, and to make it more readable.

Please add a lot more rejection cases for things like:
 * truncations of start and end tags of every element (separate test for each)
 * Bad attribute values of each attribute, e.g. parse.nodeids or parse.edgeids being other than canonical, parse.order being other than nodesfirst, edgedefault being other than directed or undirected, directed being other than boolean, not matching xmlns on graphml element (separate test for each).
 * A reasonable number for parse.nodes or parse.edges that is nonetheless incorrect; the current 99999.... test tries a number that probably doesn't even parse as an integer, which is a different test (separate tests per attribute)
 * Node or edge id value not being canonical (separate test for each)
 * Node elements after the start of edge elements
 * Data elements after node elements start
 * Invalid edge source or target attribute value, e.g., 'xyz' and 'n1000' in a graph with less than 1000 vertices (separate rejection tests needed for each)
 * Invalid edge target attribute value. 
 * Illegal character in attribute value, such as ampersand, less-than. 
 * Test for illegal single quote in single-quoted attribute value and double-quote in double-quoted attribute value, which should fail in an interestingly different way, especially if there is a space after the "illegal" quote mark (the failure may be EOF or the occurrence of a next start or end tag).
 * Illegal attribute or element child of graph, node, edge, data, key, graphml (separate tests for each possible infraction)
 * A key element other than the one we support
 * A graph-level data element other than the one we support
 * A non-boolean content value for the data element content of the graphflags_zerobasedio graph level data element.
 * EBCDIC byte code start
 * UTF-16 byte order mark
 * A UTF-8 byte order mark followed by a declared encoding of ISO-8859-1

For the purpose of testing, we will have to permit the graphml element to have one "illegal" attribute called testencoding. This will allow testing of acceptable and rejectable cases of UTF-8 transcoding to single byte (ISO-8895-1 encoding). The spec mentions two banks of two-byte UTF-8 encodings that should be tested for correct conversion (in the accepted tests above). Rejection tests should include starting with 0xC2 then an illegal second byte, 0xC3 then an illegal second byte, and an illegal first byte other than 0xC2 and 0xC3.

c/planarityApp/planarityCommandLine.c
```diff
@@ -3051,6 +3057,234 @@ int runBasicGraphMLWriteTest(void)
     return OK;
 }
 
+int runGraphMLReadWriteTest(char const *inputFileName)
```
john-boyer-phd: The GraphML testing is becoming big enough that I need to ask if you could please move them out of `planarityCommandLine.c` and to a new file called `testGraphML.c` in the `planarity-test` directory. This will include also amending the `makefile.am`, and the functions can then be imported into `planarityCommandLine.c` as extern declarations.

c/planarityApp/planarityCommandLine.c
```diff
@@ -3051,6 +3057,234 @@ int runBasicGraphMLWriteTest(void)
...
+
+    length = strlen(graphMLStr) + 1;
+    copy = (char *)malloc(length);
+    if (copy != NULL)
+        memcpy(copy, graphMLStr, length);
+
+    return copy;
+}
+
+int runGraphMLAcceptTest(char const *graphMLStr, int zeroBased, int directed)
```
john-boyer-phd: The acceptable cases data structure should be increased to contain N and M, then pass to here so the comparison of the read-in graph's N and M can be compared to the passed values, rather than to literals 1 and 2. This will enable growing the acceptable cases graphs when needed.

c/planarityApp/planarityCommandLine.c
```diff
@@ -3051,6 +3057,234 @@ int runBasicGraphMLWriteTest(void)
...
+        {
+            "<?xml version='1.0' encoding='ISO-8859-1'?>"
+            "<graphml xmlns='http://graphml.graphdrawing.org/xmlns'>"
+            "<key id='graphflags_zerobasedio' for='graph' "
+            "attr.name='graphflags_zerobasedio' attr.type='boolean'/>"
+            "<graph edgedefault='directed' parse.nodeids='canonical' "
+            "parse.edgeids='canonical' parse.order='nodesfirst'>"
+            "<data key='graphflags_zerobasedio'> false </data>"
+            "<node id='n0'/><node id='n1'/>"
+            "<edge source='n0' target='n1' directed='false'/></graph>",
```
john-boyer-phd: To really test `edgedefault`, you actually need two acceptance cases that each have a two-edge graph where you tell `runAcceptGraphMLTest()` the endpoints of one edge that should be directed and one edge that should be undirected.

In the `edgedefault='undirected'` case, one edge will need to be declared `directed='true'`. In the `edgedefault='directed'` case, one edge will need to be declared `directed='false'`.

c/planarityApp/planarityCommandLine.c
```diff
@@ -3051,6 +3057,234 @@ int runBasicGraphMLWriteTest(void)
...
+    graphP G = gp_New();
+    char *inputCopy = NULL;
+    unsigned quietModeCache = gp_GetQuietMode();
+    int Result = OK;
+
+    if (G == NULL || (inputCopy = copyGraphMLTestString(graphMLStr)) == NULL)
+        Result = NOTOK;
+
+    gp_SetQuietMode(QUIETMODE_ALL);
+    if (Result == OK && gp_ReadFromString(G, inputCopy) == OK)
```
john-boyer-phd: Because you are on the inside of the graphLib, you have the ability to call a package-private function instead of gp_ReadFromString().

The reason I mention it is that the function signatures of package-private functions can be changed without breaking backwards compatibility of graphLib. The specific way that that helps is that you may be able to call some version of `_ReadGraph()` that will give you back the line number. This would allow you to test whether the failure occurred on the line it is expected to occur on.

c/graphLib/io/graphML-reader.c
```diff
@@ -0,0 +1,1517 @@
...
+    int chars[MAXLINE + 1];
+    size_t charsRead = 0;
+    size_t index = 0;
+    size_t length = 0;
+
+    (void)theGraph;
+    (void)pLineNum;
+
+    if (expected == NULL || pMatches == NULL)
+        return NOTOK;
```
john-boyer-phd: For every `return NOTOK`, we need a `gp_ErrorMessage()` that says something about what is wrong and, most importantly, what line number the error is on.

john-boyer-phd (changes requested):
The parser looks promising, but it seemed the best idea to request some refactoring that beefs up the test suite (and a couple of other smaller refactorings) because these tests will either prove out the precision of the parser or help to prompt needed changes to it.

PR #344, opened 2026-09-23
Write sparse6 and incremental sparse6 (EPIC #38, step 4)

Deln0r:
Fourth step of #38, to the API settled there: a sparse6 write iterator with the change batch, the modification counter that guards it, and `gp_CompactEdgeStorage()`, which the reader now uses as well. The bytes are the ones nauty writes.

Not in this PR: the reader lookahead `s6_RetrieveGraphChange()`, which per your note on #341 belongs at the `gp_` level, with every format but sparse6 reporting no increment.

## Type of change

### Added

* `c/graphLib/io/s6-write-iterator.[ch]`: `s6_NewWriter()`, `s6_InitWriterWithString()`, `s6_InitWriterWithFileName()`, `s6_SetOutputErrorFlag()`, `s6_WriteGraph()` and `s6_FreeWriter()` mirror the graph6 writer; `s6_StoreGraphChange(it, e, u, v)` stores the deletion of the edge `e` in use, kept by its endpoints so that no stored change can stale another, or the addition of `{u, v}` between two real vertices that the graph lacks. Loops, parallel edges, a pair stored twice in one batch, and anything stored before the first whole graph are refused at store time. `s6_WriteGraph()` writes the whole graph as a `:` line when nothing is stored, otherwise the batch as a `;` line, which it then applies to the graph and compacts. A batch is relative to the graph written last, so a graph modified directly since then is refused and the batch discarded, and a whole write is the way to continue. Hidden edges are refused, whole or batch, because the graph in memory and the graph in the file would disagree about them. A refusal leaves the output as it was; a failure while writing or applying a `;` line marks the output as failed, since nothing could continue from it. The header `>>sparse6<<` is always written, as the graph6 writer writes its own. The one-shot path behind `gp_Write()` mirrors the graph6 one, including that `sf_Free()` does not report a failed close of the file; if you want that reported, it is a `strOrFile` change that would serve both writers.
* `c/samples/K10.g6.0-based.S6.out.s6`, `N5-all.g6.0-based.S6.out.s6`, `nauty_example.g6.0-based.S6.out.s6`, `nauty_example.s6.0-based.S6.out.s6`: the expected output of `-s` for those inputs.

### Updated

* `c/graphLib/graph.[ch]`: `gp_CompactEdgeStorage()`, between `gp_EnsureEdgeCapacity()` and `gp_ResetGraphStorage()`, fills every hole that deletions leave. It sorts the holes once, so that the last pair, when it is a hole, is always on top of the stack, and moves the last pair in use into the top hole through `gp_DeleteEdge()` and `gp_DynamicAddEdge()`, so that extensions which overload deletion stay consistent; the flags of both records travel with the edge, whose position in its two adjacency lists changes. What does not travel is per-edge extension data, such as the path a reduction edge of the K4 or K3,3 search stands for, since the overloads clear it for the vacated records and initialize it for the new ones; the header says so, and the reader and writer only ever call it between algorithms. The modification counter is incremented in `gp_InsertEdge()`, `gp_DeleteEdge()`, `gp_CopyGraph()`, `gp_CopyAdjacencyLists()`, `gp_HideEdge()`, `gp_RestoreEdge()`, `gp_IdentifyVertices()`, `gp_RestoreVertex()`, `gp_ResetGraphStorage()` and `gp_CompactEdgeStorage()`, and in `gp_ClearEdgeDirectionFlags()` and `gp_TransposeDirectedGraph()` when they change a flag. The wrappers count before calling the overloadable function, since a failure part way through has already changed the graph.
* `c/graphLib/graph.private.h`: `modificationCount` in `graphPrivateDataStruct`, read with `theGraphModificationCount()` and advanced with `gp_NoteModification()`, both package private, as you asked.
* `c/graphLib/graphDFSUtils.c`: `gp_SortVertices()` counts as well.
* `c/graphLib/io/s6-read-iterator.c`: one `gp_CompactEdgeStorage()` per `;` line instead of a fill after each deletion.
* `c/graphLib/io/graphIO.[ch]`: `WRITE_SPARSE6` for `gp_Write()` and `gp_WriteToString()`.
* `c/planarityApp/planarityTransformGraph.c`, `planarityUtils.c`, `planarity.h`, `planarityHelp.c`, `planarity.1`: `-x -s`, with the `.out.s6` suffix, the `S6` name in the expected-file convention, and the S choice in the menu.
* `TestSupport/planaritytesting/planaritytesting_utils.py`: the `s` letter among the format specifiers.
* `c/planarityApp/planarityCommandLine.c`: `runSparse6WriteTests()`, below.
* `Makefile.am`, `c/samples/Makefile.am`, `c/graphLib/graphLib.h`: the new files.

## Testing

- Single lines against the specification's example and against what `copyg -s` writes, including the padding case the specification singles out, at orders 4 and 16, the four-byte order at 63, and the graph of order 1.
- All 34 graphs of order 5 written whole and compared with `N5-all.s6`, then written incrementally, each batch being the symmetric difference from the previous graph, and compared with `N5-all.inc.s6`; after every batch the writer's graph equals the graph read and has no holes.
- Every refusal of `s6_StoreGraphChange()` and `s6_WriteGraph()`: before the first write, an edge not in use, an edge already present, a loop, a vertex out of range, wrong argument combinations, a pair twice in one batch, a direct edit between writes, a hidden edge, a digraph, a second initialization. Each leaves the writer able to write, and the four lines that come out are the ones `copyg` writes for those graphs.
- `gp_CompactEdgeStorage()` on K5 with holes in the middle and at the end, recorded by either record of their pairs: no holes afterwards, the edge count and edge set unchanged, the direction flag of the moved edge kept; a dense graph and a NULL graph behave as documented.
- `-s` on `nauty_example.g6`, `N5-all.g6`, `K10.g6` from a file and from a string, and on `nauty_example.s6`.
- Outside the tree, against `copyg -s` and `copyg -i` on every graph of orders 2 to 10, which is 12,293,433 graphs, and on random graphs of orders 16, 31, 32, 63, 64, 100, 255, 256 and 1000, in both modes: no byte differs, the header aside.
- Five guards removed one at a time, each making a named test fail: the padding case, the counter check, the hidden-edge check, the flag carry, and compaction itself, without which `-t -d` over `n8.mALL.inc.s6` fails at graph 9 as it did before #329's fill.
- Containers on Ubuntu 24.04: gcc-14 and clang-18 under `--enable-compile-warnings=error`, gcc-14 with `--enable-unsigned-char`, plain gcc and clang, and `make distcheck`. Ubuntu 26.04: clang with ASan and with UBSan. macOS Apple clang: error gate and `make check`. All clean.


john-boyer-phd (changes requested):
This is looking really great, Ian. Just a few asks:

1. It turns out `planarityCommandLine.c` is getting too big because of all the awesome testing being added. I need to do a separate issue about breaking out the backlog of tests in it into separate files in the `planarity-test` subdirectory, but for now it seems important to not grow the problem. Therefore, would you please break out the sparse6 testing into a `testSparse6.c`  in `planarity-test`? 
2. Around line 689 of `planarityRandomGraphs.c` there is a place that prompts for saving G6. Could you please add another prompt for S6 (and code in and around to support it)? The reason is that I am pretty sure sparse6 even without incremental will prove useful for saving very large maximal planar graphs for testing. A single 10 million vertex graph takes about half a gig on disk in adjacency list format, and I'm betting sparse6 will be smaller and faster to write (and read).
3. Of course, the preceding point requires being able to write sparse6 graphs larger than 100,000 vertices. That limit was placed on Graph6 because it reads the whole graphs and requires quadratic space to do so, but it is not needed for Sparse6 because you are reading and writing bytes at a time, not the whole graph, and the whole line is as sparse as the graph is. So, I'd like to ask if you could please lift the 100,000 vertex limit. Our real limit is the 178 million. Also, my apologies I didn't mention this on the reader pull request, but we have to be able to read what we write. In both cases, reading and writing, this will affect the code for processing the graph order. However, it will be really exciting to see if sparse6 will give a substantially smaller file and read/write time than the adjacency list format because graph6 is clearly not able to do this.

**Edit:**

4. It looks to be the case that `gp_CompactEdgeStorage()` cannot avoid being an overloadable in the function table if it is to be public. The problem is that `gp_DeleteEdge()` overloads will destroy extension edge data that `gp_DynamicAddEdge()` won't recover. The function works for the current use case, but would have to be made private if we stick with it as-is. The alternative pretty nice though. Instead of having to call `qsort`, the main implementation for the base class graph would just take successively endmost edge record pairs and copy them into the successive holes indicated by the `edgeHoles` stack/array. Each such copied edge record pair will have to repair its non-`NIL` adjacency list predecessors and successors to point to the new index locations. Each extension overload would do the same thing to its edge data structure, then call the base class function. The base class function would be last, so it can just set the `edgeHoles` stack size to zero at the end. Would you find it to be OK to do now, or do you want to move this function private and have me file a separate issue about how to make it public? It's not urgent, so either way would be OK.

I'm looking through all the rest still, but it's all looking really great so far. Thank you, Ian.

